Issue #87, opened 2018-04-03
Can not build rock when tarantool is installed outside /usr


rosik:
See: https://github.com/tarantool/tarantool/issues/3148

```console
$ tarantoolctl rocks install /opt/tarantool-vshard/vshard-scm-1.rockspec
...
CMake Error at /usr/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find TARANTOOL (missing: TARANTOOL_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindTarantool.cmake:26 (find_package_handle_standard_args)
  CMakeLists.txt:14 (find_package)


-- Configuring incomplete, errors occurred!
See also "/tmp/luarocks_vshard-scm-1-XIrF2G/vshard/build.luarocks/CMakeFiles/CMakeOutput.log".

Error: Build error: Failed cmake.
```

PR #88, opened 2018-04-03
Fix tarantool rocks installation problems

rosik:


rosik:
This commit fixes #87 
The original issue is here: https://github.com/tarantool/tarantool/issues/3148

PR #154, opened 2018-10-04
Fix using space.id with using space.name

filonenko-mikhail:


Khatskevich:
Thanks for the fixes. They are included to the brunch (on review) https://github.com/tarantool/vshard/commits/kh/rebalance_by_name .

PR #177, opened 2019-04-14
Allow to use luarocks to install vshard

Totktonada:
See https://github.com/tarantool/rocks/pull/42 and
https://github.com/tarantool/rocks/pull/61

### DIFF
--- a/vshard-scm-1.rockspec
+++ b/vshard-scm-1.rockspec
@@ -10,7 +10,12 @@ description = {
     license  = 'BSD',
 }
 dependencies = {
-    'tarantool >= 1.9.0';
+    'lua >= 5.1';
+}
+external_dependencies = {
+    TARANTOOL = {
+        header = 'tarantool/module.h';
+    };
 }
 build = {
     type = 'cmake';


vshard-scm-1.rockspec
```diff
@@ -10,7 +10,12 @@ description = {
     license  = 'BSD',
 }
 dependencies = {
-    'tarantool >= 1.9.0';
+    'lua >= 5.1';
+}
+external_dependencies = {
+    TARANTOOL = {
+        header = 'tarantool/module.h';
```
Gerold103: But vshard does not depend in tarantool headers.

vshard-scm-1.rockspec
```diff
@@ -10,7 +10,12 @@ description = {
     license  = 'BSD',
 }
 dependencies = {
-    'tarantool >= 1.9.0';
+    'lua >= 5.1';
```
Gerold103: Why? Vshard obviously depends on Tarantool, not just on Lua.

Totktonada: I wrote my reasoning already in https://github.com/tarantool/rocks/pull/42#issuecomment-482983290

vshard-scm-1.rockspec
Totktonada: Hm. We can remove that, yep. I added this to match tarantool/rocks specs.

vshard-scm-1.rockspec
Gerold103: We need to check Tarantool version, because VShard does not work correctly on Tarantool < 1.9. And someday we will deprecate 1.9 support too as so on. If I can't do that via rockspec, then probably it can be done via CMakeLists.txt, but current implementation of find_packet(Tarantool) relies on tarantool/module.h header, which is not needed to VShard.

Totktonada: We need to catch something like https://github.com/tarantool/stat/commit/cd70219fffc549fa6cb4e10f71e1b3cce4df8a22 and use tarantool version that is reported by `tarantool --version`.

Gerold103:
Is there any progress? This PR can help me to fix travis.

Totktonada:
No progress for now, I'll try to look into that on the week.

Totktonada:
I'll reopen if I'll find time to finish it.

PR #202, opened 2019-11-19
Prohibit update for config for router/storage.cfg from multiple fibers

tian-xi:
Add and use function `apply_storage_cfg`/`apply_router_cfg` to sync updates of `router`/`storage` config from multiple fibers.

Calls of apply_router_cfg (usage of `is_reload` argument) has to be reviewed.


Gerold103:
Hi, thanks for working on this!
1) Please, create one commit, based on the master branch. We don't create multiple commits fixing each other. Changes should be atomic. For this particular issue it should be one self-sufficient commit. The old branch from rave4life was wrong, AFAIR. I recommend you to create your own implementation.
2) Write more descriptive commit messages. Currently I don't understand what each commit does. Look at other commits in the master branch, and in the main repository.

tian-xi:
@Gerold103  Moved this to #203

PR #203, opened 2019-11-21
Sync update of storage and router config by multiple fibers.

tian-xi:
Add and use function apply_storage_cfg/apply_router_cfg to sync updates of router/storage config from multiple fibers. See commit message for details.


test/router/router.result
```diff
@@ -1347,6 +1347,67 @@ vshard.router.route(1):callro('echo', {'some_data'})
...
+fiber_cfg_status:get()
+---
+- name: f2
+  status: true
+...
+-- Second fiber gets configuration timeout error.
+vshard.router.internal.errinj.ERRINJ_CONCURRENT_CFG_TIMEOUT = true
+---
+...
+vshard.consts.ROUTER_CFG_TIMEOUT = 1
```
Gerold103: I propose you to drop vshard.consts.ROUTER/STORAGE_CFG_TIMEOUT. Because this really contradicts with how *.cfg works. In Tarantool *.cfg (box.cfg, swim.cfg, router.cfg, storage.cfg) never throws a timeout error. They may hang forever, may leave the instance in an orphan state, may return an error on a wrong configuration, but they don't throw a timeout error. I would like to keep it that way so as users would not need to check for timeouts.

test/router/router.result
```diff
@@ -1347,6 +1347,67 @@ vshard.router.route(1):callro('echo', {'some_data'})
...
+...
+fiber_cfg_status:get()
+---
+- name: f2
+  status: false
+...
+fiber_cfg_status:get()
+---
+- name: f1
+  status: true
```
Gerold103: Please, reset the errinj.ERRINJ_CONCURRENT_CFG_TIMEOUT back to nil/false. Otherwise you slowdown next tests.

test/router/router.result
```diff
@@ -1347,6 +1347,67 @@ vshard.router.route(1):callro('echo', {'some_data'})
...
+_ = test_run:cmd("setopt delimiter ';'")
+---
+...
+function fiber_cfg(name) 
+	fiber.create(function()
+    	local status, err = pcall(vshard.router.cfg, cfg)
+		fiber_cfg_status:put({
+            name = name,
+            status = status,
+		})
```
Gerold103: Something is very wrong with the indentation here.

vshard/router/init.lua
```diff
@@ -796,6 +799,28 @@ local function router_cfg(router, cfg, is_reload)
     end
 end
 
+--
+-- Synchronize update of router configuration from mutiple fibers.
+-- At each point of time only single fiber can execute 'router_cfg' for particular router.
+-- Other fibers are waiting for current call of 'router_cfg' to return.
```
Gerold103: Please, keep comments inside 0-66 columns. Besides being part of our code style (and in picodata as well), it also really helps to read, when you do it 8-12 hours per day. Please, check all other places you added too.

vshard/router/init.lua
```diff
@@ -796,6 +799,28 @@ local function router_cfg(router, cfg, is_reload)
...
+--
+local function apply_router_cfg(router, cfg, is_reload)
+    if not router.cfg_lock:put(true, consts.ROUTER_CFG_TIMEOUT) then
+        error('Configuration timeout')
+    end
+
+    if M.errinj.ERRINJ_CONCURRENT_CFG_TIMEOUT then
+        fiber.sleep(consts.ROUTER_CFG_TIMEOUT * 2)
```
Gerold103: I would not use any exact timings here. When you rely on timeouts, you may create a 'flaky' test, which will sometimes fail, and is usually hard to catch and fix. We already have plenty of them. So timeouts should be avoided when possible. Also you used really high timeout here - 2 seconds just for 1 test case. Taking into account, that parallel tests does not work here, it is a significant slowdown.

Instead, I propose you to do this.
1) Rename ERRINJ_CONCURRENT_CFG_TIMEOUT to ERRINJ_CONCURRENT_CFG_COUNT.
2) In a test set it to 0.
3) Start 2 fibers. Each time a fiber goes here and sees ERRINJ_CONCURRENT_CFG_DELAY ~= false, it increments it, and makes assertion that the counter == 1. And decrements back in the end of apply_router_cfg.

olmiik:
@Gerold103 Currently unit/garbage.test.lua and failover/failover.test.lua are failing for some jobs. While tests for storage and router are ok. Will look at this in more detail. Can it be related to changes in other lua modules? Commenting out all synchronization code, that I have added in this branch, does not seem to fix issue with mentioned tests.

Gerold103:
Thanks for the updates! I just checked - these tests fail has nothing to do with your changes. Looks like a bug in the Tarantool core: https://github.com/tarantool/tarantool/issues/4770
I am not sure we can continue VShard development until this is fixed.

Gerold103:
@olmiik seems like the bug with __pairs was fixed in the Tarantool master, so we can proceed with vshard development. Tests are working now.

PR #215, opened 2020-02-25
ci: add release publishing to rocks.tarantool.org

olegrok:


Gerold103:
Does this commit mean, that I don't need to send new rocks for each tag manually?

olegrok:
Yes

printercu:
It also pushes scm-1 rock for every new commit in master

PR #217, opened 2020-02-27
fix wrong variable for error message

R-omk:


Gerold103:
I fixed commit message and added a test. Your branch is `romk/gh-217-move-to-self-no-init-var`. I will merge it, when all tests will pass.

PR #226, opened 2020-03-10
discovery based on active buckets

jeevansirela:
In my opinion, discovery should be based on active buckets rather than just buckets. We(Loco company) faced an issue in the morning. We fixed it by doing this in our router instances.


opomuc:
Closes #255

PR #242, opened 2020-06-03
Reconnection replica in case uuid mismatches

### LINKED ISSUE #241: Don't drop reconnects on UUID mismatch
In cartridge there is another iproto server implementation (remote-control) used for accessing uninitialized instances with `net.box`. During cluster initialization there may be a race: vshard.router could be initialized earlier than vshard.storage does, so router establishes connection to remote-control. Since the storage isn't initialized yet, remote-control always reports `peer.uuid == require('uuid').NULL`.

As a result, vshard router sees uuid mismatch and closes conection and never tries to reconnect. So we stay with disconnected replica until next reconfiguration.

https://github.com/tarantool/vshard/blob/79a4dbfc4229e922cbfe4be259193a7b18dc089d/vshard/replicaset.lua#L69-L74

I'd like to avoid this situation, and keep reconnection attempts.

rosik:
In cartridge there is another iproto server implementation
(remote-control) used for accessing uninitialized instances with
`net.box`. It always reports `peer.uuid == require('uuid').NULL`. During
cluster initialization there may be a race: `vshard.router` could be
initialized earlier than `vshard.storage`, so the router establishes
connection to remote-control.

As a result, vshard router sees uuid mismatch and closes conection and
never tries to reconnect. So we stay with disconnected replica until
next reconfiguration.

This patch preserves reconnection attempts even in case uuid mismatches.

Close #241

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -70,6 +70,12 @@ local function netbox_on_connect(conn)
         log.info('Mismatch server UUID on replica %s: expected "%s", but got '..
                  '"%s"', replica, replica.uuid, conn.peer_uuid)
         conn:close()
+        fiber.new(function()
+            -- Workaround for https://github.com/tarantool/vshard/issues/241
+            -- Keep reconnection attempts even if uuid mismatch
+            fiber.sleep(consts.RECONNECT_TIMEOUT)
+            rs:connect_replica(replica)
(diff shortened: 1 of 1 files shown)


rosik:
I've just realized this patch isn't correct - with this implementation it'll never stop reconnecting.

PR #243, opened 2020-06-03
Fix race between master and replica storage.cfg

### LINKED ISSUE #237: vshard.storage.cfg fails if replica calls it before master
In cartridge we've encountered the following case:

We've got two instances, m-m replication is already set up. One is leader (rw), the other one is read-only.
We enable vshard-storage role, and both instances call `vshard.storage.cfg` almost simultaneously.

If replica calls `vshard.storage.cfg` before master, it fails.

In v0.1.15 it results in error `vshard/storage/init.lua:2389: attempt to index field '_bucket' (a nil value)`.

In earlier versions it resulted in `cfg` hanging  forever, because of `box.once` which in turn calls `wait_rw` which never ends on replica.

rosik:
If `vshard.storage.cfg` is called on reblica before than on master it
fails saying "vshard/storage/init.lua:2389: attempt to index field
'_bucket' (a nil value)".

The only thing space `_bucket` is used for is to set `on_replace`
trigger.

This patch postpones trigger initialization basing on another trigger -
`_space:on_replace()`.

Close #237

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -2435,8 +2435,22 @@ local function storage_cfg(cfg, this_replica_uuid, is_reload)
     local uri = luri.parse(this_replica.uri)
     schema_upgrade(is_master, uri.login, uri.password)
 
-    local old_trigger = box.space._bucket:on_replace()[1]
-    box.space._bucket:on_replace(bucket_generation_increment, old_trigger)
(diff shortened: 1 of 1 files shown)


rosik:
Re-implemented in https://github.com/tarantool/vshard/commit/0cd777f3990939b32d96bf964dea55e8073e65bb. Closing

PR #295, opened 2021-09-28
test: fix parallel.test.lua and sync.test.lua

ylobankov:
With the current version of test-run, that vshard uses now (d171b6e),
all tests work fine and everything is ok. The problem occures when
attempting to use a later version of test-run starting since 53389a8.
In this case, the parallel.test.lua and sync.test.lua tests start to
fail because the new test-run version catches some exception from tests
and exits with a failure.

This patch should be concidered as a preparation for updating test-run
version for vshard.

### DIFF
--- a/test/rebalancer/engine.cfg
+++ b/test/rebalancer/engine.cfg
@@ -15,7 +15,7 @@
         "memtx": {"engine": "memtx"},
         "vinyl": {"engine": "vinyl"}
     },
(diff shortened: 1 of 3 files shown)


Gerold103:
https://github.com/tarantool/vshard/commit/21f980800bde4ba2bc954f4a84afea3161639e42

Issue #293, opened 2021-08-20
Setup Github Actions


rosik:
Travis is dead, there's no CI in this repo atm. That's no good.

olegrok:
It's my failed attempt https://github.com/tarantool/vshard/commits/github-actions-ci

Issue #272, opened 2021-05-11
Resurrect CI / CD and make it nice


Totktonada:
Subtask of https://github.com/tarantool/tarantool/issues/6056

There are several problems with CI / CD in vshard:

- It uses Travis CI, which going to be paid service (.org infrastructure will gone soon, .com has tight limits).
- CI on several Linux distros is red (I don't know exact reason).
- No package deployments to our new S3 infra (see https://github.com/tarantool/tarantool/issues/3380, https://github.com/tarantool/tarantool/issues/4947, https://github.com/tarantool/tarantool/issues/5494)

We also can improve our CI in the following areas:

- Verify vshard against different tarantool versions.
- Verify build packages across different OS.
- Run tests on Mac OS.
- Verify installation using a rockspec.

The plan is to implement CI / CD for vshard on top of the GitHub Actions infrastructure, resolve existing problems and make it a way more powerful.

----

Estimated as 3 days work with the idea to reuse [smtp](https://github.com/tarantool/smtp/)'s CI / CD code after resolving of the following tasks: https://github.com/tarantool/smtp/issues/34, https://github.com/tarantool/smtp/issues/35, https://github.com/tarantool/smtp/issues/36.

PR #297, opened 2021-09-28
Add simple CI based on GitHub actions

ylobankov:
- Delete old .travis.yml file
- Add new .github/workflow/precommit_testing.yml


Totktonada (approved):
LGTM.

My wish: I would name the workflow, where we test a module only on GitHub provided Ubuntu machine as 'fast_testing' consistently across the modules (see tuple-keydef for example). However it is debatable.

Please, close the relevant issue from the commit message (#293).

It worth to mention #272: in fact, this PR is the first step on this road. You can also mention this commit as part of the issue https://github.com/tarantool/tarantool/issues/4972.

Please, leave comments in the code about 2.9: a link to the issue why it is disabled and a short description.

I would also leave a comment why we need the `fetch-depth` option.

Please, use some appropriate prefix for a commit title (I would use 'ci').

Gerold103:
(You asked me to give you all the comments instead of fixing them myself, so here you go)
Thanks for the patch! Please, merge the second commit into the first one. Because it changes a comment which you just added in the first commit.

.github/workflows/fast_testing.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          tarantool-version: ${{ matrix.tarantool }}
+
+      - name: Clone the module
+        uses: actions/checkout@v2
+        # Enable recursive submodules checkout as test-run git module is used
+        # for running tests.
+        with:
+          fetch-depth: 0
+          submodules: recursive
```
Totktonada, line 41: Cited my comment above:

> I would also leave a comment why we need the `fetch-depth` option.

Your comment is about the `submodules` option. But the `fetch-depth` one is less obvious as for me (because you need to know how the upgrade testing works, not just looks, whether there are submodules).

ylobankov: Sorry, I misunderstood you. Fixed.

.github/workflows/fast_testing.yml
```diff
@@ -0,0 +1,49 @@
+name: fast_testing
```
Gerold103, line 1: In the commit message you said: `Add new .github/workflow/precommit_testing.yml` but the patch does not add such file. It adds `.github/workflows/fast_testing.yml`.

ylobankov: Oops, right, fixed. It should be 'fast_testing' everywhere.

Gerold103 (approved):

PR #301, opened 2021-10-15
ci: disable fail-fast behaviour for tests

ylobankov:
Problem: when tests start and one of them fails, the test run stops, and
other tests are not executed, we exit after the first test failure. It's
not convenient since as a developer I would like to see all test results
and understand the impact of my changes on all project functionalities.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -6,3 +6,9 @@ add_custom_target(test
     COMMAND ${PROJECT_SOURCE_DIR}/test/test-run.py -j -1
         --builddir=${PROJECT_BINARY_DIR}
         --vardir=${PROJECT_BINARY_DIR}/test/var)
+
+add_custom_target(test-force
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -5,4 +5,5 @@ endif(POLICY CMP0037)
 add_custom_target(test
     COMMAND ${PROJECT_SOURCE_DIR}/test/test-run.py -j -1
         --builddir=${PROJECT_BINARY_DIR}
-        --vardir=${PROJECT_BINARY_DIR}/test/var)
+        --vardir=${PROJECT_BINARY_DIR}/test/var
+        --force)
```
Totktonada, line 9: In tarantool we have two make targets: `make test` and `make test-force`. I don't know whether we actully should follow this way, just highlighted that if we'll adjust `make test` in vshard, we'll lack consistency with tarantool here.

ylobankov: Thanks for your remark! Fixed.

Gerold103 (approved):

PR #315, opened 2022-01-18
router: fix undeclared vshard.error

### LINKED ISSUE #310: vshard/replicaset.lua:488: variable 'vshard' is not declared
> router-1 | 2022-01-10 21:32:12.777 [94604] main/144/main replicaset.lua:362 E> Exception during calling 'vshard.storage.call' on 'localhost:13302(admin@localhost:13302)': {"reason":"storage is disabled explicitly","code":33,"type":"ShardingError","message":"Storage is disabled: storage is disabled explicitly","name":"STORAGE_IS_DISABLED"}
    integration.vshard_storage_disabling.test ... (0.017s) error
...s/cartridge/.rocks/share/tarantool/vshard/replicaset.lua:488: variable 'vshard' is not declared

https://github.com/tarantool/vshard/blob/c092f038e4da2e1a69cc4ca90d1bf176560eb6a5/vshard/replicaset.lua#L488

Seems it should be lerror instead of vshard.error

yngvar-antonsson:
Close #310

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -485,7 +485,7 @@ local function can_backoff_after_error(e, func)
         end
     end
     if e.type == 'ShardingError' then
-        return e.code == vshard.error.code.STORAGE_IS_DISABLED
+        return e.code == lerror.code.STORAGE_IS_DISABLED
     end
     return false
 end
--- a/test/router/router2.result
+++ b/test/router/router2.result
@@ -590,13 +590,27 @@ fiber.sleep(vshard.consts.REPLICA_BACKOFF_INTERVAL)
  | ---
  | ...
 -- Success, but internally the request was retried.
-res, err = vshard.router.callro(1, 'echo', {100}, long_timeout)
+--
+-- n/a: there was a bug when an error code in the router depended
+-- on vshard.error being global. Nullify it to ensure it is not
+-- the case anymore.
+router = vshard.router
+ | ---
+ | ...
+vshard = nil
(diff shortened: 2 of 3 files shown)


sergos (approved):
LGTM

vshard/replicaset.lua
```diff
@@ -485,7 +485,7 @@ local function can_backoff_after_error(e, func)
         end
     end
     if e.type == 'ShardingError' then
-        return e.code == vshard.error.code.STORAGE_IS_DISABLED
+        return e.code == lerror.code.STORAGE_IS_DISABLED
```
olegrok, line 488: I suggest to add a test case for this error

Gerold103: Yes, I agree with Oleg. If you still want to continue with the PR, you would need to add a test. Preferably in `router2.test.lua`. In the same place where the other tests are located for catching the 'disabled' error.

Gerold103 (changes requested):
Please, try to follow the commit messages guidelines - https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message. The message should start with a subsystem prefix. Here it would be `router: fix undeclared vshard.error`.

You can also just wait until I get to that issue myself. I will then push your work and my fixes on top of it. Like I hope we will do with all new PRs to attract more contributors.

filonenko-mikhail:
lgtm

Gerold103 (approved):
Thanks for the fix and the test!

PR #316, opened 2022-01-18
router: fix undeclared var usage

### LINKED ISSUE #311: replicaset.lua:683:36: accessing undefined variable master_uuid
Seems it's a real bug:

https://github.com/tarantool/vshard/blob/c092f038e4da2e1a69cc4ca90d1bf176560eb6a5/vshard/replicaset.lua#L680-L684

yngvar-antonsson:
Close #311

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -680,7 +680,7 @@ local function replicaset_update_master(replicaset, old_master_uuid,
     if old_master_uuid == candidate_uuid then
         -- It should not happen ever, but be ready to everything.
         log.warn('Replica %s in replicaset %s reports self as both master '..
-                 'and not master', master_uuid, replicaset_uuid)
+                 'and not master', old_master_uuid, replicaset_uuid)
         return is_auto
     end
     local master = replicaset.master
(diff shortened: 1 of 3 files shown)


filonenko-mikhail:
> The change looks reasonable, still I can't get the point why we use 3 args for a format string that contains only 2 placeholders?

There are "format string with two concat parts"

vshard/replicaset.lua
```diff
@@ -680,7 +680,7 @@ local function replicaset_update_master(replicaset, old_master_uuid,
     if old_master_uuid == candidate_uuid then
         -- It should not happen ever, but be ready to everything.
         log.warn('Replica %s in replicaset %s reports self as both master '..
-                 'and not master', master_uuid, replicaset_uuid)
+                 'and not master', old_master_uuid, replicaset_uuid)
```
olegrok, line 683: I suggest to add a test case for this error

Gerold103: I agree with Oleg. Like in the other PR, you would need to add a test. Look at the end of `master_discovery.test.lua`. You also need to make the commit message conform with rules. Here it would be `router: fix undeclared var usage`.

Gerold103 (changes requested):

Gerold103 (approved):

PR #318, opened 2022-02-19
router: support return_raw in map_callrw()

### LINKED ISSUE #312: Support 'raw args' feature from the latest Tarantool to avoid args decoding on router
The newest Tarantool version has a new function setting `takes_raw_args`. Instance can create such function in `_func` and it will take arguments as a raw MessagePack buffer. The function then can decode some first arguments and leave the others not decoded. For instance, to forward them further to another instance.

Also netbox now supports `return_raw` option - works the same. Result of the remote call is a raw MessagePack buffer. Not decoded into Lua values.

These options together should help to reduce pressure on the router when it is used as a proxy. Router would need to offer a function like `vshard.router.callraw()`. It would take a MessagePack buffer with the same arguments encoded as `vshard.router.call()` and return result also as a MessagePack buffer (using `return_raw` netbox option when talks to the storage).

In order to use it the user would need to register it in `_func` with `takes_raw_args` and call it from clients. Router can't do that itself, because it is not supposed to depend on any local spaces.

Gerold103:
Closes #312

@TarantoolBot document
Title: vshard.router.map_callrw() format of `return_raw` result

`vshard.router.map_callrw()` supports `return_raw` option (in the
version of Tarantool which have it at all).

When there was an error, the result is the same as always:
`nil, err, err_uuid`. No changes.

When the call was a success, the result is a map of the format
`{[replicaset_uuid] = msgpack.object}`. The `msgpack.object` here
stores a MessagePack array with the results returned from the
storage's map-function.

The usecase is the same as for when bare netbox is used - to avoid
decoding of the results into Lua land. Helpful when the router is
used as a proxy and the results received from the storage are big.

Example:
```Lua
local res = vshard.router.map_callrw('my_func', args, {..., return_raw = true})

for replicaset_uuid, msgpack_value in pairs(res) do
    log.info('Replicaset %s returned %s', replicaset_uuid,
             msgpack_value:decode())
end
```
However this is just an example. Normally you don't need `return_raw` if you are
calling `:decode()` anyway.

The family of `vshard.router.call` functions also supports `return_raw`, but
their behaviour is the same as with bare netbox - the entire result is just an
`msgpack.object`. The same as if direct netbox call would be done. Nothing to
document here, I think.

### DIFF
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -21,9 +21,13 @@ end
 local function config_new(templ)
     local res = table.deepcopy(templ)
     local sharding = {}
+    local meta = {replicasets = {}}
     res.sharding = sharding
-    for _, replicaset_templ in pairs(templ.sharding) do
+    for i, replicaset_templ in pairs(templ.sharding) do
         local replicaset_uuid = uuid_next()
+        meta.replicasets[i] = {
+            uuid = replicaset_uuid
+        }
         local replicas = {}
         local replicaset = table.deepcopy(replicaset_templ)
         replicaset.replicas = replicas
@@ -36,7 +40,7 @@ local function config_new(templ)
         end
         sharding[replicaset_uuid] = replicaset
     end
-    return res
+    return res, meta
 end
 
 --
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -4,7 +4,7 @@ local vutil = require('vshard.util')
 local wait_timeout = 120
 
 local g = t.group('router')
-local cluster_cfg = vtest.config_new({
+local cluster_cfg, cfg_meta = vtest.config_new({
     sharding = {
         {
             replicas = {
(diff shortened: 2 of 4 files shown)


olegrok (approved):
Thanks for your patch. LGTM.

Single question:
> Closes #312

Does it really should close an issue? Are there any plans to implement callraw API?

R-omk:
> Does it really should close an issue? Are there any plans to implement callraw API?

@olegrok , for the most part, this ticket was discussed [here](https://github.com/tarantool/vshard/discussions/317)

olegrok:
> > Does it really should close an issue? Are there any plans to implement callraw API?
> 
> @olegrok , for the most part, this ticket was discussed [here](https://github.com/tarantool/vshard/discussions/317)

I read this discussion. But anyway there should be a way to take raw args on storage side. It could be useful for modules such a crud. Since box API allows to insert msgpack objects into spaces.

R-omk:
> to take raw args on storage side 

But this has nothing to do with vshard. 
Vshard simply calls functions, while the user himself can configure the function options (a function can be registered in `box.schema.func` with `takes_raw_args = true`)

olegrok:
> > to take raw args on storage side
> 
> But this has nothing to do with vshard. Vshard simply calls functions, while the user himself can configure the function options (a function can be registered in `box.schema.func` with `takes_raw_args = true`)

Hmm... I was sure that vshard proxies all calls from "vshard.storage.*" function. To be able to call ref on storage.

https://github.com/tarantool/vshard/blob/4962609d52d4a8935407cef75024f93051ea122b/vshard/storage/init.lua#L2504-L2514

Even I create a function with "takes_raw_args = true" it shouldn't have any effect.

R-omk:
@olegrok, you're right.

Gerold103:
> Thanks for your patch. LGTM.
> 
> Single question:
> 
> > Closes #312
> 
> Should it really close an issue? Are there any plans to implement callraw API?

#312 was about the router only. The `callraw` mentioned in the discussion was about having it on the router. `vshard.router.callraw` would still call `vshard.storage.call` on the storage with normal arguments, but `callraw` itself would take an `msgpack` object with an array of the same arguments as `vshard.router.call` except that the function arguments would be in the end.

This idea was discarded, because it would require you to either grant universe execution rights to connections to the router (unsafe) or call `box.schema.func.create('vshard.router.callraw', {takes_raw_args = true})` on the router. Router can't do the former action itself, because it should not depend on schema anyhow. Thus you either way need to do actions on the router. Meaning you can simply implement `callraw` as well as register it. Moreover, it can be done more optimal when you make it for a specific application.

Totktonada:
Regarding granting permissions: I found that a module-provided function like `module.grant(user_or_role)` is quite useful in terms of convenience. One don't need to dig into internals to make things just work.

In this particular case (where we need to register a function and optionally grant permission for it to given user), it can be `vshard.router.func.create('callraw'[, user])`, where `'callraw'` is a predefined name. Or `vshard.router.func.callraw.register([user])`. Something like this.

Gerold103:
https://github.com/tarantool/vshard/issues/319

PR #324, opened 2022-03-21
Use `msgpack_iterator:take_array()` method

Gerold103:
The patchset makes the router use `take_array()` method of msgpack iterator instead of manual tail cut via an intermediate Lua table.

### DIFF
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -186,9 +186,9 @@ g.test_map_callrw_raw = function(g)
     t.run_only_if(vutil.feature.netbox_return_raw)
 
     local create_map_func_f = function(res1)
-        _G.do_map = function(res2)
+        rawset(_G, 'do_map', function(res2)
             return {res1, res2}
-        end
+        end)
     end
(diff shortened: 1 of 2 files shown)


olegrok (approved):
LGTM
However it breaks all version checks introduced previously.
There is a "gap" between commits where msgpack object was introduced and is_array method was implemented.
But I believe that there should be nothing serious.

olegrok:
> LGTM However it breaks all version checks introduced previously. There is a "gap" between commits where msgpack object was introduced and is_array method was implemented. But I believe that there should be nothing serious.

I meant that to be consistent both of this conditions should be replaced with `msgpack.object ~= nil and msgpack.object({}).take_array ~= nil`.
https://github.com/tarantool/vshard/blob/cd321dd4b97952c35cad698c474a41eea5bbc35c/vshard/util.lua#L238-L240

Gerold103:
In the commit message I added specifically that distinguishing these cases wouldn't make much sense. Neither msgpack object, nor its iterator, not `take_array` methods are released.

PR #330, opened 2022-05-10
SSL and multilisten support

Gerold103:
The patchset allows to use multilisten and SSL features. Although there was a way to use multilisten before (as long as the URIs were strings and `listen` was passed to each storage in the root of its vhsard config), it is now easier.

### DIFF
--- a/test/instances/storage.lua
+++ b/test/instances/storage.lua
@@ -1,5 +1,11 @@
 #!/usr/bin/env tarantool
 local helpers = require('test.luatest_helpers')
+
+--
+-- Commonly used libraries.
+--
+_G.fiber = require('fiber')
+
 -- Do not load entire vshard into the global namespace to catch errors when code
 -- relies on that.
 _G.vshard = {
@@ -12,6 +18,7 @@ if box.ctl.set_on_shutdown_timeout then
 end
 
 box.cfg(helpers.box_cfg())
+local instance_uuid = box.info.uuid
 box.schema.user.grant('guest', 'super', nil, nil, {if_not_exists = true})
 
 local function box_error()
@@ -22,7 +29,23 @@ local function echo(...)
     return ...
 end
 
+local function get_uuid()
+    return instance_uuid
+end
+
+local function session_set(key, value)
+    box.session.storage[key] = value
+    return true
+end
+
+local function session_get(key)
+    return box.session.storage[key]
+end
+
 _G.box_error = box_error
 _G.echo = echo
+_G.get_uuid = get_uuid
+_G.session_set = session_set
+_G.session_get = session_get
 
 _G.ready = true
--- a/test/luatest_helpers/server.lua
+++ b/test/luatest_helpers/server.lua
@@ -160,6 +160,21 @@ function Server:instance_uuid()
     return uuid
 end
 
+function Server:replicaset_uuid()
+    -- Cache the value when found it first time.
+    if self.replicaset_uuid_value then
+        return self.replicaset_uuid_value
+    end
+    local uuid = self:exec(function() return box.info.cluster.uuid end)
+    if uuid == nil then
+        -- Probably didn't bootstrap yet. Listen starts before replicaset UUID
+        -- is assigned.
+        return nil
+    end
+    self.replicaset_uuid_value = uuid
+    return uuid
+end
+
 -- TODO: Add the 'wait_for_readiness' parameter for the restart()
 -- method.
 
@@ -192,6 +207,7 @@ function Server:cleanup()
     end
     self.instance_id_value = nil
     self.instance_uuid_value = nil
+    self.replicaset_uuid_value = nil
 end
 
 function Server:drop()
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -21,13 +21,9 @@ end
 local function config_new(templ)
     local res = table.deepcopy(templ)
     local sharding = {}
-    local meta = {replicasets = {}}
     res.sharding = sharding
     for i, replicaset_templ in pairs(templ.sharding) do
         local replicaset_uuid = uuid_next()
-        meta.replicasets[i] = {
-            uuid = replicaset_uuid
-        }
         local replicas = {}
         local replicaset = table.deepcopy(replicaset_templ)
         replicaset.replicas = replicas
@@ -40,7 +36,7 @@ local function config_new(templ)
         end
         sharding[replicaset_uuid] = replicaset
     end
-    return res, meta
+    return res
 end
 
 --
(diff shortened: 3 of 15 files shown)


test/unit-luatest/config_test.lua
```diff
@@ -0,0 +1,44 @@
...
+    t.assert_error(lcfg.check, config)
+
+    storage_1_a.uri = {}
+    t.assert_error(lcfg.check, config)
+
+    storage_1_a.uri = nil
+    t.assert_error(lcfg.check, config)
+
+    storage_1_a.uri = uuid.new()
+    t.assert_error(lcfg.check, config)
```
olegrok, line 43: I suggest to add "empty string" case. Since error message explicitly says about it.

test/luatest_helpers/server.lua
```diff
@@ -160,6 +160,16 @@ function Server:instance_uuid()
     return uuid
 end
 
+function Server:replicaset_uuid()
+    -- Cache the value when found it first time.
+    if self.replicaset_uuid_value then
```
olegrok, line 165: Could line 168 return box.NULL somehow? Maybe explicit check to nil is better.

test/instances/storage.lua
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env tarantool
```
olegrok, line 1: Cite commit message
```lua
                 uri = {
                        'storage:storage@127.0.0.1:3301',
                        params = {
                            transport = 'plain',
                            ssl_ca_file = ca_file,
                        }
                    },
```

Personally I don't like mixture of lua array and lua table. Such table couldn't be serialized as json, yaml, msgpack properly.

vshard/cfg.lua
```diff
@@ -130,6 +147,9 @@ local function check_replicas(replicas)
     local ctx = {master = false}
     for _, replica in pairs(replicas) do
         validate_config(replica, replica_template, ctx)
+        if replica.listen == nil then
+            replica.listen = replica.uri
+        end
```
olegrok, line 152: It looks a bit tricky. "Check" methods supposed to be read-only I think.

vshard/storage/init.lua
```diff
@@ -2692,7 +2692,11 @@ local function storage_cfg(cfg, this_replica_uuid, is_reload)
         --
         -- If a master role of the replica is not changed, then
         -- 'read_only' can be set right here.
-        box_cfg.listen = box_cfg.listen or this_replica.uri
+        local this_replicaset_cfg = vshard_cfg.sharding[this_replicaset.uuid]
+        local this_replica_cfg = this_replicaset_cfg.replicas[this_replica.uuid]
+
+        box_cfg.listen = box_cfg.listen or this_replica_cfg.listen or
+                         this_replica_cfg.uri
```
olegrok, line 2699: Still I afraid box.NULLs in such cases :)

vshard/util.lua
```diff
@@ -21,6 +21,43 @@ end
...
+--
+local function table_equals(a, b)
+    if type(a) ~= 'table' or type(b) ~= 'table' then
+        return type(a) == type(b) and a == b
+    end
+    local mta = getmetatable(a)
+    local mtb = getmetatable(b)
+    if mta and mta.__eq or mtb and mtb.__eq then
+        return a == b
+    end
```
olegrok, line 36: nit: Not completely sure that you need (even in future) this metatable magic.

olegrok (commented):
Thanks for your patch! See my comments below.

test/unit-luatest/config_test.lua
Gerold103: Done.

test/luatest_helpers/server.lua
Gerold103: I did this:
```Diff
@@ -166,6 +166,11 @@ function Server:replicaset_uuid()
         return self.replicaset_uuid_value
     end
     local uuid = self:exec(function() return box.info.cluster.uuid end)
+    if uuid == nil then
+        -- Probably didn't bootstrap yet. Listen starts before replicaset UUID
+        -- is assigned.
+        return nil
+    end
     self.replicaset_uuid_value = uuid
     return uuid
 end
```

test/instances/storage.lua
Gerold103: I took the format from EE tests. But you are right, better use `uri` key explicitly. Done.

vshard/cfg.lua
Gerold103: They fill default fields now. But defaults work only for constant values. Here I tried to make default of one field equal to another field. Anyway, I dropped this in the new version - the tests on old versions were failing because they don't support that option, and it is not needed much anyway.

vshard/storage/init.lua
Gerold103: Done:
```Diff
@@ -2695,8 +2695,12 @@ local function storage_cfg(cfg, this_replica_uuid, is_reload)
         local this_replicaset_cfg = vshard_cfg.sharding[this_replicaset.uuid]
         local this_replica_cfg = this_replicaset_cfg.replicas[this_replica.uuid]
 
-        box_cfg.listen = box_cfg.listen or this_replica_cfg.listen or
-                         this_replica_cfg.uri
+        if box_cfg.listen == nil then
+            box_cfg.listen = this_replica_cfg.listen
+            if box_cfg.listen == nil then
+                box_cfg.listen = this_replica_cfg.uri
+            end
+        end
         if not box_cfg.replication then
```

vshard/util.lua
Gerold103: I intentionally copy-pasted this from Tarantool master as is so as not to diverge too much from how it works in new versions. I might not needed it for configs but could need for something else in the future.

Gerold103:
I did a lot more changes after noticing that CI was failing on all old versions. Didn't attach the diff because it is too big and scattered over the commits.

test/luatest_helpers/vtest.lua
```diff
@@ -109,6 +159,59 @@ local function storage_new(g, cfg)
...
+            table.insert(storages, storage)
+            table.insert(fids, storage:exec(function(cfg)
+                local f = fiber.new(vshard.storage.cfg, cfg, box.info.uuid)
+                f:set_joinable(true)
+                return f:id()
+            end, {cfg}))
+        end
+    end
+    local errors = {}
+    for i, storage in pairs(storages) do
```
olegrok, line 184: It's better to use here ipairs. pairs doesn't give order guarantee (It works for now but it doesn't mean that it will work in future).

olegrok (approved):
Thanks. I left minor comments below. Also I suggest to file an issue to test SSL when bug will be fixed.

test/luatest_helpers/vtest.lua
Gerold103: Order does not matter here. I can send and collect responses in any order because they are executed in parallel all at the same time anyway.

Issue #314, opened 2022-01-17
bucket garbage collection does not work correctly with MVCC enabled (and maybe even w/o it)


CuriousGeorgiy:
# Description 
Bucket _garbage collection_ assumes the fact that each increment to `M.bucket_generation` in the `on_replace` trigger `bucket_generation_increment` for `box.space._bucket` corresponds to a tuple with `consts.BUCKET.GARBAGE` or `consts.BUCKET.SENT` status that will be _seen_ when querying `box.space._bucket` in the [bucket garbage collection loop](https://github.com/tarantool/vshard/blob/c092f038e4da2e1a69cc4ca90d1bf176560eb6a5/vshard/storage/init.lua#L1867-L1871): this assumption does not hold with MVCC enabled — the transaction manager can show versions of tuples before the `replace`. It may also not hold even by default (with MVCC disabled): the `replace` which changed the bucket status could not yet get committed to the  _WAL_ at the time the bucket garbage collection fiber queries `box.space._bucket` — this hypothesis was not tested and requires further investigation.

 # Reproducers
* with MVCC disabled: tarantool/cartridge#1667: see description in ticket
* with MVCC enabled: tarantool/tarantool#6344 — requires tarantool/tarantool#6788 to get merged (the _vshard_ bug here currently gets shadowed by a bug in memtx TREE iterators): 
with my patch applied you can see that the last bucket in node `sA` does not get garbage collected after [rebalancing](https://github.com/tarantool/cartridge/blob/fcc5bef36126bfcd9e11cffd08ff26f2539d8533/test/integration/rebalancing_test.lua#L146-L154)

# Behaviour
The following console output (with custom log messages inserted for clarity) aims to demonstrate the system's behavior with MVCC enabled and the reproducer above:
```console
sA-1 | 2022-01-14 15:28:24.001 [84516] main/161/vshard.rebalancer_worker_1 I> in rebalancer before last bucket replace
sA-1 | 2022-01-14 15:28:24.001 [84516] main/161/vshard.rebalancer_worker_1 I> before txn_commit_stmt
sA-1 | 2022-01-14 15:28:24.001 [84516] main/161/vshard.rebalancer_worker_1 I> in bucket generation_increment
sA-1 | 2022-01-14 15:28:24.001 [84516] main/161/vshard.rebalancer_worker_1 I> after txn_commit_stmt
[...]

PR #334, opened 2022-05-20
Support memtx MVCC

Gerold103:
#314
The patchset fixes a bug because of which the bucket garbage collector sometimes could ignore SENT/GARBAGE buckets when MVCC is enabled. The same without without MVCC I couldn't confirm.

Worth mentioning that I didn't test rebalancer with it. It requires to port much more tests to luatest. But I don't expect any problems. Bucket recovery could have the same issue but it also uses the generations and hence should be fixed too. It is not covered by specific tests by the same reason - too much work to port the existing tests to luatest just for that.


olegrok (approved):
Thanks for your patch. Looks OK. But Seems some changes are completely untested in CI.

I suggest to add something like:
```patch
From 7b1f389e8e1f26e2576b7e633f2ced9fdd7fd70a Mon Sep 17 00:00:00 2001
From: Oleg Babin <babinoleg@mail.ru>
Date: Sat, 21 May 2022 07:02:11 +0300
Subject: [PATCH] build: replace 2.9 version to latest in CI

There is no such version as 2.9. Also some new features requires
fresh version of Tarantool. So let's drop 2.9 in favour of
'latest'.
---
 .github/workflows/fast_testing.yml | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/.github/workflows/fast_testing.yml b/.github/workflows/fast_testing.yml
index 457f345..fb85ae0 100644
--- a/.github/workflows/fast_testing.yml
+++ b/.github/workflows/fast_testing.yml
@@ -21,7 +21,7 @@ jobs:
           - '1.10'
           - '2.7'
           - '2.8'
-          - '2.9'
+          - 'latest'
 
     runs-on: ubuntu-latest
     steps:
-- 
2.35.1
```

Just to be sure that it currently works and also it won't be broken in future.

Totktonada:
NB: Testing with the newest tarantool requires some work on the setup-tarantool action side.

Gerold103:
Bad that it is not tested on tarantool master in CI. But I don't know how CI is scripted so I better won't touch it. I can accept a PR though.

Issue #328, opened 2022-04-12
Consider usage of msgpack.object feature for rebalansing


olegrok:
Probably it's possible to declare function that receives tuples during rebalancing as `takes_raw_args` (https://github.com/tarantool/tarantool/pull/6666/commits/c03f9c79a783ee1d894cca928cdc02984364e611). And after it put directly to space (https://github.com/tarantool/tarantool/pull/6666/commits/ae48d9e551d18837947cef4facaa2d13d91580e6).

Probably it could speedup rebalancing procedure.

Issue #327, opened 2022-04-12
Does not work rebalancing of values of type 'Varbinary'


BelyaevEvgeny:
Vshard cannot rebalance value of type varbinary. Throws the following error:
```
mytaran.s1-master | 2022-04-12 12:29:56.792 [9373] main/202/vshard.rebalancer_worker_1 init.lua:2228 E> 
Error during rebalancer routes applying: receiver cce85c09-5f7a-49ae-9d52-23f905165a53, error {
"bucket_id":11,
"reason":"Tuple field 2 (value) type does not match one required by operation: expected varbinary, got string",
"code":30,
"message":"Can not receive the bucket 11 data in space \"test_space_cluster\" 
at tuple ['key16', !!binary ktRiFsYAAAAY8AeCpE5hbWWqVGVzdF9rZXkxNqNBZ2UB, 11]: 
Tuple field 2 (value) type does not match one required by operation: expected varbinary, got string",
"type":"ShardingError",
"space":"test_space_cluster",
"name":"BUCKET_RECV_DATA_ERROR",
"tuple":["key16","��b\u0016�\u0000\u0000\u0000\u0018�\u0007��Name�Test_key16�Age\u0001",11]}
```

Space settings:
```
spaces:
  test_space_cluster:
    engine: memtx
    is_local: false
    temporary: false
    sharding_key:
    - "key"
    format:
    - {name: key, type: string, is_nullable: false}
    - {name: value, type: varbinary, is_nullable: true}
    - {name: bucket_id, type: unsigned, is_nullable: false}
    indexes:
    - name: key
      unique: true
      parts: [{path: key, type: string, is_nullable: false}]
      type: TREE
    - name: bucket_id
      unique: false
      parts: [{path: bucket_id, type: unsigned, is_nullable: false}]
      type: TREE
```

Vshard can rebalance varbinary?

Totktonada:
Maybe we should go in the way proposed in https://github.com/tarantool/tarantool/issues/5933.

PR #336, opened 2022-06-01
Transfer bucket tuples as is without re-coding via Lua

Gerold103:
The patchset fixes a problem when rebalancer couldn't work if there were spaces with varbinary field type, or with double field type which stored at least one number not having a fraction. https://github.com/tarantool/vshard/issues/327

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -30,6 +30,8 @@ local test_rules = {
         '113/imsgpack',
         '113/ivconst',
         '113/ivshard',
+        '113/ivtest',
+        '113/ivutil',
     }
 }
 
--- a/test/storage-luatest/storage_1_1_test.lua
+++ b/test/storage-luatest/storage_1_1_test.lua
@@ -1,0 +1,160 @@
+local t = require('luatest')
+local vtest = require('test.luatest_helpers.vtest')
+local vutil = require('vshard.util')
+local wait_timeout = 120
+
+local group_config = {{engine = 'memtx'}, {engine = 'vinyl'}}
+
+if vutil.feature.memtx_mvcc then
+    table.insert(group_config, {
+        engine = 'memtx', memtx_use_mvcc_engine = true
+    })
+    table.insert(group_config, {
+        engine = 'vinyl', memtx_use_mvcc_engine = true
+    })
+end
+
+local test_group = t.group('storage', group_config)
+
+local cfg_template = {
+    sharding = {
+        {
+            replicas = {
+                replica_1_a = {
+                    master = true,
+                },
+            },
+        },
+        {
+            replicas = {
+                replica_2_a = {
+                    master = true,
+                },
+            },
+        },
+    },
+    bucket_count = 10
+}
+local cluster_cfg
+
+test_group.before_all(function(g)
+    cfg_template.memtx_use_mvcc_engine = g.params.memtx_use_mvcc_engine
+    cluster_cfg = vtest.config_new(cfg_template)
+
+    vtest.storage_new(g, cluster_cfg)
+    vtest.storage_bootstrap(g, cluster_cfg)
+    vtest.storage_rebalancer_disable(g)
+end)
+
+test_group.after_all(function(g)
+    g.cluster:drop()
+end)
+
+--
+-- Test how bucket_send preserves tuple field types over the network (gh-327).
+--
+test_group.test_bucket_send_field_types = function(g)
+    -- Make a space with all the field types whose msgpack representation is
+    -- nontrivial.
+    local _, err = vtest.storage_exec_each_master(g, function(engine)
+        local format = {
+            {'id', 'unsigned'},
+            {'bid', 'unsigned'},
+        }
+        local i = 1
+        local tuple = {i, box.NULL}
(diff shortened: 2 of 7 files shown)


vshard/util.lua
```diff
@@ -236,6 +236,19 @@ local function table_minus_yield(dst, src, interval)
     return dst
 end
 
+--
+-- Append all items of src to the end of dst. It is assumed both tables are
+-- arrays.
+--
+local function table_concat(dst, src)
```
olegrok, line 243: There is default `table.concat` this name confuses me a bit. Maybe table_ extend (as in python)?
https://docs.python.org/3/tutorial/datastructures.html#more-on-lists

vshard/util.lua
```diff
@@ -236,6 +236,19 @@ local function table_minus_yield(dst, src, interval)
...
+    local dst_i = #dst + 1
+    for _, v in ipairs(src) do
+        dst[dst_i] = v
+        dst_i = dst_i + 1
+    end
```
olegrok, line 248:
```
tarantool> dst = {1, 2, 3}
---
...

tarantool> src = {4, 5, 6}
---
...

tarantool> table.move(src, 1, #src, #dst + 1, dst)
---
- - 1
  - 2
  - 3
  - 4
  - 5
  - 6
...

tarantool> dst
---
- - 1
  - 2
  - 3
  - 4
  - 5
  - 6
...
```

Gerold103:
Sorry, CI is failing (the branch was green before I made the PR). I will investigate.

.luacheckrc
```diff
@@ -30,6 +30,8 @@ local test_rules = {
         '113/imsgpack',
```
olegrok, line 30: Doesn't it close https://github.com/tarantool/vshard/issues/328 as well?

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -0,0 +1,160 @@
...
+end)
+
+test_group.after_all(function(g)
+    g.cluster:drop()
+end)
+
+--
+-- Test how bucket_send preserves tuple field types over the network (gh-327).
+--
+test_group.test_bucket_send_field_types = function(g)
```
olegrok, line 56: CI is completely red. Tests should be skipped for the versions below 2.10

vshard/storage/init.lua
```diff
@@ -576,6 +577,72 @@ local function schema_downgrade_from_0_1_16_0()
...
+    local func = box.func[name]
+    if func == nil then
+        return
+    end
+    if func.takes_raw_args then
+        return
+    end
+    log.info("Upgrade %s to accept args as msgpack", name)
+    box.begin()
+    local ok, err = pcall(function()
```
olegrok, line 603: Maybe pcall(box.atomic, function() ... end) without explicit box.begin/box.commit?

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -0,0 +1,160 @@
...
+    local _, err = vtest.storage_exec_each_master(g, function(engine)
+        local format = {
+            {'id', 'unsigned'},
+            {'bid', 'unsigned'},
+        }
+        local i = 1
+        local tuple = {i, box.NULL}
+        -- Decimal appeared earlier, but this is when it can be used as a field
+        -- type. Same logic with other types.
+        if ivutil.version_is_at_least(2, 3, 0, nil, 0, 0) then
```
olegrok, line 68: A bit useless check because msgpack.object is available since 2.10 anyway.

olegrok (commented):
Thanks for your patch. I left several comments below.

Also I need to check that this patch won't break TDG because ... because of some netbox monkeypatches currently it's impossible to use takes_raw_args.
I guess `util.feature.msgpack_object = false` should help me.

test/storage-luatest/storage_1_1_test.lua
Gerold103: I want the tests work since 1.10. Decimal appeared before 2.10.

vshard/storage/init.lua
Gerold103: I don't trust built-ins like atomic.

Gerold103:
1.10 upgrade test fails on master too, I can't reproduce it locally. Don't know what happens, but caused not by this patch.

test/storage-luatest/storage_1_1_test.lua
olegrok: Ok. (Somewhy I don't have "resolve discussion" button here)

vshard/storage/init.lua
olegrok: Ok

test/storage-luatest/storage_1_1_test.lua
olegrok: Ok

vshard/util.lua
Gerold103: Renamed to `table_extend`.

vshard/util.lua
Gerold103: Thanks, I took this expression as the body of `table_extend`.

olegrok (approved):
Thanks!

Issue #332, opened 2022-05-12
Add SSL tests when it works in EE


Gerold103:
This branch https://github.com/tarantool/vshard/tree/gerold103/gh-325-ssl-and-multilisten has SSL tests. Unfortunately, they crash now. They need to be added when SSL is fixed in EE.

PR #337, opened 2022-06-01
Cover SSL with tests

Gerold103:
https://github.com/tarantool/vshard/issues/332


vshard/util.lua
```diff
@@ -287,7 +289,11 @@ local feature = {
     -- It appeared earlier but at 2.9.0 there is a strange bug about an immortal
     -- tuple - after :delete(pk) it still stays in the space. That makes the
     -- feature barely working, can be treated like it didn't exist.
-    memtx_mvcc = version_is_at_least(2, 10, 0, nil, 0, 0)
+    memtx_mvcc = version_is_at_least(2, 10, 0, nil, 0, 0),
+    -- SSL is introduced earlier. But at some point could crash on any
+    -- misconfiguration because Tarantool not always handled malformed packets
+    -- correctly. On another version SSL reconfig could kill replication.
+    ssl = is_enterprise and version_is_at_least(2, 11, 0, 'entrypoint', 2, 0),
```
olegrok, line 296: I guess in 2.10.1 it will work find

Gerold103: While SSL itself might work fine, replication reconfiguration with SSL turned on was still broken in 2.10.1. I said it in the comment above: *`On another version SSL reconfig could kill replication`*. It was fixed here: https://github.com/tarantool/tarantool/commit/c6485d8d3ef346b39b73a189e9358c1d76dca206 and merged into EE since 2.11.0-entrypoint-2.

olegrok: This commit was backported to 2.10 branch.

2.10 of CE - https://github.com/tarantool/tarantool/commit/7f916ce4c75bb97c059790c3fd1f3338f8518186
and there is corresponding commit to 2.10 EE.
So current 2.10.0 is broken but next 2.10.1 should be OK.

Gerold103: Thanks for noticing. I didn't know we use the same branch-per-version approach in EE as well. Fixed now.

olegrok (approved):

Issue #173, opened 2018-12-21
bucket_ref on replica does not prevent bucket move


Gerold103:
Bucket refs are used to prevent bucket movements during an active request. But only master now makes a decision about rebalancing, without consulting replicas. Master should ask replicas if they have any refs.

Maybe it will require prohibition of rebalancing when a replica is not available.

Gerold103:
Solution for ro_refs: bucket GC should not delete bucket content until all replicas confirmed they have no ro_refs. This does not block rebalancing or anything. GC will just be postponed a little. This will work for current vshard since it has no master-master support.

For #165 rw_refs should be checked as well on all masters. And it is significantly more complex. A master, decided to move a bucket, should not start, until the bucket is readonly on all other masters. Problems emerge from 1) unavailability of some other masters, 2) a necessity to make a bucket read-write back after some timeout.

Maybe it will require modification of _bucket format to make communications about refs easier. Refs will be hold in lua tables still, but decisions about making a bucket read-write/read-only/gc will be done via special fields in _bucket tuples.

R-omk:
Another solution:
before starting the migration of the bucket, wait until the completion of all requests on all replicas except for the replica on which the rebalancer worker is running.
At this time, all new requests should be redirected by the router to the main replica.
Now it can proceed with the migration of the bucket and enable rw locks.

R-omk:
The replica could use a bucket crowding out algorithm similar to the algorithm that could be used here
https://github.com/tarantool/vshard/issues/219

Gerold103:
Another raw idea: let replicas work independently and keep their refs. Let masters send buckets like now. But!!! on the replica side subscribe on _bucket changes. When applier on the replica will receive from the master node a request to change bucket state to sending/receiving/garbage, don't apply that request until all refs are gone. Don't know how to do that yet though. Because applier is async, and because can't yield for memtx spaces (in on_replace triggers). However maybe this idea could be elaborated. For instance, raise an error in the applier when master tries to change state of a bucket which has incompatible refs on this replica. This will break replication, but at least will prevent the existing requests from reading garbage.

Gerold103:
There is another cool idea from @R-omk which I like the most among the listed ones. The idea is basically in "let it crash", but instead "let it fail".

The ticket is about RO refs, because until #165 the master is just one. And I am not sure #165 will be implemented ever anyway. With horizontal scaling you don't need multiple masters in a replicaset.

[...]

PR #340, opened 2022-06-21
Bucket ref protection on replicas - part 1 - new fail-safe measures

Gerold103:
The patchset starts work on the big issue of bucket refs being not protected from incorrect changes of _bucket. Mostly that is about the main bug which allows to garbage collect a bucket on a master instance even if it has some read-requests in progress on replicas. That makes the requests potentially access inconsistent partially deleted data.

The fix consists of multiple steps which are being sent as separate patchsets. Otherwise there would be no progress for too long while I am busy with non-vshard activities in the meantime.

#173

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1189,8 +1189,8 @@ local function bucket_unrefro(bucket_id)
     local ref = M.bucket_refs[bucket_id]
     local count = ref and ref.ro or 0
     if count == 0 then
-        return nil, lerror.vshard(lerror.code.WRONG_BUCKET, bucket_id,
-                                  "no refs", nil)
+        return nil, lerror.vshard(lerror.code.BUCKET_IS_CORRUPTED, bucket_id,
+                                  "no ro refs on unref")
     end
     if count == 1 then
         ref.ro = 0
@@ -1244,8 +1244,8 @@ end
 local function bucket_unrefrw(bucket_id)
     local ref = M.bucket_refs[bucket_id]
     if not ref or ref.rw == 0 then
-        return nil, lerror.vshard(lerror.code.WRONG_BUCKET, bucket_id,
-                                  "no refs", nil)
+        return nil, lerror.vshard(lerror.code.BUCKET_IS_CORRUPTED, bucket_id,
+                                  "no rw refs on unref")
     end
     if ref.rw == 1 and ref.rw_lock then
         ref.rw = 0
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -83,6 +83,7 @@ if not M then
             ERRINJ_RECOVERY_PAUSE = false,
             ERRINJ_UPGRADE = false,
             ERRINJ_DISCOVERY = false,
+            ERRINJ_BUCKET_GC_PAUSE = false,
         },
         -- This counter is used to restart background fibers with
         -- new reloaded code.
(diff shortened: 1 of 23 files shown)


vshard/storage/init.lua
```diff
@@ -2685,14 +2685,21 @@ end
...
-    _, err = bucket_unref(bucket_id, mode)
-    assert(not err)
+    ok_ref, err = bucket_unref(bucket_id, mode)
+    if not ok_ref then
+        -- It should not normally happen. But the bucket could be deleted
+        -- manually or due to a bug. Then the read request could see
+        -- inconsistent data - the bucket could be half-deleted, for example.
+        -- Regardless of what user's function returns, treat it as an error.
+        return nil, err
+    end
```
R-omk, line 2702: I propose to write log with the original error (ret1) if it happened else it will be lost. Or write it to log in any case.

vshard/storage/init.lua
```diff
@@ -292,8 +293,62 @@ local function bucket_generation_increment()
...
+        local ref = M.bucket_refs[bid]
+        if ref ~= nil then
+            -- Lock shouldn't be ever set here. But could be that an admin
+            -- decided to restore a SENT bucket back into ACTIVE state due to
+            -- any reason.
+            ref.ro_lock = false
+        end
+        return
+    end
+    -- Ignore unknown statuses. Could be set no idea how, but just ignore.
```
olegrok, line 325: Maybe we can at least log them?

olegrok (approved):
LGTM in general

I left two minor comments.

vshard/storage/init.lua
Gerold103: At first I thought you are right and went to add the log, but then remembered that we don't have a ratelimited log in Lua (not publicly). That means if we have a massive bucket status rename, we will produce potentially gigabytes of logs all over the cluster. I propose to keep it like that, it is at least not worse that it was. And in future commits I plan to add validation into `on_replace`, so a bad status won't even be able to commit.

vshard/storage/init.lua
```diff
@@ -2625,14 +2685,21 @@ end
 -- @retval nil, err Error.
 -- @retval values Success.
 local function storage_call(bucket_id, mode, name, args)
-    local ret1, ret2, ret3, _
-    local ok, err = bucket_ref(bucket_id, mode)
-    if not ok then
-        return ok, err
+    local ret1, ret2, ret3
+    local ok, ok_ref, err
```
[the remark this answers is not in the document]
Gerold103: Indeed, that was a leftover from a previous version of the patch:
```Diff
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -2685,13 +2685,11 @@ end
 -- @retval nil, err Error.
 -- @retval values Success.
 local function storage_call(bucket_id, mode, name, args)
-    local ret1, ret2, ret3
-    local ok, ok_ref, err
-    ok_ref, err = bucket_ref(bucket_id, mode)
+    local ok_ref, err = bucket_ref(bucket_id, mode)
     if not ok_ref then
         return nil, err
     end
-    ok, ret1, ret2, ret3 = local_call(name, args)
+    local ok, ret1, ret2, ret3 = local_call(name, args)
     ok_ref, err = bucket_unref(bucket_id, mode)
     if not ok_ref then
         -- It should not normally happen. But the bucket could be deleted
```

vshard/storage/init.lua
Gerold103: In the new version of the patch I added support for stacked diagnostics and return the user function's exception in `err.prev` member. The caller can now log it if he wants.

Gerold103:
@olegrok please, take another look. I added a new commit to support stacked diag, and reworked the last commit to return user function's error along with unref error together.

olegrok (approved):
Thanks for fixes and  changes.

Issue #194, opened 2019-09-11
Router should not start listen before it is ready to accept requests


Gerold103:
First `vshard.router.cfg` creates internal static router object, and configures it. During configuration it calls `box.cfg{listen = ..., ...}`, but before the configuration is actually finished. It leads to a problem, that router has already started listening, but is not ready to serve requests yet. The same problem exists for storage - box.cfg is called before _bucket is created.

sergos:
https://github.com/tarantool/vshard/commit/5d935fc2783148b45aff584a60d039462488cd56 should resolve the storage part of the problem

PR #342, opened 2022-06-23
router: auto and manual enable/disable

### LINKED ISSUE #291: Error on bootstrap
On 0.1.18 when bootstap cluster with cartridge-cli got the following error:
2021-08-11 19:16:40.482 [64441] main/133/http/127.0.0.1:44012 graphql.lua:313 E> Graphql execution failed: ...-cartridge/.rocks/share/tarantool/vshard/router/init.lua:1325: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
	...-cartridge/.rocks/share/tarantool/vshard/router/init.lua:1325: in function 'info'
	...dge/.rocks/share/tarantool/cartridge/lua-api/boxinfo.lua:52: in function 'get_info'
	...e/.rocks/share/tarantool/cartridge/webui/gql-boxinfo.lua:330: in function 'defaultResolver'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:303: in function 'getFieldEntry'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:322: in function 'evaluateSelections'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:224: in function 'completeValue'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:213: in function 'completeValue'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:309: in function 'getFieldEntry'
	...-cartridge/.rocks/share/tarantool/graphql/execute.lua:322: in function <...-cartridge/.rocks/share/tarantool/graphql/execute.lua:314>
	...
	...k-cartridge/.rocks/share/tarantool/cartridge/graphql.lua:300: in function <...k-cartridge/.rocks/share/tarantool/cartridge/graphql.lua:226>
	[C]: in function 'xpcall'
	...-cartridge/.rocks/share/tarantool/errors.lua:145: in function 'pcall'
	...-cartridge/.rocks/share/tarantool/cartridge/graphql.lua:335: in function 'sub'
	...-cartridge/.rocks/share/tarantool/http/server.lua:658: in function <...l-cartridge/.rocks/share/tarantool/http/server.lua:636>
	[C]: in function 'pcall'
	...-cartridge/.rocks/share/tarantool/http/server.lua:736: in function 'process_client'
	...-cartridge/.rocks/share/tarantool/http/server.lua:1199: in function <...-cartridge/.rocks/share/tarantool/http/server.lua:1198>
	[C]: in function 'pcall'
	builtin/socket.lua:1081: in function <builtin/socket.lua:1079>

Serpentian:
This patch introduces protecting router's API while its configuration
is not done as accessing these functions at that time is not safe and
can cause low level errors like 'bad arguments' or 'no such function'.

Now all non-trivial vshard.router functions are disabled until
vshard.router.cfg (or vshard.router.new) is finished and error is
raised in case of an attempt to access them.

Manual API's enabling/disabling is also introduced in this patch.
The behavior of router's protection is similar to the storage's one.

Closes #291

### DIFF
--- a/test/instances/router.lua
+++ b/test/instances/router.lua
@@ -6,6 +6,7 @@ local helpers = require('test.luatest_helpers')
 -- able to use the libs in server:exec() calls and not get upvalue errors if the
 -- same lib is declared in the _test.lua file.
 --
+_G.ifiber = require('fiber')
 _G.imsgpack = require('msgpack')
 _G.ivtest = require('test.luatest_helpers.vtest')
 _G.iwait_timeout = _G.ivtest.wait_timeout
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -524,6 +524,7 @@ end
 
 --
 -- Create a new router in the cluster.
+-- If no cfg was passed configuration should be done manually with server:exec
 --
 local function router_new(g, name, cfg)
     if not g.cluster then
@@ -535,7 +536,9 @@ local function router_new(g, name, cfg)
     g[name] = server
     g.cluster:add_server(server)
     server:start()
-    router_cfg(server, cfg)
+    if cfg then
+        router_cfg(server, cfg)
+    end
     return server
 end
 
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -444,3 +444,104 @@ g.test_ssl = function(g)
     vtest.router_cfg(g.router, cluster_cfg)
     vtest.storage_wait_fullsync(g)
 end
+
+g.test_enable_disable = function(g)
+    --
+    -- gh-291: router enable/disable
+    --
+    local router = vtest.router_new(g, 'router_1')
+    -- do not allow router's configuration to complete
+    router:exec(function()
+        _G.ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = true
+    end)
+    router:exec(function(cfg)
+        rawset(_G, 'fiber_static', ifiber.new(ivshard.router.cfg, cfg))
+        rawset(_G, 'fiber_new', ifiber.new(ivshard.router.new,
+                                           'new_router', cfg))
+        _G.fiber_static:set_joinable(true)
+        _G.fiber_new:set_joinable(true)
+    end, {cluster_cfg})
+
+    -- emulate unconfigured box
+    router:exec(function()
+        rawset(_G, 'old_box_cfg', box.cfg)
+        box.cfg = function(...) return _G.old_box_cfg(...) end
+    end)
+
+    -- check whether errors are not nil and their messages are equal to str
+    local assert_errors_equals = function(err1, err2, str)
+        t.assert(err1 and err2)
+        t.assert_equals(err1.message, err2.message)
+        t.assert_str_contains(err1.message, str)
+    end
+
+    local err1, err2 = router:exec(function()
+        rawset(_G, 'static_router', ivshard.router.internal.routers._static_router)
+        rawset(_G, 'new_router', ivshard.router.internal.routers.new_router)
+        local _, err_1 = pcall(_G.static_router.info, _G.static_router)
+        local _, err_2 = pcall(_G.new_router.info, _G.new_router)
+        return err_1, err_2
+    end)
+    assert_errors_equals(err1, err2, 'box seems not to be configured')
+
+    -- set box status to loading
+    router:exec(function()
+        box.cfg = _G.old_box_cfg
+        rawset(_G, 'old_box_info', box.info)
+        box.info = {status = 'loading'}
+    end)
+
+    local echo_func = function()
+        return router:exec(function(timeout)
+            local echo = function(router)
+                return pcall(router.callrw, router, 1, 'echo', {1},
+                             {timeout = timeout})
+            end
+            local _, ret_val_1 = echo(_G.static_router)
+            local _, ret_val_2 = echo(_G.new_router)
+            return ret_val_1, ret_val_2
+        end, {vtest.wait_timeout})
+    end
+
+    err1, err2 = echo_func()
+    assert_errors_equals(err1, err2, 'instance status is "loading"')
+
+    -- restore proper box configuration
+    router:exec(function()
+        box.info = _G.old_box_info
+    end)
+
+    err1, err2 = echo_func()
+    assert_errors_equals(err1, err2, 'router is not configured')
+
+    -- unblock router's configuration and wait until it's finished
+    router:exec(function()
+        _G.ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = false
+        _G.fiber_static:join()
+        _G.fiber_new:join()
+    end)
+
+    -- finally a success
+    local ret1, ret2 = echo_func()
+    t.assert_equals(ret1, ret2)
+    t.assert_equals(ret1, 1)
+
+    -- manual api disabling and enabling
+    router:exec(function()
+        _G.static_router:disable()
+        _G.new_router:disable()
+    end)
+    err1, err2 = echo_func()
+    assert_errors_equals(err1, err2, 'router is disabled explicitly')
+
+    router:exec(function()
+        _G.static_router:enable()
+        _G.new_router:enable()
+    end)
+    ret1, ret2 = echo_func()
+    t.assert_equals(ret1, ret2)
+    t.assert_equals(ret1, 1)
+
+    -- we don't want this server to interfere with subsequent tests
+    g.router_1:drop()
+end
--- a/vshard/router/init.lua
+++ b/vshard/router/init.lua
@@ -39,6 +39,7 @@ if not M then
         ---------------- Common module attributes ----------------
         errinj = {
             ERRINJ_CFG = false,
+            ERRINJ_CFG_DELAY = false,
             ERRINJ_FAILOVER_CHANGE_CFG = false,
             ERRINJ_RELOAD = false,
             ERRINJ_LONG_DISCOVERY = false,
@@ -52,6 +53,8 @@ if not M then
         -- This counter is used to restart background fibers with
         -- new reloaded code.
         module_version = 0,
+        -- Flag whether box.info.status is acceptable.
+        is_loaded = false,
 
         ----------------------- Map-Reduce -----------------------
         -- Storage Ref ID. It must be unique for each ref request
@@ -95,6 +98,16 @@ local ROUTER_TEMPLATE = {
         -- when no timeout is specified.
         --
         sync_timeout = consts.DEFAULT_SYNC_TIMEOUT,
+        -- Flag whether router_cfg() is finished.
+        is_configured = false,
+        -- Flag whether the instance is enabled manually. It is true by default
+        -- for backward compatibility with old vshard.
+        is_enabled = true,
+        -- Reference to the function-proxy to most of the public functions. It
+        -- allows to avoid 'if's in each function by adding expensive
+        -- conditional checks in one rarely used version of the wrapper and no
+        -- checks into the other almost always used wrapper.
+        api_call_cache = nil,
 }
 
 local STATIC_ROUTER_NAME = '_static_router'
@@ -1211,6 +1224,9 @@ local function router_cfg(router, cfg, is_reload)
     if not is_reload then
         box.cfg(box_cfg)
         log.info("Box has been configured")
+        while M.errinj.ERRINJ_CFG_DELAY do
+            lfiber.sleep(0.01)
+        end
     end
     -- Move connections from an old configuration to a new one.
     -- It must be done with no yields to prevent usage both of not
@@ -1248,6 +1264,7 @@ local function router_cfg(router, cfg, is_reload)
     end
     discovery_set(router, vshard_cfg.discovery_mode)
     master_search_set(router)
+    router.is_configured = true
 end
 
 --------------------------------------------------------------------------------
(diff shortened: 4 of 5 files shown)


test/instances/router.lua
```diff
@@ -13,6 +13,7 @@ _G.vshard = {
     router = require('vshard.router'),
 }
 _G.ivshard = _G.vshard
+_G.ifiber = require('fiber')
```
Gerold103, line 16: I am going to make a moderately harsh review at first. Let me know if it is too complicated and you would prefer a light version after pushing which I will make some review fixes myself.

Here, please, put the require a bit higher. I am trying to keep them alphabetically sorted in this file and all new ones. It is already ruined in most of the older files, but here I want to make it right.

test/instances/router.lua
```diff
@@ -13,6 +13,7 @@ _G.vshard = {
...
```
Gerold103, line 17: In the commit message, please, make a docbot request which describes the user visible changes. The template can be found here https://github.com/tarantool/docbot. Examples - in `git log`.

test/luatest_helpers/vtest.lua
```diff
@@ -523,9 +523,17 @@ local function router_cfg(router, cfg)
 end
 
 --
--- Create a new router in the cluster.
+-- Apply the config on the given router inside fiber
 --
-local function router_new(g, name, cfg)
+local function router_fiber_cfg(router, cfg)
```
Gerold103, line 528: Lets not complicate the API too much and reuse the existing `router_new` function. In particular, lets make so that `router_new` won't call `router_cfg` if `cfg` is nil. Then you can make empty `router_new` and then call `router_cfg` in your test in a fiber (the fiber doesn't need to be on the router instance, it can be local to the test instance too - might be simpler).

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,72 @@ g.test_ssl = function(g)
     vtest.router_cfg(g.router, cluster_cfg)
     vtest.storage_wait_fullsync(g)
 end
+
+g.test_enable_disable = function(g)
+    --
+    -- gh-291: router enable/disable
+    --
+    local router_name = 'router_1'
```
Gerold103, line 452: Just inline the name, you use it only once anyway.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,72 @@ g.test_ssl = function(g)
...
+    local router = vtest.router_fiber_new(g, router_name, cluster_cfg,
```
Gerold103, line 453: If you follow my advice above about not having a dedicated `router_fiber_new`, then you also won't need this `pre_func` arg and can call it right in the test. Without bringing it up to vtest.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,72 @@ g.test_ssl = function(g)
...
+        function() ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = true end)
+    -- emulate unconfigured box
+    router:exec(function()
+        old_cfg = box.cfg
+        box.cfg = function(...) return old_cfg(...) end
+    end)
+
+    local status, ret = router:exec(function()
+            static_router = ivshard.router.internal.routers._static_router
+            return pcall(static_router.info, static_router)
```
Gerold103, line 463: It seems you have indentation step size 8 symbols. Please, set it to 4 (whitespaces, not tabs).

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,72 @@ g.test_ssl = function(g)
...
+                return pcall(static_router.callrw, static_router,
+                             1, 'echo', {1}, {timeout = timeout})
+        end, {wait_timeout})
+    end
+
+    status, ret = echo_func()
+    t.assert(not status and ret)
+    t.assert_str_contains(ret.message, 'instance status is "loading"')
+
+    -- retore proper box configuration
```
Gerold103, line 486: `retore` -> `restore`.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,72 @@ g.test_ssl = function(g)
...
+    router:exec(function()
+        box.info = old_info
+        while box.info.status ~= 'running' do fiber.sleep(0.1) end
```
Gerold103, line 489: Please, use `t.helpers.retrying` for such waitings.

test/luatest_helpers/vtest.lua
```diff
@@ -535,10 +543,39 @@ local function router_new(g, name, cfg)
...
+    if pre_func then
+        server:exec(pre_func, opts)
+    end
+    router_fiber_cfg(server, cfg)
+    return server
+end
+
+local function router_fiber_new_join(router)
+    router:exec(function()
+        if f then f:join() end
```
Gerold103, line 575: Does this mean you have a global variable called `f` on the instance? Lets not do that, especially in the generic vtest. You can declare such small common variables per-test-case. To avoid luacheck warnings, you can do it like that:
```Lua
server:exec(function()
    rawset(_G, 'myvar', 100)
end)
server:exec(function()
    assert(_G.myvar == 100)
end)
```

test/luatest_helpers/vtest.lua
```diff
@@ -535,10 +543,39 @@ local function router_new(g, name, cfg)
...
+    end)
+end
+
```
Gerold103, line 578: You might want to run luacheck locally. I don't remember how to install it - something about `luarocks` or `tarantoolctl rocks`. Anyway, I installed 100 years ago, run it it like that: `../tarantool/.rocks/bin/luacheck --codes --config .luacheckrc .` (I have `.rock` dir in my tarantool dir, which I have at the same level as vshard dir). Your commit produces these warnings:
```
Gerold103Macbook:vshard gerold$ ../tarantool/.rocks/bin/luacheck --codes --config .luacheckrc .
Checking test/luatest_helpers/vtest.lua           4 warnings

    test/luatest_helpers/vtest.lua:530:9: (W111) setting non-standard global variable f
    test/luatest_helpers/vtest.lua:531:9: (W113) accessing undefined variable f
    test/luatest_helpers/vtest.lua:575:12: (W113) accessing undefined variable f
    test/luatest_helpers/vtest.lua:575:19: (W113) accessing undefined variable f

Checking test/router-luatest/router_test.lua      15 warnings

    test/router-luatest/router_test.lua:454:20: (W112) mutating non-standard global variable ivshard
    test/router-luatest/router_test.lua:457:9: (W111) setting non-standard global variable old_cfg
    test/router-luatest/router_test.lua:458:40: (W113) accessing undefined variable old_cfg
    test/router-luatest/router_test.lua:462:13: (W111) setting non-standard global variable static_router
    test/router-luatest/router_test.lua:463:26: (W113) accessing undefined variable static_router
    test/router-luatest/router_test.lua:463:46: (W113) accessing undefined variable static_router
    test/router-luatest/router_test.lua:470:19: (W113) accessing undefined variable old_cfg
    test/router-luatest/router_test.lua:471:9: (W111) setting non-standard global variable old_info
    test/router-luatest/router_test.lua:477:30: (W113) accessing undefined variable static_router
    test/router-luatest/router_test.lua:477:52: (W113) accessing undefined variable static_router
    test/router-luatest/router_test.lua:488:20: (W113) accessing undefined variable old_info
    test/router-luatest/router_test.lua:489:47: (W113) accessing undefined variable fiber
    test/router-luatest/router_test.lua:498:9: (W112) mutating non-standard global variable ivshard
    test/router-luatest/router_test.lua:507:28: (W113) accessing undefined variable static_router
    test/router-luatest/router_test.lua:512:28: (W113) accessing undefined variable static_router
```

vshard/router/init.lua
```diff
@@ -1650,6 +1708,7 @@ local function router_new(name, cfg)
     end
     local router = table.deepcopy(ROUTER_TEMPLATE)
     setmetatable(router, router_mt)
+    router.api_call_cache = router_api_call_unsafe
```
Gerold103, line 1711: Please, add tests for `vshard.router.new()` instances using enable/disable too.

Gerold103 (changes requested):
Thanks a lot for the patch, finally a new contributor! See some of my first comments below. Lets deal with them and then proceed further.

test/instances/router.lua
Serpentian: Done.  As for the message, sorry for plagiarism but router's enabling/disabling logic is just the same as storage's one

test/luatest_helpers/vtest.lua
Serpentian: That's a great idea! Implemented.

test/luatest_helpers/vtest.lua
Serpentian: My bad, I've done that but only for sources. I stand corrected)

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,113 @@ g.test_ssl = function(g)
...
+
+    -- restore proper box configuration
+    router:exec(function()
+        box.info = _G.old
+    end)
+    -- throw the error inside function because t.helpers.retrying use pcall
+    -- in order to determine when to stop
+    local status = t.helpers.retrying({timeout = 10}, function()
+        return router:exec(function()
+            return box.info.status == 'running' and true
```
Gerold103, line 517: `and true` logically is a nop, lets drop this.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,113 @@ g.test_ssl = function(g)
...
+    err1, err2 = echo_func()
+    assert_errors_equals(err1, err2, 'instance status is "loading"')
+
+    -- restore proper box configuration
+    router:exec(function()
+        box.info = _G.old
+    end)
+    -- throw the error inside function because t.helpers.retrying use pcall
+    -- in order to determine when to stop
+    local status = t.helpers.retrying({timeout = 10}, function()
```
Gerold103, line 515: Please, use `timeout = vtest.wait_timeout`.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,113 @@ g.test_ssl = function(g)
...
+        return router:exec(function()
+            return box.info.status == 'running' and true
+                or error('box is not yet configured')
+        end)
+    end)
+    t.assert(status)
```
Gerold103, line 521: How can it be not running really? `router.lua` calls `box.cfg` unconditionally. And above you restored `box.info` to its normal value. That means the status became `running` immediately after that.

test/router-luatest/router_test.lua
```diff
@@ -444,3 +444,113 @@ g.test_ssl = function(g)
...
+        local _, err_1 = pcall(_G.static_router.info, _G.static_router)
+        local _, err_2 = pcall(_G.new_router.info, _G.new_router)
+        return err_1, err_2
+    end)
+    assert_errors_equals(err1, err2, 'box seems not to be configured')
+
+    -- set box status to loading
+    router:exec(function()
+        box.cfg = _G.old
+        _G.old = box.info
```
Gerold103, line 490: Lets give a bit more meaningful names. For example, `_G.old_box_info`, `_G.old_box_cfg`.

vshard/util.lua
```diff
@@ -313,6 +313,45 @@ local feature = {
...
+    return (instance_type == "storage" and
+            lerror.vshard(lerror.code.STORAGE_IS_DISABLED, msg) or
+            lerror.vshard(lerror.code.ROUTER_IS_DISABLED, msg))
+end
+
+--
+-- Check if box and storage/router has finished their configuration
+-- and if instance's API wasn't disabled explicitly
+--
+local function check_api_status(instance, is_loaded, instance_type)
```
Gerold103, line 326: Tbh I would just leave this function inlined in the 2 init.lua files. The reason is that when it is factored out like that, it looks too entangled with magic string constants and members of `instance` object which are not visible nor documented in the function's declaration comment. Even more confusion comes from the necessity to give `is_loaded` explicitly while the other `is_...` are taken from `instance`. I know why it is done like that but still the function looks not too good.

So I would suggest to either
- Inline its usage in init.lua files;
- Or find a way to make it less dependent on `instance` object and its type name. For example, make it accept a single argument `status` which would be `{is_loaded = ..., is_configured = ..., is_enabled = ..., name = ..., errcode = ...}`. The function would check all its stuff like now, update the `status` object, and return `true` or `nil, err` with the given `status.errcode`. `True` means all is fine. `nil, err` means you need to raise the error. And you can fetch updated `is_loaded` from the `status`. Then util.lua won't depend on storage or router neither in names nor in error codes;
- Or factor out only certain parts of the function. For instance `util.box_is_configured`, `util.box_is_running`. And the rest stays in init.lua files. Then code duplication will be insignificant.

Another reason why I would rather recommend to duplicate a few lines than try to extract it in a nice way is that the router shouldn't really depend on `box.cfg` at all. It was a mistake that it calls `box.cfg` internally. And eventually we will most likely drop it along with these `box.info.status` and `type(box.cfg)` checks. Which means extracting them into some common place probably is not needed.

Gerold103 (changes requested):
Hi! Thanks for the fixes! Getting better, now some small polish points.

Gerold103:
Looks like your commit also closes #194 - lets add `Closes` for it to the commit message.

vshard/util.lua
Serpentian: In the light of what you have told about removing `box.cfg` from the router in the future inlining seems to be the best solution. After removing there will be only one similar part in `*_api_call_unsafe` concerning `is_configured` and `is_enabled`, which is not very rational to put into common part.

Gerold103 (approved):
Thanks for working on this!

PR #348, opened 2022-07-05
Bucket ref protection on replicas - part 2 - recovery makes SENT buckets, GC tests to luatest

Gerold103:
The patchset is a next step on the way of fixing the buckets being deleted on master while still having RO refs on replicas.

Firstly, it makes bucket recovery turn SENDING buckets to SENT in case the transfer was successful. That is done because GARBAGE and SENT statuses will have more difference in the future.

Secondly, the bucket garbage tests are ported to luatest in order to change and extend them later.

Thirdly, it fixes wrong usage of `index:min()` at 1.10. It exploded on the last commit when `min(key)` returned something `> key`.

#173


test/instances/default.lua
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env tarantool
```
olegrok, line 1: Somewhy I thought it works in the same way even for 1.10...

You can also add this issue to commit message - https://github.com/tarantool/tarantool/issues/3167

vshard/util.lua
```diff
@@ -286,6 +286,37 @@ local function fiber_is_self_canceled()
...
+end
+
+end
+
+--
+-- Check if the index has any tuple matching the key.
+--
+local function index_has(idx, key)
+    return index_min(idx, key) ~= nil
+end
```
olegrok, line 318: I guess it could be done with `index:count(key) ~= 0` (that is a bit GC-friendlier).

olegrok (commented):
Thanks for your patch.
Looks good, see two comments below.

test/instances/default.lua
Gerold103: Done.

PR #349, opened 2022-07-07
test: fix flaky rebalancer/stress_add_remove_rs.test.lua

Serpentian:
Sometimes `rebalancer/stress_add_remove_rs.test.lua` hangs. This happens on both storage engines. We need to fix that.


Serpentian:
The issue was solved with 76b88963e1e89a1bd60de0e5c910c211574bc722

Issue #309, opened 2021-12-30
flaky test: rebalancer/stress_add_remove_rs.test.lua (on memtx) hangs


kyukhin:
The log can be seen [here](https://github.com/tarantool/tarantool/runs/4664663482?check_suite_focus=true#step:7:68).

Totktonada:
The logs in CI expire after some time (don't remember precisely, several months), so I suggest to attach them into issues.

[logs.txt](https://github.com/tarantool/vshard/files/7794831/logs.txt)

Gerold103:
Hangs on vinyl too sometimes. The reason is that `rebalancer_utils.start_loading()`, if called more than once during one test run, creates duplicate primary keys. Because it uses `write_iterations` as pk, which is reset to 0 on each `start_loading()`. A good-enough fix could be to truncate the test space before doing next `start_loading()`.

Also as a safety measure it is worth limiting `write_iterations` somehow. For example, make it max `N * bucket_count` and increment by modulo of this value. Then it is limited and it won't produce duplicate pks.

Serpentian:
However, I could do something wrong about truncating spaces. Here's the draft PR with truncating: #349. Unfortunately, this test passed on CI this time, but if you restart it several times it will fail some day.

Gerold103:
> Moreover, why do we bother about duplicated primary keys when in do_write_load we call do_replace.

If you start load, then stop it, and start again, then you will start with write_iterations 0 again. That causes `replace`s of the already generated primary keys into potentially other buckets. So you get tuples like `{1, 10}` and `{1, 20}` in buckets 10 and 20 on different replicasets. But when rebalancer tries to put buckets 10 and 20 onto a same replicaset, they fail with the duplicate primary key `{1}` error. I can definitely say this was one of the reasons why the test fails, because I saw in the logs an error about duplicate primary key when it clashed from 2 different buckets. Doesn't mean though that this is the only reason.

> The problem is definitely in this cancelled fiber on storage.

It doesn't necessarily mean it was cancelled on the storage. More likely it was cancelled on the router. The netbox internal fiber, for instance. Something could go wrong inside of it.

Serpentian:
Most likely, the issue is related to #341

Gerold103:
I wouldn't be so sure. In #341 they kill all the fibers explicitly. In the tests we don't do that.

Gerold103:
Sorry, truncation was never needed. I figured out the reason, See the attached PR above.

PR #354, opened 2022-07-08
Fix rebalancer stress tests flakiness

Gerold103:
See the commit messages. Here is an example of what I am talking about in the first commit:
```Lua
net = require('net.box')
fiber = require('fiber')
box.cfg{listen = 3313}
box.schema.user.grant('guest', 'super')
count = 0
c = net.connect(box.cfg.listen)
c:eval('return 1')

f = fiber.new(function()
    fiber.self():set_joinable(true)
    local cancel_count = 10
    while pcall(fiber.testcancel) do
        pcall(fiber.sleep, 0.01)
    end
    return pcall(c.eval, c, 'count = count + 1')
end)

fiber.sleep(0.1)
f:cancel()
f:join()
```
Then print `count` and you will see `1`. The request in a cancelled fiber was saved into netbox's buffer. Only the waiting for response failed. But the request was sent.

I run both stress tests 60 times locally, the error seems to be gone from both.

#309

### DIFF
--- a/test/rebalancer/rebalancer_utils.lua
+++ b/test/rebalancer/rebalancer_utils.lua
@@ -6,12 +6,16 @@ local read_iterations = 0
 local write_fiber = 'none'
 local read_fiber = 'none'
 local bucket_count = 200
+-- Set just any limit to ensure it is at least not infinite. We want our disk to
+-- have space even if the test runs too long somewhy.
+local primary_key_max = 10000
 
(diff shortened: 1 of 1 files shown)


olegrok (approved):

Serpentian (approved):
Thanks for your patch! That's genius! I've already started digging into the internals of net.box in search of a problem for this issue)

PR #356, opened 2022-07-14
replicaset: reconnect after fiber kill

### LINKED ISSUE #341: The case when storage and router iproto fiber is cancelled
Privet

There is case when something happened and storage fiber is cancelled (for e.g. cartridge hotreload or any other fiber killer).

Some affected snippet
```lua
vshard = require('vshard')
netbox = require('net.box')

cfg = {
    memtx_memory = 100 * 1024 * 1024,
    bucket_count = 3,
    rebalancer_disbalance_threshold = 10,
    rebalancer_max_receiving = 100,
    sharding = {
        ['cbf06940-0790-498b-948d-042b62cf3d29'] = {
            replicas = {
                ['8a274925-a26d-47fc-9e1b-af88ce939412'] = {
                    uri = 'storage:storage@127.0.0.1:3301',
                    name = 'storage_1_a',
                    master = true
                },
            },
        },
    },
}

vshard.storage.cfg(cfg, '8a274925-a26d-47fc-9e1b-af88ce939412')
box.schema.user.grant('storage', 'super', nil, nil, {if_not_exists=true})

vshard.router.cfg(cfg)
vshard.router.bootstrap()

local log = require('log')
local fiber = require('fiber')
rc, err = vshard.router.callrw(1, 'box.info')
assert(rc ~= nil)
log.info(rc)
--log.info(fiber.info())

c = netbox.connect('127.0.0.1:3301', {user="storage", password="storage"})
log.info('before netbox call')
log.info(c:call('box.info'))

for id, f in pairs(fiber.info()) do 
    if f.name:endswith('(net.box)') then
        fiber.kill(fiber.find(id))
    end
end

rc, err = vshard.router.callrw(1, 'box.info')
assert(rc == nil)
log.info(rc)

rc, err = vshard.router.callrw(1, 'box.info')
assert(rc == nil)

log.info('after netbox call')
local rc, res, err = pcall(c.call, c, {'box.info'})
if rc ~= true then
    log.info(res)
end

c = netbox.connect('127.0.0.1:3301', {user="storage", password="storage"})
log.info('after netbox call with reloaded connection')
log.info(c:call('box.info'))

package.loaded['vshard'] = nil
local vshard = require('vshard')
rc, err = vshard.router.callrw(1, 'box.info')
log.info(rc)
assert(rc ~= nil, tostring(err))

require('console').start() os.exit(0)
```

The question is, how to restart netbox connection under vshard.router? Or is it possible to be done on vshard side?

Serpentian:
Currently if we kill net.box's fibers the connection goes into
`error_reconnect` state. However, it's not reconnecting anymore.

This patch introduces reconnecting in that case. It should be used
wisely, though. Fiber's killing doesn't happen instantly and if the
user doesn't wait util fiber's status is `dead` and makes the request
immediately, exception will be probably thrown as the fiber can die
in the middle of request.

So, after fiber kill wait until it's really dead and make a request
only after that.

Closes #341

### DIFF
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -7,7 +7,7 @@ local uuid = require('uuid')
 local yaml = require('yaml')
 local vrepset = require('vshard.replicaset')
 
-local wait_timeout = 120
+local wait_timeout = 50
 -- Use it in busy-loops like `while !cond do fiber.sleep(busy_step) end`.
 local busy_step = 0.005
 local uuid_idx = 1
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -545,3 +545,34 @@ g.test_enable_disable = function(g)
     -- we don't want this server to interfere with subsequent tests
     g.router_1:drop()
 end
+
+g.test_explicit_fiber_kill = function(g)
+    local bids = vtest.cluster_exec_each_master(g, function()
+        return _G.get_first_bucket()
+    end)
+
+    -- Kill worker fibers of connections to masters
+    g.router:exec(function()
+        for _, rs in ipairs(ivshard.router.static.replicasets) do
+            -- Explicitly kill the connection's worker fiber.
+            -- Callback for pushes is executed inside this fiber.
+            pcall(function()
+                rs.master.conn:eval([[
+                    box.session.push(nil)
+                ]], {}, {on_push = function()
+                    ifiber.self():cancel()
+                end})
+            end)
+        end
+    end)
+
+    -- Check that the replicaset is still accessible
+    g.router:exec(function(bids)
+        for _, bid in ipairs(bids) do
+            local res, err = ivshard.router.callrw(bid, 'echo', {1},
+                                                   {timeout = iwait_timeout})
+            ilt.assert(res, 1)
+            ilt.assert(err, nil)
+        end
+    end, {bids})
+end
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -145,6 +145,40 @@ local function netbox_wait_connected(conn, timeout)
 end
 
 --
+-- Check if the connection is dead and it won't be restored automatically.
+-- Even though replicaset's connections are initialized with the option
+-- 'reconnect_after', there's cases, when no reconnect will be done (e.g.
+-- if the conn was explicitly cancelled or the connection's worker fiber
+-- was killed). In these situations we need to reestablish it manually as
+-- the missing connection to some replicaset is a critical problem which
+-- leads to the inability of the user to access some part of a data via the
+-- router.
+--
+-- Note: the function expects the conn to be non-nil.
+--
+local function netbox_is_conn_dead(conn)
+    -- Fast path - conn is almost always 'active'.
+    if conn.state == 'active' then
+        return false
+    end
+    if conn.state == 'error' or conn.state == 'closed' then
+        return true
+    end
+    if conn.state ~= 'error_reconnect' then
+        -- The connection is fetching schema or doing auth, which
+        -- means it is not dead and shoudn't be reinitialized.
+        return false
+    end
+    if not conn._fiber then
+        -- The fiber field is not present in old netbox, which correctly
+        -- reports state as 'error' when its fiber got killed. It would
+        -- be filtered out above then.
+        return false
+    end
+    return conn._fiber:status() ~= "dead"
+end
+
+--
 -- Check if the replica is not in backoff. It also serves as an update - if the
 -- replica still has an old backoff timestamp, it is cleared. This way of
 -- backoff update does not require any fibers to perform background updates.
@@ -170,7 +204,7 @@ end
 --
 local function replicaset_connect_to_replica(replicaset, replica)
     local conn = replica.conn
-    if not conn or conn.state == 'closed' then
+    if not conn or netbox_is_conn_dead(conn) then
         conn = netbox.connect(replica.uri, {
             reconnect_after = consts.RECONNECT_TIMEOUT,
             wait_connected = false
--- a/test/storage/recovery.result
+++ b/test/storage/recovery.result
@@ -122,6 +122,9 @@ while _bucket:count() ~= 2 do vshard.storage.recovery_wakeup() fiber.sleep(0.1) 
 _ = test_run:switch('storage_1_a')
 ---
 ...
+vshard.storage.internal.errinj.ERRINJ_RECOVERY_PAUSE = true
+---
+...
 _bucket:replace{1, vshard.consts.BUCKET.SENDING, util.replicasets[2]}
 ---
 - [1, 'sending', '<replicaset_2>']
--- a/test/storage/recovery.test.lua
+++ b/test/storage/recovery.test.lua
@@ -58,6 +58,7 @@ while _bucket:count() ~= 2 do vshard.storage.recovery_wakeup() fiber.sleep(0.1) 
 -- must restore buckets, when the destination is up.
 --
 _ = test_run:switch('storage_1_a')
+vshard.storage.internal.errinj.ERRINJ_RECOVERY_PAUSE = true
 _bucket:replace{1, vshard.consts.BUCKET.SENDING, util.replicasets[2]}
 _ = test_run:switch('storage_2_a')
 _bucket:replace{1, vshard.consts.BUCKET.ACTIVE}


vshard/replicaset.lua
```diff
@@ -170,7 +170,9 @@ end
 --
 local function replicaset_connect_to_replica(replicaset, replica)
     local conn = replica.conn
-    if not conn or conn.state == 'closed' then
+    -- gh-341: additionally check if connection's fiber was explicitly killed
+    if not conn or conn.state == 'closed' or (conn.state == 'error_reconnect'
+       and conn._fiber and conn._fiber:status() == 'dead') then
```
olegrok, line 175: Are there always such thing as conn._fiber?
```lua
tarantool> c
---
- peer_protocol_version: 3
  peer_uuid: a4c86733-c191-49f1-a2a4-1fa916306e3c
  peer_protocol_features:
    transactions: true
    watchers: true
    error_extension: true
    streams: true
  schema_version: 78
  protocol: Binary
  state: active
  peer_version_id: 133632
  port: '3301'
...

tarantool> c._fiber
---
- null
...

tarantool> _TARANTOOL
---
- 2.10.0-3-gea3ae031e-r493
...
```

In case if fiber is killed connection state is changed to "error":
```lua
tarantool> fiber.find(118)
---
- status: suspended
  name: :3301 (net.box)
  id: 118
...

tarantool> f = fiber.find(118)
---
...

tarantool> f:cancel()
---
...

tarantool> f
---
- the fiber is dead
...

tarantool> c
---
- peer_protocol_version: 3
  peer_uuid: a4c86733-c191-49f1-a2a4-1fa916306e3c
  peer_protocol_features:
    transactions: true
    watchers: true
    error_extension: true
    streams: true
  schema_version: 78
  protocol: Binary
  state: error
  error: fiber is cancelled
  peer_version_id: 133632
  port: '3301'
...
```

Serpentian: Yep, there's not always _fiber in connection. I've already fixed that. Speaking of state, as we open connection with `reconnect_after` option the state will be `error_reconnect`. 

Test doesn't fail locally. I'll investigate that thing. The PR is definitely not ready for review.
Sorry for that

olegrok: I'd say that it's a bit strange to rely on netbox internals. It's not documented at could be broken at the any moment (e.g. when https://github.com/tarantool/tarantool/issues/7291 mentioned in https://github.com/tarantool/tarantool/commit/0d944f901cf5400d5832a5528780bb3ecd03ae4b will be fixed).

So but as I showed above nebox with killed worker fiber has
```
  state: error
  error: fiber is cancelled
```

And probably this information could be used. It works for Tarantool 2.10, but for earlier versions it should be checked.

Serpentian: > as we open connection with `reconnect_after` option the state will be `error_reconnect`

I was wrong. This is only in the latest releases. E.g. in tarantool 2.6:
```lua
tarantool> c = require('net.box').connect('127.0.0.1:3301', {reconnect_after = 0.1})
---
...

tarantool> c
---
- peer_uuid: 5f56e0a5-1a38-4add-a63b-c139596a2a20
  opts:
    reconnect_after: 0.1
  host: 127.0.0.1
  schema_version: 78
  protocol: Binary
  state: active
  peer_version_id: 132608
  port: '3301'
...

-- <...>: fiber kill

tarantool> c
---
- peer_uuid: 5f56e0a5-1a38-4add-a63b-c139596a2a20
  opts:
    reconnect_after: 0.1
  host: 127.0.0.1
  schema_version: 78
  protocol: Binary
  state: error
  error: fiber is cancelled
  peer_version_id: 132608
  port: '3301'
...

tarantool> _TARANTOOL
---
- 2.6.0-0-g47aa4e01e
...
```

So, we should also check if conn:state() equals to `error`, not only `error_reconnect`.

Serpentian:
Looks like the tests hangs on some runners. We should kill fibers more accurately and not just all net.box's ones. Will implement this tomorrow.

Serpentian:
Actually, the problem was not about killing fibers. It was due to the the fact that in 2.7 - 2.9 `conn.error` has table type and we have to use `conn.error.message` in order to get that msg. In 1.10 and 2.10 `conn.error` is just a string.

vshard/util.lua
```diff
@@ -286,6 +286,15 @@ local function fiber_is_self_canceled()
     return not pcall(fiber.testcancel)
 end
 
+--
+-- Check if connection's fiber was explicitly killed
+--
+local conn_fiber_killed = function(conn)
+    local msg = 'fiber is cancelled'
+    return (conn.state == 'error_reconnect' or conn.state == 'error') and
```
olegrok, line 294: Probably you can avoid string comparison here...

Could `conn.state == 'error_reconnect' or conn.state == 'error')` be replaced with `conn.error ~= nil`?
Also probably error sometimes not table but `box.error` that has `code`. And comparison of codes is faster than string comparison.

Serpentian: Unfortunately, they don't have code. In 2.7 - 2.10 `conn.error` is indeed not a table but a cdata. However, it's not a `box.error` but an exception, described in lib/core/exception.cc:
```C
static const struct method_info exception_methods[] = {
	make_method(&type_Exception, "message", &Exception::get_errmsg),
	make_method(&type_Exception, "log", &Exception::log),
	METHODS_SENTINEL
};
const struct type_info type_Exception = make_type("Exception", NULL,
	exception_methods);

<...>

const struct type_info type_FiberIsCancelled =
	make_type("FiberIsCancelled", &type_Exception);

FiberIsCancelled::FiberIsCancelled(const char *file, unsigned line)
	: Exception(&type_FiberIsCancelled, file, line)
{
	error_format_msg(this, "fiber is cancelled");
}

```
It has nothing inside it but this message. That's probably the reason, why now this structure is not used as an error and is replaced by the message. So, we have to compare error's strings if we want this issue to be solved from vshard. 
However, it's not that bad as in case of replacing `conn.state == 'error_reconnect' or conn.state == 'error'` with just `conn.error` strings won't be compared in most cases.

olegrok: > Unfortunately, they don't have code. In 2.7 - 2.10 conn.error is indeed not a table but a cdata. However, it's not a box.error but an exception, described in lib/core/exception.cc:

exception itself can't be lua object (even cdata). There is wrapper around such errors - https://github.com/tarantool/tarantool/blob/master/src/lua/error.lua

Serpentian: It will be anyway comparison of strings (either it will be `base_type`, `type`, or `message`):

<details>
<summary>Tarantool  2.6 </summary>

```lua
tarantool> _TARANTOOL
---
- 2.6.0-0-g47aa4e01e
...
tarantool> c
---
- peer_uuid: 98d4e7e7-2602-4ce3-857d-82fcda96eb40
  host: 127.0.0.1
  schema_version: 78
  protocol: Binary
  state: error
  error: fiber is cancelled
  peer_version_id: 132608
  port: '3301'
...
tarantool> c.error:unpack()
---
- base_type: FiberIsCancelled
  type: FiberIsCancelled
  message: fiber is cancelled
  trace:
  - file: ./src/lua/fiber.c
    line: 46
...
```
</details>

<details>
<summary>Tarantool  2.9 </summary>

```lua
tarantool> _TARANTOOL
---
- 2.9.0-0-g15bbdab92
...
tarantool> c
---
- peer_uuid: 59b9898b-be3a-440b-b385-6224472d1894
  host: 127.0.0.1
  schema_version: 78
  protocol: Binary
  state: error
  error: fiber is cancelled
  peer_version_id: 133376
  port: '3301'
...
tarantool> c.error:unpack()
---
- base_type: FiberIsCancelled
  type: FiberIsCancelled
  message: fiber is cancelled
  trace:
  - file: /build/tarantool/src/tarantool/src/lua/fiber.c
    line: 46
...
```
</details>

vshard/util.lua
```diff
@@ -286,6 +286,14 @@ local function fiber_is_self_canceled()
     return not pcall(fiber.testcancel)
 end
 
+--
+-- Check if connection's fiber was explicitly killed
+--
+local conn_fiber_killed = function(conn)
```
olegrok, line 292: Well, probably approach is seems to be OK.
Probably you should rename function to `is_conn_fiber_killed` to conform tarantool coding conversion.

vshard/replicaset.lua
```diff
@@ -170,7 +170,8 @@ end
 --
 local function replicaset_connect_to_replica(replicaset, replica)
     local conn = replica.conn
-    if not conn or conn.state == 'closed' then
+    -- gh-341: additionally check if connection's fiber was explicitly killed
+    if not conn or conn.state == 'closed' or util.conn_fiber_killed(conn) then
```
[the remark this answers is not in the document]
olegrok: Well, let's stay as is

olegrok (approved):

vshard/replicaset.lua
```diff
@@ -170,7 +170,8 @@ end
 --
 local function replicaset_connect_to_replica(replicaset, replica)
     local conn = replica.conn
-    if not conn or conn.state == 'closed' then
+    -- gh-341: additionally check if connection's fiber was explicitly killed
+    if not conn or conn.state == 'closed' or util.is_conn_fiber_killed(conn) then
```
Gerold103, line 174: Lets turn the check to be faster on the happy-path. I.e. if the connection is fine and dandy, then you shouldn't check an error message. It is done on each request, hence we want it to be as quick as possible in most cases:
```Lua
if conn and conn.state == 'active' then
    return conn
end
-- Do more rigorous checks
-- ...
```

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,44 @@ g.test_enable_disable = function(g)
...
+        for id, f in pairs(ifiber.info()) do
+            if f.name:endswith('(net.box)') then
+                ifiber.kill(id)
+            end
+        end
+
+        local replicasets = ivshard.router.static.replicasets
+        local conn_1 = replicasets[uuids[1]].master.conn
+        local conn_2 = replicasets[uuids[2]].master.conn
+        assert(conn_1 and conn_2)
```
Gerold103, line 567: Please, use `ilt.assert`.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,44 @@ g.test_enable_disable = function(g)
...
+
+        while ivutil.is_conn_fiber_killed(conn_1) or
```
Gerold103, line 569: Lets use `ilt.helpers.retrying` with a finite timeout (`iwait_timeout`).

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,44 @@ g.test_enable_disable = function(g)
...
+    --
+    local rs_uuids = {g.replica_1_a:replicaset_uuid(),
+                      g.replica_2_a:replicaset_uuid()}
+
+    g.router:exec(function(uuids)
+        for id, f in pairs(ifiber.info()) do
+            if f.name:endswith('(net.box)') then
+                ifiber.kill(id)
+            end
+        end
```
Gerold103, line 562: I am afraid we might accidentally kill something important which is not a part of vshard and won't have that kind of protection against dead netbox fibers. For example, we could damage the always fragile luatest. I would recommend to kill just the needed fibers. You can do it like that:
```Lua
conn:eval([[
    box.session.push(nil)
]], {}, {on_push = function()
    ifiber.self():cancel()
end})
```
Here I sent a request which returns 2 responses - one "push" and the final result (which is nil here). For pushes we can install callbacks which are executed in the same fiber as netbox. It allows you to get to the worker fiber and kill it directly.

Gerold103 (changes requested):
Thanks for the patch, good stuff!

Serpentian:
The test `storage/recovery` hangs on git version of tarantool. It doesn't seem to be related to this commits. I tried to push without decreasing of `wait_timeout` and it's just the same. Will try to investigate this tomorrow.

test/luatest_helpers/vtest.lua
```diff
@@ -7,7 +7,7 @@ local uuid = require('uuid')
 local yaml = require('yaml')
 local vrepset = require('vshard.replicaset')
 
-local wait_timeout = 120
+local wait_timeout = 50
```
Serpentian, line 10: I do not insist on that change. However, on luatest versions of tests there's no output in case of hang due to timeout loop.

vshard/replicaset.lua
```diff
@@ -144,6 +144,70 @@ local function netbox_wait_connected(conn, timeout)
...
+        return true
+    end
+
+    if conn.state ~= 'error_reconnect' then
+        -- The connection is fetching schema or doing auth, which
+        -- means it is not dead and shoudn't be reinitialized.
+        return false
+    end
+
+    return conn.error == 'fiber is cancelled'
```
Serpentian, line 206: AFAICS, this is the only way we can filter out dead connection on buggy versions. It can still be triggered false-positively, but it seems to be the better option rather than ignoring these versions.

Serpentian:
It is related to these commits. I rerunned git version on another PR and it's all right. Will investigate this tomorrow.

vshard/util.lua
```diff
@@ -286,6 +286,14 @@ local function fiber_is_self_canceled()
     return not pcall(fiber.testcancel)
 end
 
+--
+-- Check if connection's fiber was explicitly killed
+--
+local is_conn_fiber_killed = function(conn)
+    local msg = 'fiber is cancelled'
+    return conn.error and (conn.error == msg or conn.error.message == msg)
```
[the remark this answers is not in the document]
Serpentian: The solution, you proposed, will work everywhere except the versions range, which starts from the time net.box was ported to C. The state of the connection in these versions is `error_reconnect` and there's no `_fiber` in it. 

The only way we can figure out that the fiber is dead is checking error message, unfortunately. Do we want that? False-positive triggering of reinitialization of connection seems to be a better option rather than just saying that reestablishing won't work on these version.  

Speaking of the earlier versions, we need nothing but checking if `conn.state` is `error`.

vshard/replicaset.lua
Serpentian: Done

vshard/util.lua
```diff
@@ -286,6 +286,14 @@ local function fiber_is_self_canceled()
     return not pcall(fiber.testcancel)
 end
 
+--
+-- Check if connection's fiber was explicitly killed
```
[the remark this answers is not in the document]
Serpentian: Done

vshard/util.lua
```diff
@@ -286,6 +286,14 @@ local function fiber_is_self_canceled()
...
+--
+local is_conn_fiber_killed = function(conn)
```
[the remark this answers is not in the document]
Serpentian: Fixed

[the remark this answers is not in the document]
Serpentian: Yep, sounds reasonable.

test/router-luatest/router_test.lua
Serpentian: Great hack

Serpentian:
The test fails on git version. It  will be fixed as soon as https://github.com/tarantool/tarantool/pull/7468 is merged

test/luatest_helpers/vtest.lua
Gerold103: I couldn't understand this part: `Moreover, it allows the timeout loop to be executed at least 2 times. If the test check the returned value more often than after 3 timeout loop, it will definitely notice the wrong value.` but ok anyway, agreed.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,41 @@ g.test_enable_disable = function(g)
     -- we don't want this server to interfere with subsequent tests
```
Gerold103, line 545: Apparently CI is failing on the new test. I see the run on the latest master is red.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,41 @@ g.test_enable_disable = function(g)
...
     g.router_1:drop()
```
Gerold103, line 546: I would omit this part *"However, from 2.10.1 worker fiber is invincible and cannot be killed at all"* from the commit message until the fix is actually submitted. We can't rely on it until it is really done.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,41 @@ g.test_enable_disable = function(g)
...
+    -- Check that the replicaset is still accessible
+    local res = g.router:exec(function(bids)
+        local res = {}
+        for i, bid in ipairs(bids) do
+            res[i] = ivshard.router.callrw(bid, 'echo', {1},
+                                          {timeout = iwait_timeout})
+        end
+        return res
+    end, {bids})
+    t.assert(res[1] == 1 and res[2] == 1)
```
Gerold103, line 584: You can simplify the test just a little:
```Diff
@@ -547,39 +547,30 @@ g.test_enable_disable = function(g)
 end
 
 g.test_explicit_fiber_kill = function(g)
-    local masters = {g.replica_1_a, g.replica_2_a}
-    local rs_uuids, bids = {}, {}
-    for _, master in ipairs(masters) do
-        table.insert(rs_uuids, master:replicaset_uuid())
-        table.insert(bids, vtest.storage_first_bucket(master))
-    end
-
+    local bids = vtest.storage_exec_each_master(g, function()
+        return _G.get_first_bucket()
+    end)
     -- Kill worker fibers of connections to masters
-    g.router:exec(function(rs_uuids)
-        local replicasets = ivshard.router.static.replicasets
-        for _, rs_uuid in ipairs(rs_uuids) do
-            local conn = replicasets[rs_uuid].master.conn
-
+    g.router:exec(function()
+        for rs_uuid, rs in pairs(ivshard.router.static.replicasets) do
             -- Explicitly kill the connection's worker fiber.
             -- Callback for pushes is executed in it.
             pcall(function()
-                conn:eval([[
+                rs.master.conn:eval([[
                     box.session.push(nil)
                 ]], {}, {on_push = function()
                     ifiber.self():cancel()
                 end})
             end)
         end
-    end, {rs_uuids})
-
+    end)
     -- Check that the replicaset is still accessible
-    local res = g.router:exec(function(bids)
-        local res = {}
-        for i, bid in ipairs(bids) do
-            res[i] = ivshard.router.callrw(bid, 'echo', {1},
-                                          {timeout = iwait_timeout})
+    g.router:exec(function(bids)
+        for _, bid in ipairs(bids) do
+            local res, err = ivshard.router.callrw(bid, 'echo', {1},
+                                                   {timeout = iwait_timeout})
+            ilt.assert_equals(err, nil)
+            ilt.assert_equals(res, 1)
         end
-        return res
     end, {bids})
-    t.assert(res[1] == 1 and res[2] == 1)
 end
```

vshard/replicaset.lua
```diff
@@ -144,6 +144,49 @@ local function netbox_wait_connected(conn, timeout)
     return timeout
 end
 
+local netbox_is_conn_dead
+
+if not (util.version_is_at_least(2, 10, 0, 'beta', 2, 0) and
+        util.version_is_at_most(2, 10, 0, nil, 0, 0)) then
```
Gerold103, line 150: Could you please elaborate what version range you are trying to catch here? I suppose this should get the versions where cancellation was handled correct, right? After inlining it I got `vers < 2.10.0-beta2 or ver > 2.10.0`. It also catches `2.11.0-entrypoint` which is broken and where it is not enough to check `conn.state`.

Right now we don't even know when the bug will be fixed. IMO I would be fine to make the version-agnostic checks like this:
```Lua
if conn.state == 'error' then
    return true
end
if conn.state ~= 'error_reconnect' then
    -- Neither of error states should mean it is fetching schema or doing auth,
    -- which means it is not dead.
    return false
end
if not conn._fiber then
    -- The fiber field is not present in old netbox, which correctly reports
    -- state as 'error' when its fiber got killed. It would be filtered out
    -- above then.
    return false
end
return conn._fiber:status() ~= "dead"
```
and just let the other versions fail. It isn't too critical, I think. People shouldn't kill netbox fibers in the first place. Up to you again. If you want, we can try to keep pursuing the version-specific checks. Then when we fix the bug in tnt master, we will make a follow-up patch to handle that version without the false-positive case.

vshard/replicaset.lua
```diff
@@ -144,6 +144,49 @@ local function netbox_wait_connected(conn, timeout)
...
+else -- 2.10.0
+
+--
+-- Tarantool 2.10.0 has a bug: after worker fiber kill connection's state is
+-- error_reconnect, but it doesn't reconnet anymore. The only way we can
+-- catch this situation is error string comparison as the connection doesn't
+-- have _fiber in itself.
+--
+netbox_is_conn_dead = function(conn)
+    if conn.state == 'error' or conn.state == 'closed' then
```
Gerold103, line 177: I would suggest to move `if conn.state == 'active' then return true ... ` right here. Otherwise you force the function's user to explicitly check the connection's state before calling `netbox_is_conn_dead()` if he wants a quick path for the common case.

I know I proposed myself to check for `'active'` state explicitly outside, but now when I look at it, I think better make `netbox_is_conn_dead()` have the happy path as well. Up to you.

Gerold103 (commented):
Thanks for the fixes, looking better!

test/router-luatest/router_test.lua
Serpentian: Yep, everything was done with the intention of merging this PR after the appearance of the system fibers in tarantool. Curently if we wanted to keep pursuing the version-specific checks we would have to introduce three (or two but with unnecessary checks) versions of function (just checking for `error` and `closed` - before 2.10.0 and after system fibers, checking error message - 2.10.0, full function you proposed - for 2.10.1).

Let's not do that and just say, that reconnect doesn't happen in 2.10.0. The user can try killing net.box's fibers as much as he wants in the versions, which will be released after introducing system fibers. 

It's not really a bug, and if you think that we should not complicate the code and increase execution time for such insignificant thing, just close this PR without merging.  

But I would really want to decrease `wait_timeout`, it's not very convenient to change it manually every time it's needed. We can make it in another PR, if you think this change is justified.

vshard/replicaset.lua
```diff
@@ -144,6 +144,45 @@ local function netbox_wait_connected(conn, timeout)
...
+-- if the conn was explicitly cancelled or the connection's worker fiber
+-- was killed). In these situations we need to reestablish it manually as
+-- the missing connection to some replicaset is a critical problem which
+-- leads to the inability of the user to access some part of a data via
+-- router.
+--
+-- Reconnecting after explicit fiber kill doesn't happen in tarantool 2.10.0
+-- as there's no way to determine if the fiber is dead other than checking
+-- the error message of the connection, which is not a good practice as this
+-- check can be easily trigerred false-positively.
```
Gerold103, line 160: It works on 2.10.0 too - you check for `conn._fiber:status() == 'dead'`. I run your test on 2.10.0 and it passed. I also tested it in console just in case:
```Lua
netbox = require('net.box')
box.cfg{listen = 3313}
box.schema.user.grant('guest', 'super')
c = netbox.connect(box.cfg.listen, {reconnect_after = 0.1})
```
```
tarantool> c._fiber:cancel()
---
...

tarantool> 2022-08-12 23:54:11.491 [18782] main/118/:3313 (net.box) net_box.lua:307 W> :3313: fiber is cancelled

---
...

tarantool> c._fiber:status()
---
- dead
...
```
I suggest to enable the new test on all versions, remove mentioning of 2.10.0 from the commit message, and from the comment.

Gerold103 (commented):
Thanks for the fixes! Looks good except one last comment!

vshard/replicaset.lua
Serpentian: Surprisingly, you are right. But the commit https://github.com/tarantool/tarantool/commit/0d944f901cf5400d5832a5528780bb3ecd03ae4b, which introduced `_fiber` in `conn`, was merged on June 22nd, which is later than any release of the tarantool 2.10.0.

I used for testing the version, which is tagged with 2.10.0. In CI, however, AFAICS the versions from tarantool's branches are used. I suppose I didn't get it right how the different versions are formed and why do we still push to them after the release. I'll try to figure that out. 

Fixed everything according to your comment.

Gerold103 (approved):

Issue #339, opened 2022-06-07
Extend CI tests to include 2.10 and master + add EE


Gerold103:
I have no idea how to do it, so here is the ticket.

Totktonada:
Regarding EE: we investigated this topic in golang connector and you can consider results in https://github.com/tarantool/go-tarantool/pull/183.

Regarding 2.10: the best option is to add support for the new tarantool versions into setup-tarantool action. I'm working on it. Precisely, I'm working on ability to set exact versions (like 1.10.8) and then plan to work on 2.10 support. However you can consider tricks implemented in CI for the smtp module: it directly installs necessary tarantool versions.

Totktonada:
Update: I resolved https://github.com/tarantool/setup-tarantool/issues/19, so `'2.10'` can be added into the testing matrix without extra workarounds.

Beware: Don't forget quotes around `'2.10'`. Otherwise it will be interpreted by the YAML parser as the number `2.1`.

sergos:
Duplicate of #331?

Gerold103:
Not exactly, I also want to include tarantool master here.

Totktonada:
> Not exactly, I also want to include tarantool master here.

You can copy-paste https://github.com/tarantool/expirationd/pull/117 for example.

PR #357, opened 2022-07-18
ci: add a debug tarantool version

Serpentian:
This patch inroduce CI testing on the latest git version of
Tarantool, which is built as `Debug`.

By default master branch is used. However, any other one can
be used too: just change everything after `debug-` to another
name.

Part of #339

### DIFF
--- a/.github/workflows/fast_testing.yml
+++ b/.github/workflows/fast_testing.yml
@@ -23,16 +23,63 @@ jobs:
           - '2.8'
           - '2.9'
           - '2.10'
+          - 'debug-master'
+
+    env:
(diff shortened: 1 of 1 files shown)


Serpentian:
Nothing too special, just copy-paste from https://github.com/tarantool/expirationd/pull/117. We really need this one in vshard's CI.

olegrok (approved):
No objections since it is copy-pasted from https://github.com/tarantool/expirationd/pull/117

Serpentian:
Maybe we also need  a release git version of tarantool?

Gerold103 (approved):
Thanks for the patch! Finally we test on latest version!

PR #359, opened 2022-07-25
test: fix flaky router/router2.test.lua

### LINKED ISSUE #358: flaky test: router/router2.test.lua
Tested on master branch. Actually, the diff varies sometimes. Here's one of the possible outputs: 

```
[001] router/router2.test.lua                                         [ fail ]
[001]
[001] Test failed! Result content mismatch:
[001] --- router/router2.result	Thu Jun 23 12:59:17 2022
[001] +++ var/rejects/router/router2.reject	Wed Jul 20 15:17:47 2022
[001] @@ -134,7 +134,7 @@
[001]   | - dead
[001]   | - dead
[001]   | - null
[001] - | - dead
[001] + | - suspended
[001]   | - suspended
[001]   | ...
[001]
[001]
```
[router.tgz](https://github.com/tarantool/vshard/files/9149739/router.tgz)

Serpentian:
See the commit message. The test was runned 250 times locally: the problem seems to be gone.

Closes #358

### DIFF
--- a/test/router/router2.test.lua
+++ b/test/router/router2.test.lua
@@ -46,7 +46,9 @@ vshard.router.cfg(cfg)
 f7 = vshard.router.static.discovery_fiber
 vshard.router.static.discovery_fiber:status()
 
-f1:status(), f2, f3:status(), f4:status(), f5, f6:status(), f7:status()
+while f6:status() ~= 'dead' do fiber.sleep(0.01) end
+f1:status(), f2, f3:status(), f4:status(), f5, f6:status()
+assert(f7:status() == 'suspended' or vshard.router.info().bucket.unknown == 0)
 
 -- Errored discovery continued successfully after errors are gone.
 vshard.router.bootstrap()
--- a/test/router/router2.result
+++ b/test/router/router2.result
@@ -127,7 +127,10 @@ vshard.router.static.discovery_fiber:status()
  | - suspended
  | ...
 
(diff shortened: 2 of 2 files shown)


test/router/router2.test.lua
```diff
@@ -46,6 +46,7 @@ vshard.router.cfg(cfg)
 f7 = vshard.router.static.discovery_fiber
 vshard.router.static.discovery_fiber:status()
 
+while f6:status() ~= 'dead' do fiber.sleep(0.01) end
 f1:status(), f2, f3:status(), f4:status(), f5, f6:status(), f7:status()
```
Gerold103, line 50: That should help to fix the exact problem of the ticket, but I also see another potential issue: by the time we see that `f6` is dead, `f7` can also become dead. This is because it works with `discovery_mode = 'once'`. The fiber will cancel itself when all is discovered.
I suggest to fix it too. For example, like this:
```Lua
assert(f7:status() == 'suspended' or vshard.router.info().bucket.unknown == 0)
```

Gerold103 (commented):
Thanks for the patch, good job!

test/router/router2.test.lua
Serpentian: Yep, that may be a problem. Fixed

Gerold103 (approved):
Thanks for the fixes!

PR #361, opened 2022-07-25
Prohibit simultaneous cfg for router and storage

### LINKED ISSUE #140: Prohibit vshard.router/storage.cfg from multiple fibers
Lets do a simple lock like latch to prohibit multiple cfg as the same time.

Serpentian:
Closes #140

### DIFF
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -41,6 +41,13 @@ local function callrw_session_set(bid, key, value)
                                  {timeout = iwait_timeout})
 end
 
+-- Check whether errors are the same and their messages contain str
+local function assert_errors_equals(err1, err2, str)
+    t.assert_equals(err1, err2)
+    t.assert_not_equals(err1, nil)
+    t.assert_str_contains(err1.message, str)
+end
+
 g.before_all(function()
     vtest.cluster_new(g, global_cfg)
 
@@ -468,13 +475,6 @@ g.test_enable_disable = function(g)
         box.cfg = function(...) return _G.old_box_cfg(...) end
     end)
 
-    -- check whether errors are not nil and their messages are equal to str
-    local assert_errors_equals = function(err1, err2, str)
-        t.assert(err1 and err2)
-        t.assert_equals(err1.message, err2.message)
-        t.assert_str_contains(err1.message, str)
-    end
-
     local err1, err2 = router:exec(function()
         rawset(_G, 'static_router', ivshard.router.internal.routers._static_router)
         rawset(_G, 'new_router', ivshard.router.internal.routers.new_router)
@@ -575,4 +575,46 @@ g.test_explicit_fiber_kill = function(g)
             ilt.assert(err, nil)
         end
     end, {bids})
+end
+
+g.test_simultaneous_cfg = function()
+    local router = vtest.router_new(g, 'router_1')
+
+    router:exec(function(cfg)
+        ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = true
+        rawset(_G, 'fiber_cfg_static', ifiber.new(ivshard.router.cfg, cfg))
+        rawset(_G, 'fiber_cfg_new', ifiber.new(ivshard.router.new,
+                                               'new_router', cfg))
+        _G.fiber_cfg_static:set_joinable(true)
+        _G.fiber_cfg_new:set_joinable(true)
+    end, {global_cfg})
+
+    local function routers_cfg()
+        return router:exec(function(cfg)
+            local static_router = ivshard.router.internal.routers._static_router
+            local new_router = ivshard.router.internal.routers.new_router
+            local _, err1 = pcall(ivshard.router.cfg, cfg)
+            local _, err2 = pcall(static_router.cfg, static_router, cfg)
+            local _, err3 = pcall(new_router.cfg, new_router, cfg)
+            return err1, err2, err3
+        end, {global_cfg})
+    end
+
+    local err1, err2, err3 = routers_cfg()
+    assert_errors_equals(err1, err2, '_static_router is in progress')
+    t.assert_str_contains(err3.message, 'new_router is in progress')
+
+    router:exec(function()
+        ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = false
+        _G.fiber_cfg_static:join()
+        _G.fiber_cfg_new:join()
+    end)
+
+    -- As soon as configuration is done router's reconfiguration is allowed
+    err1, err2, err3 = routers_cfg()
+    t.assert_equals(err1, err2)
+    t.assert_equals(err2, err3)
+    t.assert_equals(err1, nil)
+
+    vtest.drop_instance(g, router)
 end
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -543,7 +543,7 @@ g.test_enable_disable = function(g)
     t.assert_equals(ret1, 1)
 
     -- we don't want this server to interfere with subsequent tests
-    g.router_1:drop()
+    vtest.drop_instance(g, router)
 end
 
 g.test_explicit_fiber_kill = function(g)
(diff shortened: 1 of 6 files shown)


vshard/router/init.lua
```diff
@@ -1741,7 +1756,13 @@ end
...
     if M.static_router then
         -- Reconfigure.
-        router_cfg(M.static_router, cfg, false)
+        router_cfg_fiber_safe(M.static_router, cfg, false)
+    elseif M.routers[STATIC_ROUTER_NAME] then
+        -- M.static_router is still not set but M.routers[STATIC_NAME] is
+        -- already there. This can only mean, that configuration is still in
+        -- progress. But simultaneous configuration is prohibited.
+        error(lerror.vshard(lerror.code.ROUTER_CFG_IS_IN_PROGRESS,
+                            STATIC_ROUTER_NAME))
```
Serpentian, line 1765: This one is for consistency of errors. Without that condition during the first configuration of static router via `vshard.router.cfg` in several fibers simultaneously we will get `ROUTER_ALREADY EXISTS` error, which is not very self-explaining as we allow `vshard.router.cfg` to be invoked several times for reconfiguration.

vshard/router/init.lua
```diff
@@ -1741,7 +1756,13 @@ end
...
     else
         -- Create new static instance.
         local router, err = router_new(STATIC_ROUTER_NAME, cfg)
```
Gerold103, line 1768: Wouldn't this work too?
```Diff
@@ -1754,16 +1754,8 @@ end
 -- static `vshard.router.cfg()` API.
 --
 local function legacy_cfg(cfg)
-    if M.static_router then
-        -- Reconfigure.
-        router_cfg_fiber_safe(M.static_router, cfg, false)
-    elseif M.routers[STATIC_ROUTER_NAME] then
-        -- M.static_router is still not set but M.routers[STATIC_NAME] is
-        -- already there. This can only mean, that configuration is still in
-        -- progress. But simultaneous configuration is prohibited.
-        error(lerror.vshard(lerror.code.ROUTER_CFG_IS_IN_PROGRESS,
-                            STATIC_ROUTER_NAME))
-    else
+    local router = M.routers[STATIC_ROUTER_NAME]
+    if not router then
         -- Create new static instance.
         local router, err = router_new(STATIC_ROUTER_NAME, cfg)
         if router then
@@ -1773,6 +1765,9 @@ local function legacy_cfg(cfg)
         else
             return nil, err
         end
+    else
+        -- Reconfigure.
+        router_cfg_fiber_safe(router, cfg, false)
     end
 end
```

Gerold103 (changes requested):
Thanks for the patch! Finally we can fix this issue!

vshard/router/init.lua
Serpentian: Great solution

Serpentian:
Thanks for review! Rebased and fixed all comments.

test/storage-luatest/storage_2_test.lua
```diff
@@ -368,3 +368,31 @@ test_group.test_storage_callro_refrw_loss = function(g)
     test_storage_callro_refrw_loss(g)
     test_storage_callro_refrw_loss(g, 'test_error')
 end
+
+test_group.test_simultaneous_cfg = function(g)
```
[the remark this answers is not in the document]
Serpentian: Just didn't get how we separate different files with storage tests at first. Now I see)

test/storage-luatest/storage_2_test.lua
```diff
@@ -368,3 +368,31 @@ test_group.test_storage_callro_refrw_loss = function(g)
...
+    local replica_uuid = vtest.replica_name_to_uuid('replica_1_a')
+
+    g.replica_1_a:exec(function(cfg, uuid)
+        ivshard.storage.internal.errinj.ERRINJ_CFG_DELAY = true
+        rawset(_G, 'fiber', ifiber.new(ivshard.storage.cfg, cfg, uuid))
+        _G.fiber:set_joinable(true)
+    end, {cluster_cfg, replica_uuid})
+
+    local storage_cfg = function()
```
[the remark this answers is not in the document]
Serpentian: Yep, I'm fighting with this habit from C++

test/router-luatest/router_test.lua
```diff
@@ -41,6 +41,14 @@ local function callrw_session_set(bid, key, value)
                                  {timeout = iwait_timeout})
 end
 
+-- check whether errors are not nil and their messages
+-- are equal and contains str
+local function assert_errors_equals(err1, err2, str)
+    t.assert(err1 and err2)
+    t.assert_equals(err1.message, err2.message)
+    t.assert_str_contains(err1.message, str)
+end
```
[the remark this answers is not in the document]
Serpentian: Just more strings to compare, comparing only messages seemed to be enough. But your solution is all right too, fixed.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +546,45 @@ g.test_enable_disable = function(g)
...
+    end)
+
+    local routers_cfg = function()
+        return router:exec(function(cfg)
+            local _, err1 = pcall(ivshard.router.cfg, cfg)
+            local _, err2 = pcall(static_router.cfg, static_router, cfg)
+            local _, err3 = pcall(new_router.cfg, new_router, cfg)
+            return err1, err2, err3
+        end, {cluster_cfg})
+    end
```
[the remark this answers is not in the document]
Serpentian: Yes, it will work, except the fact, that there's no `ivshard.router.internal.static_router` before the cfg is finished.

test/storage-luatest/storage_1_test.lua
```diff
@@ -97,3 +97,29 @@ test_group.test_sharded_spaces = function(g)
         s1:drop()
     end, {g.params.engine})
 end
+
+test_group.test_simultaneous_cfg = function(g)
+    g.replica_1_a:exec(function(cfg)
+        _G.ivshard.storage.internal.errinj.ERRINJ_CFG_DELAY = true
```
Gerold103, line 103: You don't need `_G` for `ivshard` and other `i...` variables - they are explicitly allowed globals in our `.luacheckrc` file. For `fiber_cfg` you need it because it is not ignored in luacheck config, but for `ivshard`, `ifiber`, `iwait_timeout`, `iverror` and others you don't.

Here and in all similar places in the patchset.

test/router-luatest/router_test.lua
```diff
@@ -545,3 +545,45 @@ g.test_enable_disable = function(g)
...
+    router:exec(function(cfg)
+        ivshard.router.internal.errinj.ERRINJ_CFG_DELAY = true
+        rawset(_G, 'fiber_cfg_static', ifiber.new(ivshard.router.cfg, cfg))
+        rawset(_G, 'fiber_cfg_new', ifiber.new(ivshard.router.new,
+                                               'new_router', cfg))
+        _G.fiber_cfg_static:set_joinable(true)
+        _G.fiber_cfg_new:set_joinable(true)
+    end, {global_cfg})
+
+    local routers_cfg = function()
```
Gerold103, line 561: I would propose to declare functions rather like that:
```Lua
local function routers_cfg()
```
instead of
```Lua
local routers_cfg = function()
```
See motivation in old comment in the storage commit.

Gerold103 (commented):
Thanks for fixes! Nice job! Just 2 last tiny comments.

test/storage-luatest/storage_1_test.lua
Serpentian: Thank you for explanation

test/router-luatest/router_test.lua
Serpentian: Fixed)

Gerold103 (changes requested):
We are almost there! Last comment - last 2 commit messages are identical. I suppose the last one should be about `router:` instead of `storage:`; it should mention `vshard.router.cfg()`; and it should do `Closes #140`.

PR #367, opened 2022-08-16
ci: disable router/router.test.lua

Serpentian:
Tarantool's CI fails constantly on router/router.test.lua.
Let's disable this test before the fix is commited.


Serpentian:
Disabled not only on release versions, but also on debug, in order not to rerun `router/router.test.lua` on vshard's CI.

PR #368, opened 2022-08-16
test: fix flaky router/router.test.lua

### LINKED ISSUE #366: flaky test: router/router.test.lua
```
router/router.test.lua                                          [ fail ]

Test failed! Result content mismatch:
--- router/router.result	Mon Aug 15 20:19:12 2022
+++ /home/runner/work/vshard/vshard/test/var/rejects/router/router.reject	Mon Aug 15 20:23:24 2022
@@ -498,7 +498,7 @@
 ...
 vshard.router.callro(bucket_id, 'space_get', {'test', {1}})
 ---
-- [1, 1]
+- null
 ...
 vshard.router.callro(bucket_id + 1500, 'space_get', {'test', {1}}) -- nothing
 ---
@@ -618,7 +618,7 @@
 ...
 future:wait_result()
 ---
-- [[1, 1]]
+- []
 ...
 future:is_ready()
 ---
```

Serpentian:
Closes #366

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1093,17 +1093,27 @@ local function vclock_lesseq(vc1, vc2)
 end
 
 local function wait_lsn(timeout, interval)
-    local vclock = box.info.vclock
+    local info = box.info
+    local current_id = info.id
+    local vclock = info.vclock
     local deadline = fiber_clock() + timeout
     repeat
         local done = true
         for _, replica in ipairs(box.info.replication) do
+            -- We should not check the current instance as there's
+            -- no downstream. Moreover, it's not guaranteed that the
+            -- first replica is the same as the current one, so ids
+            -- have to be compared.
+            if replica.id == current_id then
+                goto continue
+            end
             local down = replica.downstream
-            if down and (down.status == 'stopped' or
-                         not vclock_lesseq(vclock, down.vclock)) then
+            if not down or (down.status == 'stopped' or
+                            not vclock_lesseq(vclock, down.vclock)) then
                 done = false
                 break
             end
+            ::continue::
         end
         if done then
             return true
--- a/test/router/suite.ini
+++ b/test/router/suite.ini
@@ -3,7 +3,6 @@ core = tarantool
 description = Router tests
 script = test.lua
 is_parallel = False
-disabled = router.test.lua
 lua_libs = ../lua_libs/util.lua ../lua_libs/git_util.lua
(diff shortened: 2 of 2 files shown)


test/router/router.test.lua
```diff
@@ -175,6 +175,8 @@ uuids = {}
 for uuid, _ in pairs(map) do table.insert(uuids, uuid) end
 uuids
 
+vshard.router.internal.static_router.failover_fiber:wakeup()
+
 --
 -- gh-69: aliases for router.call - callro and callrw.
 --
```
Serpentian, line 182: Rerunned test 200 times locally, the problem seems to be gone. Don't really like that solution as I cannot see any relation to changes done in 2500190d11a74b394e424885b3aa156a523c0a7e and why specifically manual waking of failover fiber removes flakiness. Still investigating

Serpentian: AFAICS, any increasing of the time, needed to execute the test after restarting of the `storage_1_a` seems to be enough to solve the flakiness of the test. However, this is not the root of the problem. 

I switched between storages in the first replicaset after inserting the tuple to the master and discovered, that in case of test failure there was no data on the replica, it was only on master. However, `vshard.storage.sync` said that the data is synchronized, which wasn't actually true. I described everything in commit message.

vshard/storage/init.lua
```diff
@@ -1091,17 +1091,26 @@ local function vclock_lesseq(vc1, vc2)
 end
 
 local function wait_lsn(timeout, interval)
+    local current_id = box.info().id
     local vclock = box.info.vclock
     local deadline = fiber_clock() + timeout
```
Gerold103, line 1096: Nice fix, thanks for figuring this out! I doubt it is caused by any of our commits. Something must have changed in core, because this function didn't change for ages except a recent patch when the function was basically just renamed.
I would advice to consider this diff though, in order not to call the expensive `box.info` too many:
```Diff
 local function wait_lsn(timeout, interval)
-    local current_id = box.info().id
-    local vclock = box.info.vclock
+    local info = box.info
+    local current_id = info.id
+    local vclock = info.vclock
     local deadline = fiber_clock() + timeout
```

Gerold103 (commented):
Thanks for the patch!

vshard/storage/init.lua
Serpentian: Sounds reasonable. Fixed

Gerold103 (approved):
Thanks for the quick fix!

Issue #346, opened 2022-07-04
Bucket GC and send do not account foreign keys


Gerold103:
Foreign keys are supported both in SQL and in raw box API. Bucket send and GC send/delete tuples without thinking they could be attached to tuples in other spaces. Need to figure out what to do here. At least properly ban the foreign-keyed spaces from being sharded. Or try to support their sharding somehow.

sergos:
The ban approach is very much looks like [dictionary](https://github.com/tarantool/vshard/issues/172) feature request, where holder of a FK is dictionary.

Gerold103:
The first estimation (2sp) is for deciding what to do - ban or not. If not, then find a way to keep the FKs not broken.

PR #373, opened 2022-10-04
storage: introduce `on_bucket_event` trigger

### LINKED ISSUE #372: Mark the rebalancing transactions
Buckets migration looks like a regular client data activity, indistinguishable in the replication flow.
Although the solution should be brought at the different level, we can provide a temporary solution to enable number of clients eager to have efficient data change capturing. 
Since all rebalancing activities are done in transactions, we can add a fake operation in the beginning of the transaction, so that client will only have to decode this first operation to skip all following ops in the transaction without decoding. The operation can be a regular DML into a _specific system space_. There's only one limitation - the space should be memtx or vinyl depending on the transaction, since Tarantool doesn't support multi-engine transactions. The mark operation itself is an option and can be omit in case the _specific system space_ is not present.

In `bucket_recv_xc` and in `gc_bucket_in_space_xc` the call to `box.begin()` should be followed by an operation in a system space, named `_vhsard_rebalancing_{memtx|vinyl}` depending on the engine, where current bucket is.

Serpentian:
Buckets migration looks like a regular client data activity,
indistinguishable in the replication flow. In order to allow
efficient change data capturing we can add a fake operation in the
beginning of the transaction, so that a client could determing which
operations can be skipped (using HEADER.tsn) as they were done not
by the user but by gc or rebalancer.

In order not to implement "dirty hack" with special cases, this
commit introduces the ability for a user to set the trigger.

Closes https://github.com/tarantool/vshard/issues/372

Co-authored-by: Vladislav Shpilevoy <v.shpilevoy@tarantool.org>

@TarantoolBot document
Title: vshard.storage.on_bucket_event()

vshard.storage.on_bucket_event([trigger-function[, old-trigger-function]])

Create a trigger, which is executed when the data from the user spaces
is changed (deleted or inserted) due to the rebalancing process. It's
triggered on every data batch changed.

Parameters:
  * trigger-function (function) – function which will become the trigger
    function;
  * old-trigger-function (function) – existing trigger function which will
    be replaced by trigger-function;

Return:
  nil or function pointer

As everything executed inside triggers is already in a transaction, you
shouldn't use transactions, yield-operations, changes to different space
engines.

The trigger-function can have up to three parameters:
  * (string) event_type. In order to distinguish event you could
    compare this argument with supported event types:
      - 'bucket_data_recv_txn',
      - 'bucket_data_gc_txn';
  * (unsigned) bucket_id;
  * (table) data. E.g. includes an array of all spaces (data.spaces),
    affected by a transaction, in which trigger-function is executed.

Example

```lua
vshard.storage.on_bucket_event(function(event, bucket_id, data)
    if event == 'bucket_data_recv_txn' then
        -- Handle it.
        for idx, space in ipairs(data.spaces) do
            ...
        end
    elseif event == 'bucket_data_gc_txn' then
        -- Handle it.
        ...
    end
end)
```


vshard/consts.lua
```diff
@@ -47,6 +47,8 @@ return {
     DEFAULT_BUCKET_SEND_TIMEOUT = 10,
     DEFAULT_BUCKET_RECV_TIMEOUT = 10,
 
+    MIGRATION_NOTIFY_SPACE_NAME = '_vshard_migration_notify_',
```
sergos, line 50: Better name it `_PREFIX` since complete name is concatenated lately with engine name

vshard/storage/init.lua
```diff
@@ -1477,6 +1492,7 @@ local function bucket_recv_xc(bucket_id, from, data, opts)
             end
             limit = limit - 1
             if limit == 0 then
+                migration_notify(space.engine)
```
sergos, line 1495: That's a wrong place to put the notification ops: it should be placed right after the `box.begin()` so that it'll be the first operation in the transaction, not the last.

vshard/storage/init.lua
```diff
@@ -1488,6 +1504,7 @@ local function bucket_recv_xc(bucket_id, from, data, opts)
                 limit = consts.BUCKET_CHUNK_SIZE
             end
         end
+        migration_notify(space.engine)
```
sergos, line 1507: ditto

vshard/storage/init.lua
```diff
@@ -2006,12 +2023,14 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
         space:delete(util.tuple_extract_key(tuple, pk_parts))
         limit = limit - 1
         if limit == 0 then
+            migration_notify(space.engine)
```
sergos, line 2026: ditto

vshard/storage/init.lua
```diff
@@ -2006,12 +2023,14 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
...
             box.commit()
             bucket_generation =
                 bucket_guard_xc(bucket_generation, bucket_id, status)
             goto restart
         end
     end
+    migration_notify(space.engine)
```
sergos, line 2033: ditto

sergos (changes requested):

vshard/cfg.lua
```diff
@@ -205,6 +205,15 @@ local function check_discovery_mode(value)
     end
 end
 
+local function check_migration_notify(value)
+    for engine, _ in pairs(value) do
+        if engine ~= 'memtx' and engine ~= 'vinyl' then
+            error("Expected 'memtx' or 'vinyl' as keys " ..
+                  "of 'migration_notify' dictionary: got " .. engine)
```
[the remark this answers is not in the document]
Serpentian: Didn't really got what you mean

vshard/consts.lua
Serpentian: Fixed

vshard/storage/init.lua
Serpentian: Fixed as well as in all other cases

vshard/cfg.lua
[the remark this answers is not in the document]
sergos: Change the message. By now you'll get with 'garble' as an argument:
`Expected 'memtx' or 'vinyl' as keys  of 'migration_notify' dictionary: got garble`
I propose to make it
`Expected 'memtx' or 'vinyl' as keys  of 'migration_notify' dictionary, got: garble`

vshard/storage/init.lua
sergos: I bet you're missing one at the line 1507 - when limit is reached we start another transaction.

Serpentian: You're right. Fixed

vshard/cfg.lua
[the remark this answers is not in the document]
Serpentian: Fixed

sergos (approved):
Final minor commit message updates:
"operation in the end of the transaction" - at the beginning of the transaction
"in the range [HEADER.tsn, HEADER.lsn]" - with this HEADER.tsn
"done by garbage collector or rebalancer" - done by rebalancer

Gerold103 (commented):
There is a new proposal in the ticket: https://github.com/tarantool/vshard/issues/372#issuecomment-1309575158. This way might be much less crutchy and significantly more useful in general. Can you talk to customers of this feature if they can leverage the triggers-way like described in the comment instead of us doing those spaces in master?

Serpentian:
> There is a new proposal in the ticket: [#372 (comment)](https://github.com/tarantool/vshard/issues/372#issuecomment-1309575158). This way might be much less crutchy and significantly more useful in general. Can you talk to customers of this feature if they can leverage the triggers-way like described in the comment instead of us doing those spaces in master?

Agree with you, this solution is a big crutch, triggers looks much better. Don't really think they will be against this idea, but I'll definitely ask them.

vshard/storage/init.lua
```diff
@@ -975,6 +982,10 @@ local function on_master_enable(new_func, old_func)
     end
 end
 
+local function on_bucket_event(new_func, old_func)
+    M._on_bucket_event(new_func, old_func)
```
Gerold103, line 986: You should to `return M._on...()`. Without `return` you won't return the resulting installed function, which is quite useful when people set closures as triggers. Lets use that in the tests too.

vshard/storage/init.lua
```diff
@@ -1589,6 +1600,7 @@ local function bucket_recv_xc(bucket_id, from, data, opts)
     end
     local bucket_generation = M.bucket_generation
     local limit = consts.BUCKET_CHUNK_SIZE
+    local event_type = 'bucket_data_recv_txn'
```
Gerold103, line 1603: For the record, I am fine if you would store these in `vshard.consts`. We only shouldn't expose `vshard.consts` anywhere in the docs, but can use it for internals easily. Up to you.

vshard/storage/init.lua
```diff
@@ -1597,6 +1609,7 @@ local function bucket_recv_xc(bucket_id, from, data, opts)
             return nil, lerror.box(err)
         end
         box.begin()
+        M._on_bucket_event:run(event_type, bucket_id, space_name)
```
Gerold103, line 1612: We should change the last argument to be more extendible. To this format:
```
{
    spaces = {name1, name2, ...}
}
```
Which would be in our case so far always `{spaces = {space_name}}`.
The reasons are:
- There might be multiple spaces per-txn potentially. In fact, I think that is a matter of short time. See https://github.com/tarantool/vshard/issues/346. With foreign keys we will have to touch multiple spaces per txn;
- We can't just provide storage engine name instead of space names, because in the future the transaction manager might become single for both engines. Hence allowing FKs between engines. Thus making txns engine-specific only temporarily;
- When the space names array is wrapped into a dict, it allows us to add here more keys than just `spaces` in the future, if needed.

All the same for bucket GC.

vshard/storage/init.lua
```diff
@@ -3721,6 +3736,7 @@ return {
     cfg = function(cfg, uuid) return storage_cfg_fiber_safe(cfg, uuid, false) end,
     on_master_enable = storage_make_api(on_master_enable),
     on_master_disable = storage_make_api(on_master_disable),
+    on_bucket_event = storage_make_api(on_bucket_event),
```
Gerold103, line 3739: This one probably should be allowed to call even before `vshard.storage.cfg` starts and ends. Otherwise rebalancing might start between `cfg` ends and when a user manages to set the triggers.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
         box.space.test:drop()
     end)
 end
+
+--
+-- Custom triggers on bucket changes
+--
+test_group.test_on_bucket_event = function(g)
+    -- Initial problem: buckets migration looks like a regular
+    -- client data activity, indistinguishable in the replication flow.
```
Gerold103, line 166: JFYI, comment lines can be 80 symbols long.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
...
+    -- In order to allow efficient change data capturing we can add a fake
+    -- operation in the beginning of the transaction, so that a client could
+    -- determing which operations can be skipped (using HEADER.tsn) as they
+    -- were done not by the user but by gc or rebalancer.
+    vtest.cluster_exec_each_master(g, function(engine)
+        -- Space, in which these fake operations will be written
+        rawset(_G, 'notify_name', '_vshard_migration_notify_' .. engine)
```
Gerold103, line 173: Lets make the name less conflicting with potential future vshard-private naming and make it something like `test_migration_notify` or similar.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
...
+        local s = box.schema.space.create('data', {
+             engine = engine,
+             format = {
+                {'id', 'unsigned'},
+                {'bucket_id', 'unsigned'},
+             },
+        })
+        s:create_index('id', {parts = {'id'}})
+        s:create_index('bucket_id', {parts = {'bucket_id'}, unique = false})
+        rawset(_G, 'notify_trigger', function(event_type, bucket_id)
```
Gerold103, line 186: Lets validate all the trigger arguments. To ensure nothing is missing.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
...
+                box.space[_G.notify_name]:delete({bucket_id})
+            end
+        end)
+        -- Set trigger
+        ivshard.storage.on_bucket_event(_G.notify_trigger)
+    end, {g.params.engine})
+
+    local uuid_1 = g.replica_1_a:exec(function()
+        return box.info.cluster.uuid
+    end)
```
Gerold103, line 198: `vtest.replicaset_name_to_uuid(1)`. You can also do it right in `bucket_send()` invocation place using `ivtest`.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
...
+        ivshard.storage.bucket_send(bid, uuid_1)
+    end, {uuid_1})
+
+    -- Read xlog, get all messages written to the special space
+    local function get_notify_wal_rows(storage)
+        return storage:exec(function()
+            local xlog = require('xlog')
+            local notifications = {}
+            local notify_space = box.space[_G.notify_name]
+            for _, request in xlog.pairs('00000000000000000000.xlog') do
```
Gerold103, line 212: That for sure is quite a complicated way. Lets not depend on whether it is stored in xlog or not, and for sure not depend on file names. To validate that everything ends up in one txn you can use `box.txn.on_commit()` trigger. Here you can read up all the rows and append them to some global table. Then further down the test you take that table and check each saved txn content. That would allow not to touch the files and nor look for what is the current's xlog name. And you won't depend on 1.10 - txn triggers are supported there.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,99 @@ test_group.test_bucket_send_field_types = function(g)
...
+                    table.insert(notifications, request)
+                end
+            end
+            return notifications
+        end)
+    end
+
+    g.replica_1_a:exec(function()
+        ilt.helpers.retrying({timeout = iwait_timeout}, function()
+            if box.space.data:count() < 1 then
```
Gerold103, line 223: `vshard.storage.bucket_send()` is synchronous, it is finished when the function returns. Don't need to wait for it.

Gerold103 (changes requested):
Thanks for the patch! Good stuff, just nits now.

vshard/storage/init.lua
Serpentian: Actually the same thing applies to `on_master_enable/disable`. AFAICS, by it is supposed to be executed before `cfg` by design. I think we should remove `storage_make_api` for them to.

vshard/storage/init.lua
Serpentian: Moreover, the user may want to check the list of installed triggers with `vshard.storage.on_bucket_event()`. Let's also return values for the `on_master_enable/disable` too.

Serpentian: Strange question. Why aren't `on_master_enable/disable` documented?  There's not even a ticket on doc repo about them.

Serpentian: Lol. Actually, the trigger, created from lua doesn't return function (e.g. `net.box.on_schema_reload`). Trigger related functions, exported from C, do that (e.g.`box.sessions.on_auth`)

vshard/storage/init.lua
Serpentian: Consts looks cleaner, I suppose. Better to use them

vshard/storage/init.lua
Serpentian: Fixed

test/storage-luatest/storage_1_1_test.lua
Serpentian: Such a great proposal. Fixed

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -157,3 +157,76 @@ test_group.test_bucket_send_field_types = function(g)
...
+        ilt.assert_equals(#_G.migration_notifications, 1)
+    end)
+
+    -- Cleanup
+    vtest.cluster_exec_each_master(g, function()
+        ivshard.storage.on_bucket_event(nil, _G.notify_trigger)
+        box.space[_G.notify_name]:drop()
+        box.space.data:drop()
+    end)
+end
```
Gerold103, line 232: Please, apply this diff:
<details><summary>Diff</summary>

```Diff
diff --git a/test/storage-luatest/storage_1_1_test.lua b/test/storage-luatest/storage_1_1_test.lua
index e1b6229..cd49050 100644
--- a/test/storage-luatest/storage_1_1_test.lua
+++ b/test/storage-luatest/storage_1_1_test.lua
@@ -1,4 +1,5 @@
 local t = require('luatest')
+local vconst = require('vshard.consts')
 local vtest = require('test.luatest_helpers.vtest')
 local vutil = require('vshard.util')
 
@@ -159,74 +160,107 @@ test_group.test_bucket_send_field_types = function(g)
 end
 
 --
--- Custom triggers on bucket changes
+-- Custom triggers on bucket changes.
 --
 test_group.test_on_bucket_event = function(g)
-    -- Initial problem: buckets migration looks like a regular
-    -- client data activity, indistinguishable in the replication flow.
-    -- In order to allow efficient change data capturing we can add a fake
-    -- operation in the beginning of the transaction, so that a client could
-    -- determing which operations can be skipped (using HEADER.tsn) as they
-    -- were done not by the user but by gc or rebalancer.
     vtest.cluster_exec_each_master(g, function(engine)
-        -- Space, in which these fake operations will be written
-        rawset(_G, 'notify_name', 'test_migration_notify_' .. engine)
-        box.schema.space.create(_G.notify_name, {
+        local event_space = box.schema.space.create('events', {
             engine = engine,
-        }):create_index('primary')
-        local s = box.schema.space.create('data', {
-             engine = engine,
-             format = {
-                {'id', 'unsigned'},
-                {'bucket_id', 'unsigned'},
-             },
         })
-        s:create_index('id', {parts = {'id'}})
-        s:create_index('bucket_id', {parts = {'bucket_id'}, unique = false})
+        event_space:create_index('primary')
+
+        local function data_space_new(name)
+            local s = box.schema.space.create(name, {
+                engine = engine,
+                format = {
+                    {'id', 'unsigned'},
+                    {'bucket_id', 'unsigned'},
+                },
+            })
+            s:create_index('id', {parts = {'id'}})
+            s:create_index('bucket_id', {
+                parts = {'bucket_id'}, unique = false
+            })
+            return s
+        end
+        local data_space1 = data_space_new('data1')
+        local data_space2 = data_space_new('data2')
+
+        local test_txn_id = 0
+        local test_events = {}
+        local test_commits = {}
+        rawset(_G, 'test_trigger', function(event_type, bucket_id, data)
+            test_txn_id = test_txn_id + 1
+            local tid = test_txn_id
+            local event_data = {event_type, bucket_id, data}
+            ilt.assert(box.is_in_txn())
+            ilt.assert(not test_events[tid])
+            event_space:insert{tid, event_data}
+            test_events[tid] = event_data
 
-        -- Set on_bucket_event trigger
-        rawset(_G, 'migration_notifications', {})
-        rawset(_G, 'notify_trigger', function(event_type, bucket_id, opts)
-            -- Set on_commit trigger
             box.on_commit(function(iterator)
-                local notify_space = box.space[_G.notify_name]
-                for idx, _, _, space_id in iterator() do
-                    if space_id == notify_space.id then
-                        table.insert(_G.migration_notifications, idx)
+                for _, t1, t2, space_id in iterator() do
+                    if space_id == event_space.id then
+                        test_commits[tid] = true
+                        break
                     end
                 end
             end)
-            ilt.assert_not_equals(opts and opts.spaces, nil)
-            if event_type == 'bucket_data_recv_txn' or
-                    event_type == 'bucket_data_gc_txn' then
-                box.space[_G.notify_name]:delete({bucket_id})
+        end)
+        ivshard.storage.on_bucket_event(_G.test_trigger)
+
+        rawset(_G, 'process_events', function(expected_events)
+            -- Compare only the items, not looking at the order. Because the
+            -- spaces can be sent in any order, can't rely on one.
+            ilt.assert_items_equals(test_events, expected_events)
+            -- Ensure that each saved event is also persisted.
+            for tid, event_data in pairs(test_events) do
+                ilt.assert_equals(box.space.events:get{tid}, {tid, event_data})
+                ilt.assert_not_equals(test_commits[tid], nil)
+                test_commits[tid] = nil
+                box.space.events:delete{tid}
             end
+            test_events = {}
+            -- No unexpected events anywhere.
+            ilt.assert(not next(test_commits))
+            ilt.assert_equals(box.space.events.index[0]:min(), nil)
         end)
-        ivshard.storage.on_bucket_event(_G.notify_trigger)
     end, {g.params.engine})
 
-    g.replica_2_a:exec(function()
+    local rs1_uuid = g.replica_1_a:exec(function()
+        return box.info.cluster.uuid
+    end)
+    local bid = g.replica_2_a:exec(function(rs1_uuid)
         local bid = _G.get_first_bucket()
-        local uuid = ivtest.replicaset_name_to_uuid('replica_1_a')
-        box.space.data:insert({bid, bid})
-        ivshard.storage.bucket_send(bid, uuid)
+        box.space.data1:insert({10, bid})
+        box.space.data2:insert({20, bid})
+        local ok, err = ivshard.storage.bucket_send(
+            bid, rs1_uuid, {timeout = _G.iwait_timeout})
+        ilt.assert_equals(err, nil)
+        ilt.assert(ok)
         _G.bucket_gc_wait()
-    end)
+        return bid
+    end, {rs1_uuid})
 
-    -- Rebalancer related notifications
-    g.replica_1_a:exec(function()
-        ilt.assert_equals(#_G.migration_notifications, 1)
-    end)
+    g.replica_2_a:exec(function(events)
+        _G.process_events(events)
+    end, {{
+        [1] = {vconst.BUCKET_TRIGGER.GC, bid, {spaces = {'data1'}}},
+        [2] = {vconst.BUCKET_TRIGGER.GC, bid, {spaces = {'data2'}}},
+    }})
 
-    -- GC related notifications
-    g.replica_2_a:exec(function()
-        ilt.assert_equals(#_G.migration_notifications, 1)
-    end)
+    g.replica_1_a:exec(function(events)
+        _G.process_events(events)
+    end, {{
+        [1] = {vconst.BUCKET_TRIGGER.RECV, bid, {spaces = {'data1'}}},
+        [2] = {vconst.BUCKET_TRIGGER.RECV, bid, {spaces = {'data2'}}},
+    }})
 
     -- Cleanup
     vtest.cluster_exec_each_master(g, function()
-        ivshard.storage.on_bucket_event(nil, _G.notify_trigger)
-        box.space[_G.notify_name]:drop()
-        box.space.data:drop()
+        ivshard.storage.on_bucket_event(nil, _G.test_trigger)
+        box.space.events:drop()
+        box.space.data1:drop()
+        box.space.data2:drop()
     end)
 end
diff --git a/vshard/consts.lua b/vshard/consts.lua
index f606666..5d6e058 100644
--- a/vshard/consts.lua
+++ b/vshard/consts.lua
@@ -10,8 +10,11 @@ return {
     },
 
     BUCKET_TRIGGER = {
-      RECV = 'bucket_data_recv_txn',
-      GC = 'bucket_data_gc_txn'
+        -- Txn triggers allow to attach anything to the bucket transaction. That
+        -- was requested by a customer to be able to detect vshard book-keeping
+        -- actions like bucket sending right in the journal.
+        RECV = 'bucket_data_recv_txn',
+        GC = 'bucket_data_gc_txn'
     },
 
     STATUS = {
```
</details>

The problems with the old test were that
- Its comment rather belongs to the event declarations;
- Need to test the third argument of the trigger. Testing just for `~= nil` wouldn't be enough. We could break the third argument layout and the test wouldn't notice;
- Need to test multiple spaces to ensure that they are split into separate txns and events;
- Need to test the other 2 arguments of the event as well. Testing just event count wouldn't be enough;
- I am not sure I understood this `box.space[_G.notify_name]:delete({bucket_id})` line. What is it for? It would be just NOP. Because this space was always empty in your test.
- Need to test if `bucket_send()` was ok. It wouldn't raise an exception on an error.
- About `ivtest.replicaset_name_to_uuid('replica_1_a')` I was wrong, sorry. It won't work properly in general case. It generates the UUIDs on fly. In your version of the test it worked by luck, because `replica_1_a` on the storage got the same UUID as `replicaset 1` on the test node. But otherwise it is obviously wrong - we tried to use **replicaset**_name_to_uuid to get UUID of a **replica**. Replicaset and replica UUIDs are not related.

vshard/consts.lua
```diff
@@ -9,6 +9,11 @@ return {
         GARBAGE = 'garbage',
     },
 
+    BUCKET_TRIGGER = {
+      RECV = 'bucket_data_recv_txn',
+      GC = 'bucket_data_gc_txn'
```
Gerold103, line 14: Please, fix the alignment (you used 2-space indentation, but it should be 4-space). And rename `BUCKE_TRIGGER` to `BUCKET_EVENT`.

Gerold103 (changes requested):
Thanks for the fixes! I think we will push this soon, just last amendments.

vshard/consts.lua
Serpentian: Fixed

test/misc/reconfigure.result
```diff
@@ -129,10 +129,10 @@ function on_master_disable() disable_count = disable_count + 1 end
 function on_master_enable() enable_count = enable_count + 1 end
```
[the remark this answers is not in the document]
Serpentian: Got you, removed request to Tarantool Bot

test/luatest_helpers/vtest.lua
```diff
@@ -625,4 +625,5 @@ return {
     wait_timeout = wait_timeout,
```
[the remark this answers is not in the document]
Serpentian: Fixed

test/storage-luatest/storage_1_1_test.lua
Serpentian: The new test looks great, applied it almost entirely. Just fixed places, luacheck doesn't like

test/luatest_helpers/vtest.lua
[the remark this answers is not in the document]
Gerold103: The new doc request is still wrong ...
You say `data.spaces.space_name`, but there is no such thing. The names are stored in an array, like `data.spaces[1]`, `data.spaces[2]`, etc. Same in the example where you show `if data.spaces.space_name then` - it won't work.

test/luatest_helpers/vtest.lua
```diff
@@ -625,4 +625,5 @@ return {
...
     busy_step = busy_step,
     drop_instance = drop_instance,
+    replicaset_name_to_uuid = replicaset_name_to_uuid,
```
Gerold103, line 628: That is not needed now.

Gerold103 (changes requested):
Thanks for the fixes! And we are almost done, one new nit and an old comment about a doc request is update (see above).

Gerold103 (approved):
Thanks for the fixes!

PR #374, opened 2022-10-05
Fixed rewrited error arguments

vaintrub:
This patch solves the problem when the error argument and the error name variable match, the variable is overwritten.

https://github.com/tarantool/vshard/blob/c508c8228ec81872cd246d3e062294a12d2d3e5d/vshard/error.lua#L109-L113

On line 256 we add the `name` key to ret, and on line 261 we overwrite it
https://github.com/tarantool/vshard/blob/c508c8228ec81872cd246d3e062294a12d2d3e5d/vshard/error.lua#L255-L261

### DIFF
--- a/test/multiple_routers/multiple_routers.result
+++ b/test/multiple_routers/multiple_routers.result
@@ -176,10 +176,11 @@ util = require('util')
 util.check_error(vshard.router.new, 'router_2', configs.cfg_2)
 ---
 - null
-- type: ShardingError
-  code: 21
+- code: 21
+  type: ShardingError
   name: ROUTER_ALREADY_EXISTS
   message: Router with name router_2 already exists
+  router_name: router_2
 ...
 -- Reload router module.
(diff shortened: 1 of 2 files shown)


Gerold103 (commented):
Thanks for the patch! Please, make the commit message comply with our styling rules: https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message and double-check the grammar for too visible mistakes.

In your case the title should be like `error: fix conflicting member names`. And the commit message should explain what and why was changed, like you did in the PR comment.

vaintrub:
> Thanks for the patch! Please, make the commit message comply with our styling rules: https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message and double-check the grammar for too visible mistakes.
> 
> In your case the title should be like `error: fix conflicting member names`. And the commit message should explain what and why was changed, like you did in the PR comment.

Thanks! I have corrected the commit message

Gerold103 (approved):
Thanks for the patch!

PR #382, opened 2022-11-17
storage: fix usage of Tarantool internal API for getting schema version

### LINKED ISSUE #381: Start using `box.info.schema_version` instead of `box.internal.schema_version`
https://github.com/tarantool/tarantool/issues/7904 exports schema version to public API and Lua:
start using it, when available, instead of unreliable internal API.

CuriousGeorgiy:
tarantool/tarantool#7904 exports schema version to public API and Lua: start using it, when available, instead of unreliable internal API.

Closes #381

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1735,7 +1735,8 @@ end
 local sharded_spaces_cache_schema_version = nil
 local sharded_spaces_cache = nil
 local function find_sharded_spaces()
-    if sharded_spaces_cache_schema_version == box.internal.schema_version() then
+    local schema_version = util.schema_version()
+    if sharded_spaces_cache_schema_version == schema_version then
         return sharded_spaces_cache
     end
     local spaces = {}
@@ -1749,7 +1750,7 @@ local function find_sharded_spaces()
             end
         end
     end
-    sharded_spaces_cache_schema_version = box.internal.schema_version()
+    sharded_spaces_cache_schema_version = schema_version
     sharded_spaces_cache = spaces
     return spaces
 end
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -345,6 +345,31 @@ local feature = {
     csw = fiber.self().csw ~= nil,
 }
 
+local schema_version = function()
+    return box.info.schema_version
+end
+do
+    -- Since 2.11 the internal function is replaced by the info field.
+    -- Use pcall() in case box.info would raise an error on an unconfigured
+    -- instance (it does at 1.10).
+    local ok, res = pcall(schema_version)
+    if not ok or res == nil then
+        -- Fallback to older versions when the internal func was available.
+        schema_version = box.internal.schema_version
+        assert(schema_version ~= nil)
+    end
+    -- Can't tell whether box.info or the internal func will start yielding at
+    -- some day. If they do, that needs to be caught, because it potentially
+    -- might make the schema version not very much usable as a cache generation
+    -- number.
+    if feature.csw then
+        local csw1 = fiber.self().csw()
+        local _ = schema_version()
+        local csw2 = fiber.self().csw()
+        assert(csw2 == csw1)
+    end
+end
+
 return {
     uri_eq = uri_eq,
     tuple_extract_key = tuple_extract_key,
@@ -361,4 +386,5 @@ return {
     index_min = index_min,
     index_has = index_has,
     feature = feature,
+    schema_version = schema_version,
 }


vshard/storage/init.lua
```diff
@@ -1735,7 +1735,8 @@ end
 local sharded_spaces_cache_schema_version = nil
 local sharded_spaces_cache = nil
 local function find_sharded_spaces()
-    if sharded_spaces_cache_schema_version == box.internal.schema_version() then
+    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
+    if sharded_spaces_cache_schema_version == (box.info.schema_version or box.internal.schema_version()) then
```
Gerold103, line 1739: 1. Please, limit the commit title by 50 symbols.
2. Your changes are out of 80 symbols border and duplicate not so trivial code. Please, apply my diff below:
<details><summary>Diff</summary>

```Diff
diff --git a/vshard/storage/init.lua b/vshard/storage/init.lua
index 413e929..e729900 100644
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1735,8 +1735,8 @@ end
 local sharded_spaces_cache_schema_version = nil
 local sharded_spaces_cache = nil
 local function find_sharded_spaces()
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    if sharded_spaces_cache_schema_version == (box.info.schema_version or box.internal.schema_version()) then
+    local schema_version = util.schema_version()
+    if sharded_spaces_cache_schema_version == schema_version then
         return sharded_spaces_cache
     end
     local spaces = {}
@@ -1750,8 +1750,7 @@ local function find_sharded_spaces()
             end
         end
     end
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    sharded_spaces_cache_schema_version = box.info.schema_version or box.internal.schema_version()
+    sharded_spaces_cache_schema_version = schema_version
     sharded_spaces_cache = spaces
     return spaces
 end
diff --git a/vshard/util.lua b/vshard/util.lua
index a7f911d..92c4042 100644
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -345,6 +345,31 @@ local feature = {
     csw = fiber.self().csw ~= nil,
 }
 
+local schema_version = function()
+    return box.info.schema_version
+end
+do
+    -- Since 2.11 the internal function is replaced by the info field.
+    -- Use pcall() in case box.info would raise an error on an unconfigured
+    -- instance (it does at 1.10).
+    local ok, res = pcall(schema_version)
+    if not ok or res == nil then
+        -- Fallback to older versions when the internal func was available.
+        schema_version = box.internal.schema_version
+        assert(schema_version)
+    end
+    -- Can't tell whether box.info or the internal func will start yielding at
+    -- some day. If they do, that needs to be caught, because it potentially
+    -- might make the schema version not very much usable as a cache generation
+    -- number.
+    if feature.csw then
+        local csw1 = fiber.self().csw()
+        local _ = box.info.schema_version
+        local csw2 = fiber.self().csw()
+        assert(csw2 == csw1)
+    end
+end
+
 return {
     uri_eq = uri_eq,
     tuple_extract_key = tuple_extract_key,
@@ -361,4 +386,5 @@ return {
     index_min = index_min,
     index_has = index_has,
     feature = feature,
+    schema_version = schema_version,
 }
```
</details>

Gerold103 (changes requested):
Thanks for the patch!

vshard/storage/init.lua
CuriousGeorgiy: Thanks a lot for the suggested diff! I have one question regarding it:
```diff
@@ -1735,8 +1735,8 @@ end
 local sharded_spaces_cache_schema_version = nil
 local sharded_spaces_cache = nil
 local function find_sharded_spaces()
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    if sharded_spaces_cache_schema_version == (box.info.schema_version or box.internal.schema_version()) then
+    local schema_version = util.schema_version()
+    if sharded_spaces_cache_schema_version == schema_version then
         return sharded_spaces_cache
     end
     local spaces = {}
@@ -1750,8 +1750,7 @@ local function find_sharded_spaces()
             end
         end
     end
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    sharded_spaces_cache_schema_version = box.info.schema_version or box.internal.schema_version()
+    sharded_spaces_cache_schema_version = schema_version
```

What if `pairs(box.space)` yields in this snippet? Are you sure you don't want to use the latest schema version, when you update `sharded_spaces_cache_schema_version`, i.e.:

```diff
@@ -1735,8 +1735,8 @@ end
 local sharded_spaces_cache_schema_version = nil
 local sharded_spaces_cache = nil
 local function find_sharded_spaces()
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    if sharded_spaces_cache_schema_version == (box.info.schema_version or box.internal.schema_version()) then
+    if sharded_spaces_cache_schema_version == util.schema_version() then
         return sharded_spaces_cache
     end
     local spaces = {}
@@ -1750,8 +1750,7 @@ local function find_sharded_spaces()
             end
         end
     end
-    -- If Tarantool public API for schema version is not available, fallback to unreliable internal API.
-    sharded_spaces_cache_schema_version = box.info.schema_version or box.internal.schema_version()
+    sharded_spaces_cache_schema_version = util.schema_version()
```

Gerold103: There is nothing to yield. But in case a yield is ever added somehow, then it is better to remember the old version. Otherwise if you remember the version at the end of pairs, then you might miss some spaces added or deleted since the pairs started and won't notice that. Your cache will think that it is of the latest version.

In that sense the old code was also not quite safe.

CuriousGeorgiy: > Otherwise if you remember the version at the end of pairs, then you might miss some spaces added or deleted since the pairs started and won't notice that. 

This point didn't come to my mind, thanks for elaborating!

Applied your diff completely.

Gerold103 (changes requested):
Please, apply this diff. Otherwise 2.10 fails:
```Diff
@@ -364,7 +364,7 @@ do
     -- number.
     if feature.csw then
         local csw1 = fiber.self().csw()
-        local _ = box.info.schema_version
+        local _ = schema_version()
         local csw2 = fiber.self().csw()
         assert(csw2 == csw1)
     end
```

CuriousGeorgiy:
@Gerold103 I'm sorry, overlooked that one: fixed.

Gerold103 (approved):
Thanks for the fix! It fails to compile on tnt master, but that is not a fault of this patch. Locally it compiles and passes all tests on my machine just fine.

PR #385, opened 2022-11-24
ci: install all dependencies from doc for build

Serpentian:
The list of installed apt packages for default github images constantly changes. These cause the fail of building tarantool from the master branch on CI.

Let's install all packages, described in tarantool's build docs. Already installed ones will be just skipped by the apt program.

NO_TEST=ci
NO_DOC=ci


Gerold103 (approved):
Thanks!

PR #391, opened 2023-01-10
Pick a replica according prefered zone (highest priority replica zone) in round-robin manner.

eugenepaniot:
Pick a replica according prefered zone (highest priority replica zone) in round-robin manner.

Current `callr*` calls use only first replica. Current `callbr*` calls use replicas in round-robin manner only, without zonal preferences that leads to extra cross zonal traffic (that costs some penies).

Current PR suggests two new methods and new load-balancing (replica selection) algorithm:
* callbzro - prefer, but not limit, Inner zone balanced method with round-robin. Similar to existing `callbro`;
* callbzre - prefer, but not limit, Inner zone balanced method with round-robin. Similar to existing `callbre`;

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -577,13 +577,35 @@ local function replicaset_template_multicallro(prefer_replica, balance)
     local function pick_next_replica(replicaset, now)
         local r
         local master = replicaset.master
+
         if balance then
+            local prefered_zone = replicaset.priority_list[1].zone
             local i = #replicaset.priority_list
             while i > 0 do
                 r = replicaset_balance_replica(replicaset)
                 i = i - 1
                 if r:is_connected() and (not prefer_replica or r ~= master) and
-                   replica_check_backoff(r, now) then
+                    replica_check_backoff(r, now)
+                then
+                    -- Pick a replica according prefered zone (highest priority replica zone) in round-robin manner.
+                    if balance == 2 and prefered_zone then
+                        -- save current rr-cursor position
+                        local cbi = replicaset.balance_i
+                        -- find next replica
+                        local nr = replicaset_balance_replica(replicaset)
+
+                        if prefered_zone and r.zone and r.zone == prefered_zone -- current replica is in prefered_zone
+                            and nr.zone and nr.zone ~= prefered_zone -- next replica is from different zone (reached prefered_zone replicas)
+                            and (not prefer_replica or nr ~= master)
+                        then
+                            -- reset cursor to the main position
+                            replicaset.balance_i = 1
+                        else
+                            -- restore rr-cursor position
+                            replicaset.balance_i = cbi
+                        end
+                    end
+
                     return r
                 end
             end
@@ -637,6 +659,7 @@ local function replicaset_template_multicallro(prefer_replica, balance)
                 end
             end
             opts.timeout = timeout
+
             net_status, storage_status, retval, err =
                 replica_call(replica, func, args, opts)
             now = fiber_clock()
@@ -990,8 +1013,10 @@ local replicaset_mt = {
         callrw = replicaset_master_call;
         callro = replicaset_template_multicallro(false, false);
         callbro = replicaset_template_multicallro(false, true);
+        callbzro = replicaset_template_multicallro(false, 2);
         callre = replicaset_template_multicallro(true, false);
         callbre = replicaset_template_multicallro(true, true);
+        callbzre = replicaset_template_multicallro(true, 2);
         map_call = replicaset_map_call,
         update_master = replicaset_update_master,
     };
@@ -1235,6 +1260,7 @@ local function buildall(sharding_cfg)
         local function replica_cmp_weight(r1, r2)
             -- Master has priority over replicas with the same
             -- weight.
+
             if r1.weight == r2.weight then
                 return r1 == new_replicaset.master
             else
--- a/vshard/router/init.lua
+++ b/vshard/router/init.lua
@@ -576,11 +576,15 @@ local function router_call_impl(router, bucket_id, mode, prefer_replica,
     local call
     if mode == 'read' then
         if prefer_replica then
-            if balance then
+            if balance == 2 then
+                call = 'callbzre'
+            elseif balance then
                 call = 'callbre'
             else
                 call = 'callre'
             end
+        elseif balance == 2 then
+            call = 'callbzro'
         elseif balance then
             call = 'callbro'
         else
@@ -698,6 +702,10 @@ local function router_callbro(router, bucket_id, ...)
     return router_call_impl(router, bucket_id, 'read', false, true, ...)
 end
 
+local function router_callbzro(router, bucket_id, ...)
+    return router_call_impl(router, bucket_id, 'read', false, 2, ...)
+end
+
 local function router_callrw(router, bucket_id, ...)
     return router_call_impl(router, bucket_id, 'write', false, false, ...)
 end
@@ -708,6 +716,10 @@ end
 
 local function router_callbre(router, bucket_id, ...)
     return router_call_impl(router, bucket_id, 'read', true, true, ...)
+end
+
+local function router_callbzre(router, bucket_id, ...)
+    return router_call_impl(router, bucket_id, 'read', true, 2, ...)
 end
 
 local function router_call(router, bucket_id, opts, ...)
@@ -1660,6 +1672,8 @@ local router_mt = {
         call = router_make_api(router_call),
         callro = router_make_api(router_callro),
         callbro = router_make_api(router_callbro),
+        callbzro = router_make_api(router_callbzro),
+        callbzre = router_make_api(router_callbzre),
         callrw = router_make_api(router_callrw),
         callre = router_make_api(router_callre),
         callbre = router_make_api(router_callbre),


Serpentian:
Thanks for the patch! Please get acquainted with [code review procedure](https://github.com/tarantool/tarantool/wiki/Code-review-procedure) and make sure your PR comply with our styling rules. For example:

* A patch should include tests for the new features;
* The title of the commit should be `storage: add callro methods based on node zone`. The commit message should explain why and what you did;
* We don't usually use comments after a line of code.

vshard/router/init.lua
```diff
@@ -710,6 +718,10 @@ local function router_callbre(router, bucket_id, ...)
     return router_call_impl(router, bucket_id, 'read', true, true, ...)
 end
 
+local function router_callbzre(router, bucket_id, ...)
+    return router_call_impl(router, bucket_id, 'read', true, 2, ...)
```
Serpentian, line 722: If we want to introduce such feature, we should probably make API more explicit. Using `2` as the value for `balance` doesn't seem too good. A suppose, the better solution would be to make `balance` a string, and if its value is something like `prefer_zone`, balance replicas in a way, you propose. In all other cases of `balance`, simple rr can be used

eugenepaniot:
Sorry, very busy now to implement tests. Will try to do it soon. It will be very helpful if someone from QA Team helps with it. Thank you.

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
...
         if balance then
+            local prefered_zone = replicaset.priority_list[1].zone
             local i = #replicaset.priority_list
             while i > 0 do
                 r = replicaset_balance_replica(replicaset)
                 i = i - 1
                 if r:is_connected() and (not prefer_replica or r ~= master) and
-                   replica_check_backoff(r, now) then
+                    replica_check_backoff(r, now)
+                then
```
Gerold103, line 589: Please, revert unnecessary changes like these. These 3 lines didn't have to change at all. Besides, we never put `then` on a separate line. Lets keep it consistent with our code style.

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
     local function pick_next_replica(replicaset, now)
         local r
         local master = replicaset.master
+
```
Gerold103, line 580: Please, drop not needed empty lines.

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
...
+                            and (not prefer_replica or nr ~= master)
+                        then
+                            -- Reset cursor to the main position if next replica is in different zone.
+                            replicaset.balance_i = 1
+                        else
+                            -- Restore rr-cursor position.
+                            replicaset.balance_i = cbi
+                        end
+                    end
+
```
Gerold103, line 606: You should strive to keep the code indentation to the minimum. For example, here it is easily achievable by just handling simple cases first and doing early return. Then you can handle the zone balancing with -4 spaces indentation:
```Diff
                 if r:is_connected() and (not prefer_replica or r ~= master) and
-                    replica_check_backoff(r, now)
-                then
+                   replica_check_backoff(r, now) then
+                    if balance ~= 'prefered_zone' or not prefered_zone then
+                        return r
+                    end
                     -- Pick a replica according prefered zone (highest priority replica zone) in round-robin manner
-                    if balance == "prefer_zone" and prefered_zone then
-                        local cbi = replicaset.balance_i
-                        local nr = replicaset_balance_replica(replicaset)
+                    local cbi = replicaset.balance_i
+                    local nr = replicaset_balance_replica(replicaset)
 
-                        if prefered_zone and r.zone and r.zone == prefered_zone
-                            and nr.zone and nr.zone ~= prefered_zone
-                            and (not prefer_replica or nr ~= master)
-                        then
-                            -- Reset cursor to the main position if next replica is in different zone.
-                            replicaset.balance_i = 1
-                        else
-                            -- Restore rr-cursor position.
-                            replicaset.balance_i = cbi
-                        end
+                    if prefered_zone and r.zone and r.zone == prefered_zone
+                        and nr.zone and nr.zone ~= prefered_zone
+                        and (not prefer_replica or nr ~= master)
+                    then
+                        -- Reset cursor to the main position if next replica is in different zone.
+                        replicaset.balance_i = 1
+                    else
+                        -- Restore rr-cursor position.
+                        replicaset.balance_i = cbi
                     end
-
```

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
...
+            local prefered_zone = replicaset.priority_list[1].zone
             local i = #replicaset.priority_list
             while i > 0 do
                 r = replicaset_balance_replica(replicaset)
                 i = i - 1
                 if r:is_connected() and (not prefer_replica or r ~= master) and
-                   replica_check_backoff(r, now) then
+                    replica_check_backoff(r, now)
+                then
+                    -- Pick a replica according prefered zone (highest priority replica zone) in round-robin manner
```
Gerold103, line 590: Please, keep line width limit at 80 symbols.

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
...
+                    if balance == "prefer_zone" and prefered_zone then
+                        local cbi = replicaset.balance_i
+                        local nr = replicaset_balance_replica(replicaset)
+
+                        if prefered_zone and r.zone and r.zone == prefered_zone
+                            and nr.zone and nr.zone ~= prefered_zone
```
Gerold103, line 596: No need to check `r.zone and r.zone == ...`. You can just do `r.zone == ...`. `nil` is comparable with other types just fine.

vshard/replicaset.lua
```diff
@@ -577,13 +577,33 @@ local function replicaset_template_multicallro(prefer_replica, balance)
...
+                            and (not prefer_replica or nr ~= master)
+                        then
+                            -- Reset cursor to the main position if next replica is in different zone.
```
Gerold103, line 599: Replicas are not sorted by zones. Hence if in the replica list the first one is in zone1, then one in zone2, and others in zone1, your code will skip the tail-replicas always. That doesn't look good.

There are several solutions - don't touch balance_i. Simply skip replicas from other zones manually. Or sort the replicas by zones in the replica list. Or add a key-value dictionary to each replicaset object, where key = zone name, value = replica list in this zone.

I think it might be too complicated to do the last 2 options, so I suggest to just skip not fitting replicas during the iteration. It seems this code solves your problem and is multiple times shorter:
```Diff
@@ -577,33 +577,18 @@ local function replicaset_template_multicallro(prefer_replica, balance)
     local function pick_next_replica(replicaset, now)
         local r
         local master = replicaset.master
-
         if balance then
-            local prefered_zone = replicaset.priority_list[1].zone
+            local prefered_zone
+            if balance == 'prefered_zone' then
+                prefered_zone = replicaset.priority_list[1].zone
+            end
             local i = #replicaset.priority_list
             while i > 0 do
                 r = replicaset_balance_replica(replicaset)
                 i = i - 1
                 if r:is_connected() and (not prefer_replica or r ~= master) and
-                    replica_check_backoff(r, now)
-                then
-                    -- Pick a replica according prefered zone (highest priority replica zone) in round-robin manner
-                    if balance == "prefer_zone" and prefered_zone then
-                        local cbi = replicaset.balance_i
-                        local nr = replicaset_balance_replica(replicaset)
-
-                        if prefered_zone and r.zone and r.zone == prefered_zone
-                            and nr.zone and nr.zone ~= prefered_zone
-                            and (not prefer_replica or nr ~= master)
-                        then
-                            -- Reset cursor to the main position if next replica is in different zone.
-                            replicaset.balance_i = 1
-                        else
-                            -- Restore rr-cursor position.
-                            replicaset.balance_i = cbi
-                        end
-                    end
-
+                   replica_check_backoff(r, now) and
+                   (not prefered_zone or prefered_zone == r.zone) then
                     return r
                 end
             end
```
Of course, we would need tests to be sure.

vshard/replicaset.lua
```diff
@@ -1235,6 +1258,7 @@ local function buildall(sharding_cfg)
         local function replica_cmp_weight(r1, r2)
             -- Master has priority over replicas with the same
             -- weight.
+
```
Gerold103, line 1261: Please, drop this line.

vshard/router/init.lua
```diff
@@ -576,11 +576,15 @@ local function router_call_impl(router, bucket_id, mode, prefer_replica,
     local call
     if mode == 'read' then
         if prefer_replica then
-            if balance then
+            if balance == "prefer_zone" then
```
Gerold103, line 579: Please, change 'balance' to numeric IDs. String comparison in Lua not always works for constant time and might degrade to linear time. Like this:
```Diff
@@ -65,6 +65,10 @@ local fiber_yield = fiber.yield
 local fiber_cond_wait = util.fiber_cond_wait
 local gsc = util.generate_self_checker
 
+local BALANCE_NONE = 0
+local BALANCE_RR = 1
+local BALANCE_BEST_ZONE_RR = 2
+
 --
 -- on_connect() trigger for net.box
 --
@@ -578,7 +582,7 @@ local function replicaset_template_multicallro(prefer_replica, balance)
         local r
         local master = replicaset.master
 
-        if balance then
+        if balance ~= BALANCE_NONE then
             local prefered_zone = replicaset.priority_list[1].zone
             local i = #replicaset.priority_list
             while i > 0 do
@@ -588,7 +592,7 @@ local function replicaset_template_multicallro(prefer_replica, balance)
                     replica_check_backoff(r, now)
                 then
                     -- Pick a replica according prefered zone (highest priority replica zone) in round-robin manner
-                    if balance == "prefer_zone" and prefered_zone then
+                    if balance == BALANCE_BEST_ZONE_RR and prefered_zone then
                         local cbi = replicaset.balance_i
                         local nr = replicaset_balance_replica(replicaset)
 
@@ -1009,12 +1013,12 @@ local replicaset_mt = {
         wait_connected_all = replicaset_wait_connected_all,
         call = replicaset_master_call;
         callrw = replicaset_master_call;
-        callro = replicaset_template_multicallro(false, false);
-        callbro = replicaset_template_multicallro(false, true);
-        callbzro = replicaset_template_multicallro(false, "prefer_zone");
-        callre = replicaset_template_multicallro(true, false);
-        callbre = replicaset_template_multicallro(true, true);
-        callbzre = replicaset_template_multicallro(true, "prefer_zone");
+        callro = replicaset_template_multicallro(false, BALANCE_NONE);
+        callbro = replicaset_template_multicallro(false, BALANCE_RR);
+        callbzro = replicaset_template_multicallro(false, BALANCE_BEST_ZONE_RR);
+        callre = replicaset_template_multicallro(true, BALANCE_NONE);
+        callbre = replicaset_template_multicallro(true, BALANCE_RR);
+        callbzre = replicaset_template_multicallro(true, BALANCE_BEST_ZONE_RR);
         map_call = replicaset_map_call,
         update_master = replicaset_update_master,
     };
```

Gerold103 (commented):
Thanks for the patch! See some comments below. You also need to add tests. We don't have a special QA team for writing them. Authors of patches do this. Without tests merging this PR might take unknown time, because we would need to write them ourselves and that is complicated, since this feature is not planned anywhere. How to write them - look at examples in test/ folder. You need the new sub-folders ending with `-luatest` in there.

vshard/router/init.lua
eugenepaniot: Thank you for the comments. Will check it.

Issue #384, opened 2022-11-23
flaky test: router/map-reduce


Gerold103:
```
router/map-reduce.test.lua                                      [ fail ]

Test failed! Result content mismatch:
--- router/map-reduce.result	Wed Nov 23 08:58:11 2022
+++ /home/runner/work/vshard/vshard/test/var/rejects/router/map-reduce.reject	Wed Nov 23 09:02:15 2022
@@ -186,6 +186,7 @@
  | ...
 ok, err = vshard.router.map_callrw('echo', {1}, timeout_opts)
  | ---
+ | - error: '/home/runner/work/vshard/vshard/vshard/router/init.lua:820: Usage: future:wait_result(timeout)'
  | ...
 assert(not ok and util.is_timeout_error(err))
  | ---
@@ -203,7 +204,7 @@
  | ...
 test_run:wait_cond(function() return lref.count == 0 end)
  | ---
- | - true
+ | - false
  | ...
```
Here is the reason: https://github.com/tarantool/tarantool/issues/7970.

Nonetheless we can't just let it be fixed in core. We have to support older versions. The proposal is to add a new `vshard.util` functions for waiting on the futures safely. Similar to `vshard.util.fiber_cond_wait()` but it will be `vshard.util.future_wait()`. Should be used instead of all `:wait_result()` places in vshard.

PR #392, opened 2023-01-11
test: fix flaky router/map_reduce

Serpentian:
Currently map_reduce (or any other test related to the functions, in
which loop over futures is used) can fail with the usage exception
due to the fact, that negative values are not allowed in wait_result.

This commit introduces new util function, which is supposed to be used
instead of wait_result. It throws box.error.TIMEOUT in case of a
negative value as argument.

Note, that errors are inconsistent: the function may return either
TimedOut error (which is one, returned from the original wait_result)
or box.error.TIMEOUT.

Closes https://github.com/tarantool/vshard/issues/384

NO_DOC=bugfix

### DIFF
--- a/test/unit-luatest/util_test.lua
+++ b/test/unit-luatest/util_test.lua
@@ -1,7 +1,9 @@
 local t = require('luatest')
 local luuid = require('uuid')
 local server = require('test.luatest_helpers.server')
+local vtest = require('test.luatest_helpers.vtest')
 local vutil = require('vshard.util')
+local net = require('net.box')
 
 local test_group = t.group('util')
 
@@ -208,4 +210,55 @@ end
 test_group.test_index = function(g)
     test_index_templ(g, 'memtx')
     test_index_templ(g, 'vinyl')
+end
+
+test_group.test_future_wait_exception = function(g)
+    -- Before 2.10.3 executing future_wait in trigger cause silent hang.
+    t.run_only_if(vutil.version_is_at_least(2, 10, 3, nil, 0, 0))
+    local conn = net.connect(g.server.net_box_uri, {wait_connected = false})
+    t.assert_equals(conn.state, 'initial')
+
+    -- Test that no exceptions're thrown from future_wait
+    local _, err
+    conn:on_connect(function()
+        local f = conn:eval('return true', {}, {is_async = true})
+        _, err = vutil.future_wait(f)
+    end)
+
+    conn:wait_connected()
+    t.assert_str_contains(err.message, 'Synchronous requests are not allowed')
+end
+
+test_group.test_future_wait_timeout = function(g)
+    g.server:exec(function()
+        local fiber = require('fiber')
+        rawset(_G, 'do_sleep', true)
+        rawset(_G, 'fiber_sleep', function()
+            while _G.do_sleep do
+                fiber.sleep(0.001)
+            end
+            return true
+        end)
+    end)
+
+    local function stop_sleep()
+        g.server:exec(function()
+            _G.do_sleep = false
+        end)
+    end
+
+    local conn = net.connect(g.server.net_box_uri)
+    local f = conn:call('_G.fiber_sleep', {}, {is_async = true})
+
+    -- Default timeout error
+    local _, err = vutil.future_wait(f, 1e-6)
+    t.assert_equals(vtest.error_is_timeout(err), true)
+
+    -- Negative timeout should return timeout error too
+    _, err = vutil.future_wait(f, -1)
+    t.assert_equals(vtest.error_is_timeout(err), true)
+
+    stop_sleep()
+    -- Everything is all right, wait for result
+    t.assert_equals(vutil.future_wait(f)[1], true)
 end
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -63,6 +63,7 @@ local util = require('vshard.util')
 local fiber_clock = fiber.clock
 local fiber_yield = fiber.yield
 local fiber_cond_wait = util.fiber_cond_wait
+local future_wait = util.future_wait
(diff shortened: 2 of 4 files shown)


vshard/util.lua
```diff
@@ -279,6 +279,24 @@ local function fiber_cond_wait(cond, timeout)
...
+    local res, err
+    -- Handle negative timeouts in order to avoid usage error.
+    if timeout and timeout < 0 then
+        err = lerror.timeout()
+        goto fail
+    end
+    res, err = cond:wait_result(timeout)
+    if not err then
+        return res
+    end
```
Serpentian, line 295: As I said in commit message, errors are inconsistent. Do we want them to be?

Frankly speaking, I don't see it as a big problem, as tnt is a one big inconsistency. If we ever need to deal with timeout error processing, we'll anyway have to use the function, similar to `vtest.error_is_timeout`.

But if we want it, here's the way we may do it:

```diff
diff --git a/vshard/util.lua b/vshard/util.lua
index 45abca2..6e212b2 100644
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -293,6 +293,10 @@ local function future_wait(cond, timeout)
     if not err then
         return res
     end
+
+    if err.type == 'TimedOut' then
+        err = lerror.timeout()
+    end
 ::fail::
     return nil, err
 end

```

test/unit-luatest/util_test.lua
```diff
@@ -209,3 +211,44 @@ test_group.test_index = function(g)
     test_index_templ(g, 'memtx')
     test_index_templ(g, 'vinyl')
 end
+
+test_group.test_future_wait_exception = function(g)
+    -- Before 2.10.3 executing future_wait in trigger cause silent hang.
+    t.run_only_if(vutil.version_is_at_least(2, 10, 3, nil, 0, 0))
```
Serpentian, line 217: Corresponding commit: https://github.com/tarantool/tarantool/commit/0d944f901cf5400d5832a5528780bb3ecd03ae4b

vshard/util.lua
```diff
@@ -279,6 +279,32 @@ local function fiber_cond_wait(cond, timeout)
...
 
+local function future_wait_xc(cond, timeout)
+    -- Handle negative timeouts in order to avoid usage error.
+    -- The idea is the same as in fiber_cond_wait_xc function.
+    if timeout and timeout < 0 then
+        error(lerror.timeout())
+    end
+    -- pcall() is inevitable as wait_result can throw an error
+    -- if executed in net.box's trigger for example.
+    local res, err = cond:wait_result(timeout)
```
Gerold103, line 290: The comment should be where you do `pcall()`. Here is looks strange that you mention `pcall()` and don't do it.

test/unit-luatest/util_test.lua
```diff
@@ -209,3 +211,44 @@ test_group.test_index = function(g)
...
+test_group.test_future_wait_timeout = function(g)
+    g.server:exec(function()
+        rawset(_G, 'fiber_sleep', function(time)
+            require('fiber').sleep(time)
+            return true
+        end)
+    end)
+
+    local conn = net.connect(g.server.net_box_uri)
+    conn:wait_connected()
```
Gerold103, line 242: You don't need to wait for connection. `netbox.connect` does it by default.

test/unit-luatest/util_test.lua
```diff
@@ -209,3 +211,44 @@ test_group.test_index = function(g)
...
+    local f = conn:call('_G.fiber_sleep', {0}, {is_async = true})
+    t.assert_equals(vutil.future_wait(f)[1], true)
+
+    f = conn:call('_G.fiber_sleep', {math.huge}, {is_async = true})
+    local _, err = vutil.future_wait(f, 1e-6)
+    t.assert_equals(vtest.error_is_timeout(err), true)
+    f:discard()
```
Gerold103, line 249: You probably shouldn't discard it, if you use it again below.

test/unit-luatest/util_test.lua
```diff
@@ -209,3 +211,44 @@ test_group.test_index = function(g)
...
+
+    conn:wait_connected()
+    t.assert_str_contains(err.message, 'Synchronous requests are not allowed')
+end
+
+test_group.test_future_wait_timeout = function(g)
+    g.server:exec(function()
+        rawset(_G, 'fiber_sleep', function(time)
+            require('fiber').sleep(time)
+            return true
```
Gerold103, line 237: In the test case below you send an infinite timeout. Even if you later call `future:discard()`, this fiber will keep hanging in `sleep` until restart. I suggest to add a global flag here like: `while _G.do_sleep do fiber.sleep(0.001) end`. Then in the end you do `_G.do_sleep = false` to unblock the fiber and properly wait for the future to end.

The reason why we shouldn't leave such request-fibers is that they might affect global limits like `box.cfg.net_msg_max`.

vshard/replicaset.lua
```diff
@@ -717,7 +717,7 @@ local function replicaset_map_call(replicaset, func, args, opts)
     --
     map = table.new(0, replica_count)
     for uuid, future in pairs(futures) do
-        res, err = future:wait_result(timeout)
+        res, err = util.future_wait(future, timeout)
```
Gerold103, line 720: Lets add this `local future_wait = util.future_wait` to the beginning of the file (same place where `fiber_cond_wait` is saved) and use this `future_wait()` everywhere. The reason is that in Lua a table index operation like `table.key` is not free, it is computed at runtime on each use.

Same in the router.

Gerold103 (changes requested):
Thanks for the patch! Only some small points, good work.

Serpentian:
Everything is fixed

Gerold103 (approved):
Thanks for the fixes!

PR #393, opened 2023-01-11
ci: fix ubuntu version

Serpentian:
Currently when running CI on 2.7 and 2.8 versions of tarantool
the 2.6.0-0-g47aa4e01e is being installed instead.

Fetching versions with setup-tarantool doesn't work properly with
ubuntu-22.04, so, let's use 20.04, as it's done in all other repos.
Let's also update setup-tarantool version while we're here.

Co-authored-by: Yaroslav Lobankov <y.lobankov@tarantool.org>

NO_TEST=ci
NO_DOC=ci

### DIFF
--- a/.github/workflows/fast_testing.yml
+++ b/.github/workflows/fast_testing.yml
@@ -27,11 +27,11 @@ jobs:
     env:
       TNT_RELEASE_PATH: /home/runner/tnt-release
 
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Serpentian:
@Gerold103, we should merge this ASAP, as all PRs are not currently tested on 2.7 and 2.8 versions

Gerold103 (approved):
Thanks for the patch!

PR #396, opened 2023-01-30
test: introduce router reload test

### LINKED ISSUE #141: Test vshard.router reload over different vshard versions
The same as #125 , but for the router module.

Serpentian:
Currently reloading from versions before commit https://github.com/tarantool/vshard/commit/dd70cfb2c5ec36ab7d5355b0024e5f6d21bb8f9f
("router: auto and manual enable/disable") doesn't work, as
variables `is_enabled` and `api_call_cache` weren't handled
properly. Let's set `is_enabled` flag to true on reload for
backward compatibility. As for `api_call_cache` it should be
set on router configuration, not its creation, as reconfiguration
is invoked anyway on reload.

The commit also introduces a test, which checks availability of
basic services and methods, when the router module is reloaded to
the new version.

Closes https://github.com/tarantool/vshard/issues/141

NO_DOC=bugfix

### DIFF
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -741,6 +741,19 @@ local function service_wait_for_activity(service, activity, opts)
     end)
 end
 
+-- Git directory of the project and data directory of the test.
+-- Used in evolution tests to fetch old versions of vshard.
+local sourcedir = fio.abspath(os.getenv('PACKPACK_GIT_SOURCEDIR') or
+                              os.getenv('SOURCEDIR'))
+if not sourcedir then
+    local script_path = debug.getinfo(1).source:match("@?(.*/)")
+    script_path = fio.abspath(script_path)
+    sourcedir = fio.abspath(script_path .. '/../../../')
+end
+
+-- May be nil, if VARDIR is not specified.
+local vardir = fio.abspath(os.getenv('VARDIR'))
+
 return {
     error_is_timeout = error_is_timeout,
     config_new = config_new,
@@ -772,4 +785,6 @@ return {
     service_wait_for_activity = service_wait_for_activity,
     wait_for_not_nil = wait_for_not_nil,
     wait_for_nil = wait_for_nil,
+    sourcedir = sourcedir,
+    vardir = vardir,
 }
--- a/test/router-luatest/reload_test.lua
+++ b/test/router-luatest/reload_test.lua
@@ -1,0 +1,266 @@
+local t = require('luatest')
+local vtest = require('test.luatest_helpers.vtest')
+local git_util = require('test.lua_libs.git_util')
+local fio = require('fio')
+
+local g = t.group('reload_router')
+local cfg_template = {
+    sharding = {
+        {
+            replicas = {
+                replica_1_a = {
+                    master = true,
+                },
+                replica_1_b = {},
+            },
+        },
+        {
+            replicas = {
+                replica_2_a = {
+                    master = true,
+                },
+            },
+        },
+    },
+    bucket_count = 100
+}
+local global_cfg
+
+g.before_all(function()
+    global_cfg = vtest.config_new(cfg_template)
+
+    -- The test works in the following directory
+    local vardir = vtest.vardir or fio.tempdir()
+    g.vshard_copy_path =  vardir .. '/vshard_copy'
+    t.assert_equals(fio.mkdir(g.vshard_copy_path), true)
+    -- Copy source to the temporary directory
+    t.assert_equals(fio.mkdir(g.vshard_copy_path .. '/.git'), true)
+    t.assert_equals(fio.copytree(vtest.sourcedir .. '/.git',
+        g.vshard_copy_path .. '/.git'), true)
+
+    -- Hash of the latest commit for testing router on the latest version.
+    g.latest_hash = git_util.log_hashes({args = '-1', dir = vtest.sourcedir})[1]
+
+    -- No need to reload storages. Just run them on the latest version.
+    vtest.cluster_new(g, global_cfg)
+    vtest.cluster_bootstrap(g, global_cfg)
+    vtest.cluster_rebalancer_disable(g)
+
+    -- Basic storage configuration
+    vtest.cluster_exec_each_master(g, function()
+        local test = box.schema.space.create('test', {format = {
+            {'id', 'unsigned'},
+            {'bucket_id', 'unsigned'},
+        }})
+
+        test:create_index('primary')
+        test:create_index('bucket_id', {unique = false, parts = {2}})
+    end)
+
+    vtest.cluster_exec_each(g, function()
+        rawset(_G, 'insert', function(space_name, tuple)
+            return box.space[space_name]:insert(tuple)
+        end)
+        rawset(_G, 'get', function(space_name, key)
+            return box.space[space_name]:get(key)
+        end)
+    end)
+end)
+
+g.after_all(function()
+    g.cluster:drop()
+end)
+
+--
+-- Reload test template:
+--     1. Invoke create_router_at:
+--         * Checkout to old version;
+--         * Create and start a router;
+--     3. Test smth on the old version;
+--     4. Invoke reload_router():
+--         * Checkout to the latest version;
+--         * Reload the module.
+--     5. Test smth on the new version
+--     6. Drop a router with vtest.drop_instance
+--
+
+local function create_router_at(hash)
+    git_util.exec('checkout', {args = hash .. ' -f', dir = g.vshard_copy_path})
+    local router = vtest.router_new(g, 'router')
+    router:exec(function(vshard_path)
+        -- Force 'require' to use new directory
+        package.path = string.format('%s/?.lua;%s/?/init.lua;',
+                                     vshard_path, vshard_path)
+        for package_name in pairs(package.loaded) do
+            if package_name:startswith('vshard') then
+                package.loaded[package_name] = nil
+            end
+        end
+    end, {g.vshard_copy_path})
+    vtest.router_cfg(router, global_cfg)
+    return router
+end
+
+--
+-- Reloads router. If service_name is provided, then
+-- the function also waits until the service is restarted.
+local function reload_router(router, service_name)
+    git_util.exec('checkout', {args = g.latest_hash .. ' -f',
+                               dir = g.vshard_copy_path})
+    router:exec(function(service_name)
+        local service
+        local internal = ivshard.router.internal
+        if service_name ~= nil then
+            service = internal.static_router[service_name]
+            ilt.assert_not_equals(service, nil)
+        end
+
+        ilt.assert_equals(ivshard.router.module_version(), 0)
+        package.loaded['vshard.router'] = nil
+        ivshard.router = require('vshard.router')
+        ilt.assert_equals(ivshard.router.module_version(), 1)
+
+        if service ~= nil then
+            ilt.helpers.retrying({timeout = ivtest.wait_timeout,
+                                  delay = ivtest.busy_step}, function()
+                if service == internal.static_router[service_name] then
+                    error('Service have not been reloaded yet')
+                end
+            end)
+        end
+    end, {service_name})
+end
+
+local function test_basic_template(router)
+    router:exec(function()
+        -- Basic test of callrw and callro
+        local bucket_id = 1
+        ivshard.router.callrw(bucket_id, 'insert', {'test', {1, bucket_id}})
+        ivshard.router.callrw(bucket_id, 'vshard.storage.sync', {})
+        local res, err = ivshard.router.callro(bucket_id, 'get', {'test', {1}})
+        ilt.assert_equals(err, nil)
+        ilt.assert_equals(res, {1, bucket_id})
+    end)
+end
+
+g.test_basic = function(g)
+    -- Latest meaningful commit:
+    --     "router: fix reload problem with global function refs".
+    local hash = '139223269cddefe2ba4b8e9f6e44712f099f4b35'
+    local router = create_router_at(hash)
+    test_basic_template(router)
+    reload_router(router)
+    test_basic_template(router)
+    vtest.drop_instance(g, router)
+end
+
+local function test_discovery_template(g, router)
+    router:exec(function(uuid)
+        local router = ivshard.router.internal.static_router
+        local discovery = router.discovery_service
+
+        -- Everything is all right
+        ivtest.service_wait_for_ok(discovery,
+            {on_yield = ivshard.router.discovery_wakeup})
+        ilt.assert_equals(router.known_bucket_count, 100)
+
+        -- Break connection and the number of known buckets
+        local conn = router.replicasets[uuid].master.conn
+        local known = router.known_bucket_count
+        router.replicasets[uuid].master.conn = nil
+        router.known_bucket_count = known - 1
+
+        ivtest.service_wait_for_error(discovery, 'Error during discovery',
+            {on_yield = ivshard.router.discovery_wakeup})
+
+        -- Restore everything
+        router.replicasets[uuid].master.conn = conn
+        router.known_bucket_count = known
+
+        -- It's all right again
+        ivtest.service_wait_for_ok(discovery,
+            {on_yield = ivshard.router.discovery_wakeup})
+    end, {g.replica_2_a:replicaset_uuid()})
+end
+
+g.test_discovery = function(g)
+    -- Service_info was introduced in the following commit,
+    -- the test cannot be run without it:
+    --     "router: add saving of background service statuses".
+    local hash = 'f5f386a5a35e6e5efd8f4f2ed1b3d208fdae9095'
+    local router = create_router_at(hash)
+    test_discovery_template(g, router)
+    reload_router(router, 'discovery_service')
+    test_discovery_template(g, router)
+    vtest.drop_instance(g, router)
+end
+
+local function test_master_search_template(g, router, auto_master_cfg)
+    vtest.router_cfg(router, auto_master_cfg)
+
+    -- Working with first replicaset (2 instances)
+    local rs_uuid = g.replica_1_a:replicaset_uuid()
+    local master_uuid = g.replica_1_a:instance_uuid()
+    local replica_uuid = g.replica_1_b:instance_uuid()
+
+    router:exec(function()
+        local router = ivshard.router.internal.static_router
+        ivtest.service_wait_for_ok(router.master_search_service,
+            {on_yield = ivshard.router.master_search_wakeup})
+        for _, rs in pairs(router.replicasets) do
+            ilt.assert_not_equals(rs.master, nil)
+        end
+
+        -- Stop master search
+        ivshard.router.internal.errinj.ERRINJ_MASTER_SEARCH_DELAY = true
+        ilt.helpers.retrying({timeout = ivtest.wait_timeout,
+                              delay = ivtest.busy_step}, function()
+            local errinj = ivshard.router.internal.errinj
+            if errinj.ERRINJ_MASTER_SEARCH_DELAY ~= 'in' then
+                error('Master search is not stopped yet')
+            end
+        end)
+    end)
+
+    -- Change master and wait, until it's found
+    global_cfg.sharding[rs_uuid].replicas[master_uuid].master = false
+    global_cfg.sharding[rs_uuid].replicas[replica_uuid].master = true
+    vtest.cluster_cfg(g, global_cfg)
+
+    router:exec(function(rs, master)
+        -- Continue and check master search
+        ivshard.router.internal.errinj.ERRINJ_MASTER_SEARCH_DELAY = false
+        local router = ivshard.router.internal.static_router
+        ivtest.service_wait_for_new_ok(router.master_search_service,
+            {on_yield = ivshard.router.master_search_wakeup})
+        ilt.assert_equals(master, router.replicasets[rs].master.uuid)
+    end, {rs_uuid, replica_uuid})
+
+    -- Restore configuration of the cluster
+    global_cfg.sharding[rs_uuid].replicas[master_uuid].master = true
+    global_cfg.sharding[rs_uuid].replicas[replica_uuid].master = false
+    vtest.cluster_cfg(g, global_cfg)
+end
+
+g.test_master_search = function(g)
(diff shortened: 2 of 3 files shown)


Serpentian:
1. Now the test just checks basic work of master_search and discovery. What else do we want to test in reloading over versions?
2. Is it ok to use old git infrastructure? IMO, it makes no sense to rewrite smth, that works.

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+local global_cfg
+
+g.before_all(function()
+    global_cfg = vtest.config_new(cfg_template)
+
+    -- We cannot work at the same directory, where the test is.
+    rawset(_G, 'vshard_copy_path', lutil.BUILDDIR ..
+           '/test/var/vshard_git_tree_copy_router_luatest')
+    -- Cleanup the dir after previous build
+    os.execute('rm -rf ' .. _G.vshard_copy_path)
```
Gerold103, line 40: This shouldn't be needed, dangerous. Please, try to keep all files and folders local to the folder created for this run by luatest. I know it is done this way in `lua_libs/util.lua`, but `util.lua` is basically an anti-pattern how things should be done. I also know that I did it, now I see my design mistakes clearly.

test/lua_libs/util.lua
```diff
@@ -1,7 +1,7 @@
 local fiber = require('fiber')
```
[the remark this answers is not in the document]
Serpentian: Are you sure? The test indeed takes a significant amount of time to run (about 30 seconds on my machine), but it doesn't hang on the following setup:
```
 | Tarantool 3.0.0-entrypoint-183-g55295f5f1
 | Target: Linux-x86_64-Debug
```

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
+local t = require('luatest')
+local vtest = require('test.luatest_helpers.vtest')
+local os = require('os')
+
+-- Legacy utils
+local lutil = require('test.lua_libs.util')
```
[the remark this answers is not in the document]
Serpentian: I copy-pasted `SOURCEDIR` to `vtest`, created `VARDIR` there too. I don't really need `BUILDDIR`. As for `git_checkout`, I don't really like function from `lua_libs/util` and I don't see any sense in creating similar one: we only need to create a directory in `before_all` part, after which we can use `git_utils.exec` to checkout between versions in that directory.

Speaking of `advice.detachedHead false`, I just really don't like that big warning, which pops up, when the test is run with `--verbose` flag:

<details><summary>Verbose test output</summary>

```
[001] router-luatest/reload_test.lua                                  
[001] Note: switching to '139223269cddefe2ba4b8e9f6e44712f099f4b35'.
[001] 
[001] You are in 'detached HEAD' state. You can look around, make experimental
[001] changes and commit them, and you can discard any commits you make in this
[001] state without impacting any branches by switching back to a branch.
[001] 
[001] If you want to create a new branch to retain commits you create, you may
[001] do so (now or later) by using -c with the switch command. Example:
[001] 
[001]   git switch -c <new-branch-name>
[001] 
[001] Or undo this operation with:
[001] 
[001]   git switch -
[001] 
[001] Turn off this advice by setting config variable advice.detachedHead to false
[001] 
[001] HEAD is now at 1392232 router: fix reload problem with global function refs
[001] Previous HEAD position was 1392232 router: fix reload problem with global function refs
[001] HEAD is now at 8d2bb27 router: fix reload
[001] Previous HEAD position was 8d2bb27 router: fix reload
[001] HEAD is now at f5f386a router: add saving of background service statuses
[001] TAP version 13
[001] 1..3
[001] # Started on Wed Apr 19 04:51:38 2023
[001] # Starting group: reload_router
[001] ok     1	reload_router.test_basic
[001] Previous HEAD position was f5f386a router: add saving of background service statuses
[001] HEAD is now at 8d2bb27 router: fix reload
[001] Previous HEAD position was 8d2bb27 router: fix reload
[001] HEAD is now at f5f386a router: add saving of background service statuses
[001] Previous HEAD position was f5f386a router: add saving of background service statuses
[001] HEAD is now at 8d2bb27 router: fix reload
[001] ok     2	reload_router.test_discovery
[001] ok     3	reload_router.test_master_search
[001] # Ran 3 tests in 32.197 seconds, 3 successes, 0 failures
[001] [ pass ]
```

</details>

Without it we clearly can see, which versions are tested (`HEAD is now ...`). But I don't mind dropping it.

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+}
+
+local global_cfg
+
+g.before_all(function()
+    global_cfg = vtest.config_new(cfg_template)
+
+    -- We cannot work at the same directory, where the test is.
+    rawset(_G, 'vshard_copy_path', lutil.BUILDDIR ..
+           '/test/var/vshard_git_tree_copy_router_luatest')
```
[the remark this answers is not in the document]
Serpentian: Now `VARDIR` is used

test/reload_evolution-luatest/reload_router_test.lua
Serpentian: Now `VARDIR` is used

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
+local t = require('luatest')
```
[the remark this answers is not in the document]
Serpentian: Done

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+--     4. Invoke reload_router():
+--         * Checkout to the latest commit;
+--         * Reload the module.
+--     5. Test smth on the new version
+--     6. Drop a router.
+--
+
+local function before_trigger(g, hash)
+    -- Initial version: 0.1.18
+    local init_hash = hash or 'c910cd5'
```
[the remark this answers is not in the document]
Serpentian: Done

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+    lgit_util.exec('checkout', {args = init_hash .. ' -f',
+                                dir = _G.vshard_copy_path})
+    -- Router over old vshard
+    g.router = vtest.router_new(g, 'router', global_cfg)
```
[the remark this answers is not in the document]
Serpentian: I see it now. Actually, it looks like the test didn't work at all in the previous version: I found several bugs in router reload, let's fix them alongside.

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+    reload_router(g)
+    test_basic_template(g)
+end
+
+local function test_discovery_template(g)
+    -- We need to wait, until all of the services are indeed reloaded
+    -- (or started). Otherwise, we may take an old instance of service,
+    -- which cannot be used to monitor status of the service as it is
+    -- not updated.
+    t.helpers.retrying({timeout = 10}, function()
```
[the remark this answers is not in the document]
Serpentian: Fixed

test/reload_evolution-luatest/reload_router_test.lua
```diff
@@ -0,0 +1,277 @@
...
+--     2. Create and start a router;
+--     3. Test smth on the old version;
+--     4. Invoke reload_router():
+--         * Checkout to the latest commit;
+--         * Reload the module.
+--     5. Test smth on the new version
+--     6. Drop a router.
+--
+
+local function before_trigger(g, hash)
```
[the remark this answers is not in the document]
Serpentian: Done

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,263 @@
...
+--
+
+local function create_router_at(hash)
+    git_util.exec('checkout', {args = hash .. ' -f', dir = vshard_copy_path})
+    local router = vtest.router_new(g, 'router-'..hash)
+    router:exec(function(vshard_path)
+        -- Force 'require' to use new directory
+        package.path = string.format('%s/?.lua;%s/?/init.lua;',
+                                     vshard_path, vshard_path)
+        package.loaded['vshard.router'] = nil
```
olegrok, line 93: Probably you need to clear all packages started with vshard. Otherwise old vshard.router could use new vshard.consts/utils/...

olegrok (commented):
Thanks for your patch. I left several questions.

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,263 @@
...
+local global_cfg
+local latest_hash
+local vshard_copy_path
+
+g.before_all(function()
+    global_cfg = vtest.config_new(cfg_template)
+
+    -- The test works in the following directory
+    vshard_copy_path = vtest.VARDIR..'/vshard_copy_router_luatest'
+    os.execute('mkdir ' .. vshard_copy_path)
```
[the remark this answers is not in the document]
Serpentian: Great proposal. Fixed

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,263 @@
...
+    -- Copy sources to the new location
+    os.execute('cp -rf ' .. vtest.SOURCEDIR .. '/.git ' .. vshard_copy_path)
```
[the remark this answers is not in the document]
Serpentian: We can also use `fio.copytree` instead of `os.execute("cp -r")` and check its return code

test/router-luatest/reload_test.lua
Serpentian: Sounds reasonable. Fixed

test/luatest_helpers/vtest.lua
```diff
@@ -741,6 +741,18 @@ local function service_wait_for_activity(service, activity, opts)
...
+-- Used in evolution tests to fetch old versions of vshard.
+local SOURCEDIR = fio.abspath(os.getenv('PACKPACK_GIT_SOURCEDIR') or
+                              os.getenv('SOURCEDIR'))
+if not SOURCEDIR then
+    local script_path = debug.getinfo(1).source:match("@?(.*/)")
+    script_path = fio.abspath(script_path)
+    SOURCEDIR = fio.abspath(script_path .. '/../../../')
+end
+
+local VARDIR = fio.abspath(os.getenv('VARDIR') or './var')
```
olegrok, line 754: Please describe why do you need this directory (as for SOURCEDIR). And probably why can't you use temporary directory for that purpose.

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+                },
+            },
+        },
+    },
+    bucket_count = 100
+}
+
+local global_cfg
+local latest_hash
+local vshard_copy_path
```
olegrok, line 30: I don't know are there any guidelines but cartridge tests save such variables directly to `g` (e.g. https://github.com/tarantool/cartridge/blob/5569997bac5ad74c792442d914d3e6d20c3df604/test/integration/restart_replication_test.lua#L20). Up to you.

test/router-luatest/reload_test.lua
olegrok: Seems here you need to do something like:
```
for package_name in pairs(package.loaded) do
    if package_name:startswith('vshard') then
        package.loaded[package_name] = nil
    end
end
```

I don't see such changes

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+
+    -- Working with first replicaset (2 instances)
+    local rs_uuid, master_uuid = unpack(g.replica_1_a:exec(function()
+        local info = box.info()
+        return {info.cluster.uuid, info.uuid}
+    end))
+
+    local replica_uuid = g.replica_1_b:exec(function()
+        return box.info.uuid
+    end)
```
olegrok, line 198: Can you replace it with `instance:call('info')`?

olegrok (commented):
Thanks for changes. I left several comments/questions.

test/router-luatest/reload_test.lua
Serpentian: I applied the code, you proposed. But I thought  `package.loaded["vshard"] = nil` does the same thing, but with reloading `vshard/init.lua` too. Is it true?

test/luatest_helpers/vtest.lua
Serpentian: Actually, I don't. Temporary directories will work just fine, they're unique enough to allow running test in parallel. Thank you for idea

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,259 @@
...
+    g.latest_hash = git_util.log_hashes({args = '-1', dir = vtest.SOURCEDIR})[1]
+
+    -- No need to reload storages. Just run them on the latest version.
+    vtest.cluster_new(g, global_cfg)
+    vtest.cluster_bootstrap(g, global_cfg)
+    vtest.cluster_rebalancer_disable(g)
+
+    -- Basic storage configuration
+    vtest.cluster_exec_each_master(g, function()
+        box.schema.space.create('test'):create_index('primary')
```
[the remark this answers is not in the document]
Serpentian: Definitely. Thank you a lot for review

olegrok (approved):
Thanks for your changes

test/luatest_helpers/vtest.lua
```diff
@@ -741,6 +741,16 @@ local function service_wait_for_activity(service, activity, opts)
...
 end
 
+-- Git directory of the project and data directory of the test.
+-- Used in evolution tests to fetch old versions of vshard.
+local SOURCEDIR = fio.abspath(os.getenv('PACKPACK_GIT_SOURCEDIR') or
+                              os.getenv('SOURCEDIR'))
+if not SOURCEDIR then
+    local script_path = debug.getinfo(1).source:match("@?(.*/)")
+    script_path = fio.abspath(script_path)
+    SOURCEDIR = fio.abspath(script_path .. '/../../../')
```
Gerold103, line 751: Nit - lets call it `sourcedir`. Uppercase is usually for constants.

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+    vtest.drop_instance(g, router)
+end
+
+local function test_discovery_template(g, router)
+    -- We need to wait, until all of the services are indeed reloaded or started
+    t.helpers.retrying({timeout = vtest.wait_timeout}, function()
+        if not router:grep_log('discovery_f has been started', 1024) then
+            error("Discovery haven't started yet")
+        end
+    end)
```
Gerold103, line 142: Can't you use `vshard.router.info.services` instead of the logs?

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+end
+
+local function test_master_search_template(g, router, auto_master_cfg)
+    vtest.router_cfg(router, auto_master_cfg)
+
+    -- Working with first replicaset (2 instances)
+    local rs_uuid, master_uuid = unpack(g.replica_1_a:exec(function()
+        local info = box.info()
+        return {info.cluster.uuid, info.uuid}
+    end))
```
Gerold103, line 194: Wouldn't it work like this (without unpack)?
```Lua
    local rs_uuid, master_uuid = g.replica_1_a:exec(function()
        local info = box.info()
        return info.cluster.uuid, info.uuid
    end)
```

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+    -- Restore configuration of the cluster
+    global_cfg.sharding[rs_uuid].replicas[master_uuid].master = true
+    global_cfg.sharding[rs_uuid].replicas[replica_uuid].master = false
+    vtest.cluster_cfg(g, global_cfg)
+
+    router:exec(continue_and_check_master_search, {rs_uuid, master_uuid})
+end
+
+g.test_master_search = function(g)
+    local hash = 'f5f386a5a35e6e5efd8f4f2ed1b3d208fdae9095'
```
Gerold103, line 246: Please, add the commit message here in a comment.

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,265 @@
...
+    for _, rs in pairs(auto_master_cfg_template.sharding) do
+        rs.master = 'auto'
+        for _, r in pairs(rs.replicas) do
+            r.master = nil
+        end
+    end
+    local auto_master_cfg = vtest.config_new(auto_master_cfg_template)
+
+    -- Service_info was introduced in the following commit:
+    --     "router: add saving of background service statuses".
```
Gerold103, line 260: Why mention it if this test case is about master search?

vshard/router/init.lua
```diff
@@ -1709,6 +1695,20 @@ local function router_api_call_unsafe(func, router, ...)
     return func(router, ...)
 end
 
+local function router_cfg_fiber_safe(router, cfg, is_reload)
+    if router.is_cfg_in_progress then
+        error(lerror.vshard(lerror.code.ROUTER_CFG_IS_IN_PROGRESS, router.name))
+    end
+
+    router.api_call_cache = router_api_call_unsafe
```
Gerold103, line 1703: Wouldn't it work to set the api call cache in the same place where you set `is_enabled` on reload now?

Gerold103 (commented):
Thanks for the fixes! Very good, only nits left!

test/router-luatest/reload_test.lua
Gerold103: This place hangs on my PC right now. Apparently, I have more than 1024 bytes written after the discovery was started. Would be good to redo it on service info.

test/luatest_helpers/vtest.lua
Serpentian: Fixed

test/router-luatest/reload_test.lua
Serpentian: We can place it right into `reload_router`, if you don't mind. We can just compare tables of services, which is cheap.

vshard/router/init.lua
Serpentian: No problem, we can place it there

test/router-luatest/reload_test.lua
Serpentian: Because the test uses it, so old version must have service_info. I extended comment

test/router-luatest/reload_test.lua
Serpentian: Fixed

test/router-luatest/reload_test.lua
Serpentian: It would

test/lua_libs/util.lua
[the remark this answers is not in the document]
Gerold103: \> 40 seconds on mine. What is the reason for that? Did you measure what took the longest time? Can it be sped up? This is awfully slow tbh.

test/reload_evolution-luatest/reload_router_test.lua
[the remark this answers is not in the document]
Gerold103: Are you sure? I don't see it being used in the patch.

test/lua_libs/util.lua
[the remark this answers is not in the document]
Serpentian: Here're the results of the time measurements:

`everything`: 41.93
`test_basic`: 11.05 sec
`test_discovery`: 11.09
`test_master_search`: 21.14

----

1. I accidentally waited for reloading of `discovery_service` in `test_basic`. Removing it makes the time of test 31.68.
2. Decreased delay in `helpers.retrying`: 31.49
3. Removing additional checks after restoring cluster in master search gives us 21.50

Looks like that's all we can do about it

test/reload_evolution-luatest/reload_router_test.lua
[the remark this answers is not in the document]
Serpentian: Temporary directories are used now. They're unique enough to allow running test in parallel. See `before_all`:
```lua
    g.vshard_copy_path = fio.tempdir()
    -- Copy source to the temporary directory
    t.assert_equals(fio.mkdir(g.vshard_copy_path .. '/.git'), true)
    t.assert_equals(fio.copytree(vtest.sourcedir .. '/.git',
        g.vshard_copy_path .. '/.git'), true)
```

However, new test-run already uses temporary directories by default, and creating new one just for that test don't make a lot of sense. So, let's allow user or test-run to decide, where to put our files and use `fio.tempdir` only as fallback option, if no `VARDIR` is provided.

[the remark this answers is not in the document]
Gerold103: In the new version you will call `fio.tempdir()` on each load of `vtest` when `VARDIR` is not specified. Even when `vtest.vardir` is not needed. Which looks like an overkill, because 1) the non-main instances won't have `VARDIR`, 2) `fio.tempdir()` actually creates a directory. So after running all luatests you will have N temp directories created where N - number of storages and routers started and restarted ever. Even though you needed it just once for a single test.

I would suggest to do that just in the new test file. Not in `vtest.lua`.

test/router-luatest/reload_test.lua
```diff
@@ -0,0 +1,271 @@
...
+    vtest.drop_instance(g, router)
+end
+
+local function test_master_search_template(g, router, auto_master_cfg)
+    vtest.router_cfg(router, auto_master_cfg)
+
+    -- Working with first replicaset (2 instances)
+    local rs_uuid, master_uuid = g.replica_1_a:exec(function()
+        local info = box.info()
+        return info.cluster.uuid, info.uuid
```
Gerold103, line 204: Now we have to use `util.replicaset_uuid()`. Lets make a rebase and test on the latest tarantool core.

Gerold103 (changes requested):
Thanks for the work! Just one new comment, related to 3.0.

test/router-luatest/reload_test.lua
Serpentian: As rebase was already done, I only fixed getting names. I propose to use `server:replicaset_uuid()` and `server:instance_uuid()` rather than `utill` in favor of code simplicity and readability. I know, that `box.info` is called twice in such situation, but it doesn't significantly affects the time test runs.

PR #406, opened 2023-03-28
test: fix flaky router_test

### LINKED ISSUE #403: flaky test: router-luatest/router_test.lua
```
router-luatest/router_test.lua                                  
TAP13 parse failed (Descending test id on line: 'not ok 11\trouter.test_router_service_info').

No result file (router-luatest/router.result) found.
Run the test with --update-result option to write the new result file.
[ fail ]
```

I see no other useful information in logs. [A link to the fail in CI](https://github.com/tarantool/tarantool/actions/runs/4472734416/jobs/7859396051).

Serpentian:
router_test.lua fails constantly on integration CI in tarantool with error 'Peer closed' after simulating unconfigured box. The reason for this problem is very unclear, but the following changes fix it.

Closes #403

NO_DOC=testfix

### DIFF
--- a/test/router-luatest/router_test.lua
+++ b/test/router-luatest/router_test.lua
@@ -469,24 +469,23 @@ g.test_enable_disable = function(g)
         _G.fiber_new:set_joinable(true)
     end, {global_cfg})
 
-    -- emulate unconfigured box
-    router:exec(function()
-        rawset(_G, 'old_box_cfg', box.cfg)
-        box.cfg = function(...) return _G.old_box_cfg(...) end
-    end)
-
     local err1, err2 = router:exec(function()
+        -- emulate unconfigured box
+        local old_box_cfg = box.cfg
+        box.cfg = function(...) return old_box_cfg(...) end
+
         rawset(_G, 'static_router', ivshard.router.internal.routers._static_router)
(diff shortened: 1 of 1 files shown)


test/router-luatest/router_test.lua
```diff
@@ -469,24 +469,23 @@ g.test_enable_disable = function(g)
         _G.fiber_new:set_joinable(true)
     end, {global_cfg})
 
-    -- emulate unconfigured box
-    router:exec(function()
-        rawset(_G, 'old_box_cfg', box.cfg)
+    local err1, err2 = router:exec(function()
+        -- emulate unconfigured box
+        local old_box_cfg = box.cfg
         box.cfg = function(...) return _G.old_box_cfg(...) end
```
Gerold103, line 475: You have to change this line to use `old_box_cfg`, not `_G.old_box_cfg`.

Gerold103 (changes requested):
Thanks for the patch! I couldn't find the reason either. I suspect that luatest somewhere might be trying to do `type(box.cfg)` between execs, or tries to do `box.cfg.<key>`, they fail, and it terminates the connection.

But I couldn't locate any places which could behave like that.

test/router-luatest/router_test.lua
Serpentian: My bad( Fixed

PR #407, opened 2023-03-28
Support Tarantool 3.0

Gerold103:
It is not released yet and not even pushed to master branch. But vshard has to be ready in advance to make the integration tests pass. For trying 3.0 yourself use https://github.com/tarantool/tarantool/pull/8289.

### DIFF
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -370,6 +370,13 @@ local feature = {
     end)(),
     error_stack = version_is_at_least(2, 4, 0, nil, 0, 0),
     csw = fiber.self().csw ~= nil,
+    info_replicaset = (function()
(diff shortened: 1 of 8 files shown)


vshard/util.lua
```diff
@@ -397,6 +404,19 @@ do
...
 end
 
+local replicaset_uuid
+if feature.info_replicaset then
+    replicaset_uuid = function(info)
+        info = info ~= nil and info or box.info
+        return info.replicaset.uuid
+    end
+else
+    replicaset_uuid= function(info)
```
olegrok, line 414: space before `=` is missed

olegrok (approved):
Thanks for your patch. LGTM with one nitpick.

PR #417, opened 2023-05-17
test: fix flaky rebalancer/rebalancer.test.lua

### LINKED ISSUE #360: flaky test: rebalancer/rebalancer.test.lua
It hangs on both storage engines. Here're logs:
[linux_2.7_test-force.txt](https://github.com/tarantool/vshard/files/9180176/linux_2.7_test-force.txt)

Also failed on version 2.9

Serpentian:
Currently the test hangs sometimes on the line, which waits for the message "Rebalance routes are sent" to appear in logs.

According to the logs this message appears, when hang happens, but it is located before the bunch of "aa..a", which means, that rebalancer managed to send routes before waiting actually started.

This happens due to the fact, that rebalancer is waked up before waiting. Let's not do that.

Closes #360

NO_DOC=testfix

### DIFF
--- a/test/rebalancer/rebalancer.result
+++ b/test/rebalancer/rebalancer.result
@@ -149,9 +149,6 @@ test_run:switch('box_1_a')
 vshard.storage.rebalancer_enable()
 ---
 ...
-vshard.storage.rebalancer_wakeup()
----
-...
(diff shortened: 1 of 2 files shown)


Gerold103 (approved):
Thanks for the patch!

Issue #389, opened 2022-12-05
flaky test: router/router


Serpentian:
```text
Test failed! Result content mismatch:
--- router/router.result	Mon Dec  5 10:16:19 2022
+++ /home/runner/work/vshard/vshard/test/var/rejects/router/router.reject	Mon Dec  5 10:19:41 2022
@@ -415,9 +415,9 @@
 ---
 - - status: unknown
   - uuid: <replicaset_2>
-    status: available_ro
+    status: unreachable
   - uuid: <replicaset_2>
-    status: available_ro
+    status: unreachable
 ...
 _ = test_run:cmd('start server storage_2_a')
 ---
```

Reproduced on Tarantool 2.7

Serpentian:
I cannot reproduce it locally in any way. But I got the following output, while making read-only request for the second bucket:

```bash
[001] router/router.test.lua                                          [ fail ]
[001]
[001] Test failed! Result content mismatch:
[001] --- router/router.result	Fri May 19 02:44:31 2023
[001] +++ var/rejects/router/router.reject	Tue May 23 13:46:06 2023
[001] @@ -401,23 +401,23 @@
[001]  _ = test_run:cmd('stop server storage_2_a')
[001]  ---
[001]  ...
[001] -util.check_error(vshard.router.call, 1, 'read', 'echo', {123})
[001] ----
[001] -- null
[001] -- bucket_id: 1
[001] -  code: 8
[001] -  unreachable_uuid: <replicaset_2>
[001] -  name: UNREACHABLE_REPLICASET
[001] -  message: There is no active replicas in replicaset <replicaset_2>
[001] -  type: ShardingError
[001] +util.check_error(vshard.router.call, 2, 'read', 'echo', {123})
[001] +---
[001] +- 123
[001] +- null
[001]  ...
[001]  vshard.router.buckets_info(0, 3)
[001]  ---
[001]  - - status: unknown
[001]    - uuid: <replicaset_2>
[001] -    status: available_ro
[001] +    status: unreachable
[001]    - uuid: <replicaset_2>
[001] -    status: available_ro
[001] +    status: unreachable
[001] +...
[001] +util.check_error(vshard.router.call, 2, 'read', 'echo', {123})
[001] +---
[001] +- 123
[001] +- null
[001]  ...
[001]  _ = test_run:cmd('start server storage_2_a')
```

[...]

PR #421, opened 2023-05-23
test: fix flaky router/router.test.lua

Serpentian:
Currently the test may fail on retrieving info about buckets after
stopping the master. The router.call is made before we check statuses
of these buckets.

Deleting this call always results in `unreachable` state for buckets,
which means that if this call is done quickly we get test flakiness, as
failover sometimes doesn't have time to update nearest available replica
object.

Let's explicitly wake up failover and wait, until replica object is
updated.

Closes https://github.com/tarantool/vshard/issues/389

NO_DOC=testfix

### DIFF
--- a/test/router/router.result
+++ b/test/router/router.result
@@ -401,6 +401,14 @@ vshard.router.buckets_info(0, 3)
 _ = test_run:cmd('stop server storage_2_a')
 ---
 ...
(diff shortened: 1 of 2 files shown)


Gerold103 (approved):
Thanks for the fix!

PR #424, opened 2023-06-27
test: use dofile() for configs instead of require

Totktonada:
The main reason of this change is that we're going to introduce a new built-in module `config`. See https://github.com/tarantool/tarantool/issues/8724 for details.

Introducing a new module in the global namespace is legal according to our compatibility rules. Quoted from https://github.com/orgs/tarantool/discussions/6182:

> Adding a new built-in module or a new global value is considered as
> the compatible change.

Moreover, in the real-world applications configs may be changed and reloaded, so they likely will be loaded using dofile() or loadfile() without require()'s caching.


Gerold103 (approved):
Thanks for the patch!

Issue #436, opened 2023-09-06
map_callrw can't find c stored procedure


vol0n:
### Problem
We have a c stored procedure `execute`, earlier we called it via replicaset api like this:
```
replica:callbre(
                'execute',
                args,
                { is_async = true }
)
```
Now we've started to use `map_callrw` router api, which gives us an error saying it is not defined:
```
router.map_callrw('execute', args)

---
- null
- code: 33
  base_type: ClientError
  type: ClientError
  message: Procedure 'execute' is not defined
  trace: [{'file': './src/box/lua/call.c', 'line': 112}]
- 5bb12a3c-4de8-4059-8172-8c926d6a6bc2
...
```

The function was created like this:
```
box.schema.func.create(
            'execute',
            { if_not_exists = true, language = 'C' }
        )
```
And it is callable through box.func api:
```
box.func['execute']:call()
---
- error: 'Failed to decode dispatched data: Invalid(Tuple, Some("expected
    tuple of 2 elements, got 0"))'
...
```

Tarantool version: 2.11.0
Vshard version: latest

### Hint
After digging a bit in the code I found that `map_callrw(f, args)` calls `storage_map(f, args)`, the latter calls `net.box.self.call(f, args)` and this internally in tarantool calls `box_lua_find` which tries to find the function in lua globals.

On the other hand, `replica:callbre(f, args)` calls `net.box.conn.call`, the latter first [checks](https://github.com/tarantool/tarantool/blob/master/src/box/call.c#L195) whether the function is in cache and only then uses `box_lua_find`.

So the problem seems to be in using `self.call` inside `storage_map`

Gerold103:
Can we just always use `box.func`? `netbox.self` is used because in the old times it was the only way to call a local function without requiring its Lua or C module.

Serpentian:
Sure, we can. But it doesn't change the fact, that `netbox.self` doesn't work.

Gerold103:
It does mean that the ticket is not blocked though. I suggest we drop the 'blocked' label.

PR #437, opened 2023-09-18
storage: fix local_call not finding C stored procedures

Serpentian:
Currently, if function was created as C stored procedure or as Lua
persistent function (with body argument) via box.schema.func.create,
all types of router.call and router.map_callrw cannot find it and
return `Procedure 'name' is not defined` error.

This is cased by the fact that both of these function use local_call,
which invokes net_box.self.call. It didn't work with these type of
functions before Tarantool 3.0.0-beta1-18. Let's use box.func, where
it's needed instead of net_box.self.call in local_call.

Closes https://github.com/tarantool/vshard/issues/436

NO_DOC=bugfix

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -265,8 +265,27 @@ end
 -- The function returns pcall() as is, because is used from places where
 -- exceptions are not allowed.
 --
-local function local_call(func_name, args)
+local local_call
+
+if util.version_is_at_least(3, 0, 0, 'beta', 1, 18) then
+
+local_call = function(func_name, args)
     return pcall(netbox_self_call, netbox_self, func_name, args)
+end
+
+else -- < 3.0.0-beta1-18
+
+-- net_box.self.call() doesn't work with C stored and Lua persistent
+-- functions before 3.0.0-beta1-18, so we try to call it via func.call
+-- API prior to using net_box.self API.
+local_call = function(func_name, args)
+    local func = box.func and box.func[func_name]
+    if not func then
+        return pcall(netbox_self_call, netbox_self, func_name, args)
+    end
+    return pcall(func.call, func, args)
+end
+
 end
 
 local function master_call(replicaset, func, args, opts)
--- a/test/storage-luatest/storage_1_test.lua
+++ b/test/storage-luatest/storage_1_test.lua
@@ -168,3 +168,17 @@ test_group.test_recovery_bucket_stat = function(g)
 
     vtest.cluster_cfg(g, global_cfg)
 end
+
+test_group.test_local_call = function(g)
+    -- box.func was introduced in 2.2.1.
+    t.run_only_if(vutil.version_is_at_least(2, 2, 1, nil, 0, 0))
+    g.replica_1_a:exec(function()
+        local body = [[function(a, b) return a + b end]]
+        box.schema.func.create('sum', {body = body})
+        local bid = _G.get_first_bucket()
+        local ok, ret = ivshard.storage.call(bid, 'read', 'sum', {1, 2})
+        ilt.assert_equals(ret, 3)
+        ilt.assert_equals(ok, true)
+        box.func.sum:drop()
+    end)
+end


vshard/storage/init.lua
```diff
@@ -251,7 +251,17 @@ end
 -- exceptions are not allowed.
```
Gerold103, line 251: > NO_TEST=<already tested>

But where?

vshard/storage/init.lua
```diff
@@ -251,7 +251,17 @@ end
...
 --
 local function local_call(func_name, args)
-    return pcall(netbox_self_call, netbox_self, func_name, args)
+    -- If function was created with box.schema.func we call it via
+    -- func.call API, as net_box.self.call doesn't work with C stored
+    -- procedures up to Tarantool 3.0. If box.func is inaccessible,
+    -- which is true for all versions below 2.2, we fallback to
+    -- net_box.self.call too.
```
Gerold103, line 258: If something is working one way for all versions < N and in another way for all versions > N, then this doesn't have to be a runtime check. Please, make `local_call()` depend on the core version. For earlier versions we can try the fallback approach. For newer ones - simply always use `netbox_self_call()` like before.

As "make it depend on core version" I mean that you can define `local_call()` differently depending on the version. Similar to how `util.index_min()` is defined.

Gerold103 (commented):
Thanks for the patch!

vshard/storage/init.lua
Serpentian: It was in commit msg, but github suppresses everything inside <...> in PR's comment. However, this is not important now, as I added the test, as it turns out this problem also happens with functions, created with `box.schema.func.create(..., {body = " ... "})`

vshard/storage/init.lua
Serpentian: > If something is working one way for all versions < N and in another way for all versions > N, then this doesn't have to be a runtime check.

We've already done version dependent thing via runtime checks: [netbox_is_conn_dead](https://github.com/tarantool/vshard/blob/9fda3546ded3f74fa03dda371a984986e2b367d1/vshard/replicaset.lua#L165C1-L185).

We cannot use `netbox_self_call` for newer versions, as there's no version, where it works now. I don't know, when I'll do it ([ticket](https://github.com/tarantool/tarantool/issues/9131)). So, let's use runtime check for now.

Serpentian: The only version dependent code we can make now: before 2.2.0 always use `netbox_self_call`, as there're no `box.func` API, in all other cases try to fallback. I don't really see much sense in such approach, but this can be done, if you insist

Gerold103: > netbox_is_conn_dead

Here it is justified, because it has the fast-path which is version-agnostic and works almost always:
```Lua
    if conn.state == 'active' then
        return false
    end
```
In your case you've added branching on the path which is executed always, so this is different.

> We cannot use netbox_self_call for newer versions, as there's no version, where it works now

If this is not fixed even in master, then we can't merge nor test it until it is done in master. If this is already done in master, then you can always check even a not yet released version by the commit number. For example, the latest master I see is `3.0.0-beta1-5-ge844cd09b`, which means 5 commits after `3.0.0-beta1`. `vshard.version` module supports commit numbers, so you can do "if version >= 3.0.0-beta1-5 then use the new way, otherwise use the fallback approach".

Serpentian: No, it's not done in master, the bug is still there (it's not related to release). Nonetheless, I don't get, why we're blocking the PR, if we can fix it in vshard right now, it's even tested.

So, at the end of a day we mark https://github.com/tarantool/vshard/issues/436 as blocked by https://github.com/tarantool/tarantool/issues/9131?

vshard/storage/init.lua
```diff
@@ -266,7 +266,17 @@ end
...
+    -- If function was created with box.schema.func we call it via
+    -- func.call API, as net_box.self.call doesn't work with C stored
+    -- procedures up to Tarantool 3.X. If box.func is inaccessible,
+    -- which is true for all versions below 2.2, we fallback to
+    -- net_box.self.call too.
+    if box.func == nil or box.func[func_name] == nil then
+        return pcall(netbox_self_call, netbox_self, func_name, args)
+    end
+
+    local func = box.func[func_name]
```
Gerold103, line 278: Lets at least avoid the double-lookup in `box.func` and do it like this:
```Lua
local func = box.func and box.func[func_name]
if not func then
    return pcall(netbox_self_call, netbox_self, func_name, args)
end
return pcall(func.call, func, args)
```

vshard/storage/init.lua
Gerold103: Ok, I see the full picture now, thanks for the explanation. Lets finalize this PR and file a ticket to make the `local_call` depend on the version in the future.

Serpentian: https://github.com/tarantool/tarantool/issues/9131 was fixed by the Core team. Let's make version banching

Gerold103 (approved):
Thanks for the fixes!

PR #443, opened 2023-11-09
Config option to disable user management

### LINKED ISSUE #435: Allow full control of storage user outside of vshard
Currently, vshard grants the storage user all the necessary privileges. However, this is not always necessary - in some cases, the user is controlled outside vshard, and the granted privileges can override this control. This is essential for Tarantool 3.0 startup flow to have an option that could determine if vshard should grant privileges.

Gerold103:
The patch introduces a new option `schema_management_mode` which allows to turn off automatic access management (users, grants, function exports). It also adds helper functions for exports deploy.

Closes #435.

### DIFF
--- a/vshard/storage/schema.lua
+++ b/vshard/storage/schema.lua
@@ -1,0 +1,334 @@
+local llog = require('log')
+local lutil = require('vshard.util')
+
+local MODULE_INTERNALS = '__module_vshard_storage_schema'
+-- Update when change behaviour of anything in the file, to be able to reload.
+local MODULE_VERSION = 1
+
+local M = rawget(_G, MODULE_INTERNALS)
+if not M then
+    M = {
+        ----------------------- Common module attributes -----------------------
+        module_version = MODULE_VERSION,
+        -- Flag whether Tarantool core version was already checked and the
+        -- schema is up to date with this version.
+        is_core_up_to_date = false,
+        -- Index which is a trigger to shard its space by numbers in this index.
+        -- It must have at first part either unsigned, or integer or number type
+        -- and be not nullable. Values in this part are considered as bucket
+        -- identifiers.
+        shard_index = nil,
+        errinj = {
+            ERRINJ_UPGRADE = false,
+        },
+    }
+else
+    return M
+end
+
+local schema_version_mt = {
+    __tostring = function(self)
+        return string.format('{%s}', table.concat(self, '.'))
+    end,
+    __serialize = function(self)
+        return tostring(self)
+    end,
+    __eq = function(l, r)
+        return l[1] == r[1] and l[2] == r[2] and l[3] == r[3] and l[4] == r[4]
+    end,
+    __lt = function(l, r)
+        for i = 1, 4 do
+            local diff = l[i] - r[i]
+            if diff < 0 then
+                return true
+            elseif diff > 0 then
+                return false
+            end
+        end
+        return false
+    end,
+}
+
+local function schema_version_make(ver)
+    return setmetatable(ver, schema_version_mt)
+end
+
+--
+-- VShard versioning works in 4 numbers: major, minor, patch, and a last helper
+-- number incremented on every schema change, if first 3 numbers stay not
+-- changed. That happens when users take the latest master version not having a
+-- tag yet. They couldn't upgrade if not the 4th number changed inside one tag.
+
+-- The schema first time appeared with 0.1.16. So this function describes schema
+-- before that - 0.1.15.
+--
+local function schema_init_0_1_15_0(username, password)
+    llog.info("Initializing schema %s", schema_version_make({0, 1, 15, 0}))
+    box.schema.user.create(username, {
+        password = password,
+        if_not_exists = true,
+    })
+    box.schema.user.grant(username, 'replication', nil, nil,
+                          {if_not_exists = true})
+
+    local bucket = box.schema.space.create('_bucket')
+    bucket:format({
+        {'id', 'unsigned'},
+        {'status', 'string'},
+        {'destination', 'string', is_nullable = true}
+    })
+    bucket:create_index('pk', {parts = {'id'}})
+    bucket:create_index('status', {parts = {'status'}, unique = false})
+
+    local storage_api = {
+        'vshard.storage.sync',
+        'vshard.storage.call',
+        'vshard.storage.bucket_force_create',
+        'vshard.storage.bucket_force_drop',
+        'vshard.storage.bucket_collect',
+        'vshard.storage.bucket_send',
+        'vshard.storage.bucket_recv',
+        'vshard.storage.bucket_stat',
+        'vshard.storage.buckets_count',
+        'vshard.storage.buckets_info',
+        'vshard.storage.buckets_discovery',
+        'vshard.storage.rebalancer_request_state',
+        'vshard.storage.rebalancer_apply_routes',
+    }
+    for _, name in ipairs(storage_api) do
+        box.schema.func.create(name, {setuid = true})
+        box.schema.user.grant(username, 'execute', 'function', name)
+    end
+    box.space._schema:replace({'vshard_version', 0, 1, 15, 0})
+end
+
+local function schema_upgrade_to_0_1_16_0(username)
+    -- Since 0.1.16.0 the old versioning by 'oncevshard:storage:<number>' is
+    -- dropped because it is not really extendible nor understandable.
+    llog.info("Insert 'vshard_version' into _schema")
+    box.space._schema:replace({'vshard_version', 0, 1, 16, 0})
+    box.space._schema:delete({'oncevshard:storage:1'})
+
+    -- vshard.storage._call() is supposed to replace some internal functions
+    -- exposed in _func; to allow introduction of new functions on replicas; to
+    -- allow change of internal functions without touching the schema.
+    local func = 'vshard.storage._call'
+    llog.info('Create function %s()', func)
+    box.schema.func.create(func, {setuid = true})
+    box.schema.user.grant(username, 'execute', 'function', func)
+    -- Don't drop old functions in the same version. Removal can happen only
+    -- after 0.1.16. Or there should appear support of rebalancing from too old
+    -- versions. Drop of these functions now would immediately make it
+    -- impossible to rebalance from old instances.
+end
+
+local function schema_downgrade_from_0_1_16_0()
+    llog.info("Remove 'vshard_version' from _schema")
+    box.space._schema:replace({'oncevshard:storage:1'})
+    box.space._schema:delete({'vshard_version'})
+
+    local func = 'vshard.storage._call'
+    llog.info('Remove function %s()', func)
+    box.schema.func.drop(func, {if_exists = true})
+end
+
+--
+-- Make vshard.storage.bucket_recv() understand raw args as msgpack object. It
+-- doesn't necessarily make it faster, but is essential to preserve the original
+-- tuples as is, without their contortion through Lua tables. It would break
+-- field types like MP_BIN (varbinary).
+--
+local function schema_upgrade_bucket_recv_raw()
+    if not lutil.feature.msgpack_object then
+        return
+    end
+    if box.func == nil then
+        return
+    end
+    local name = 'vshard.storage.bucket_recv'
+    local func = box.func[name]
+    if func == nil then
+        return
+    end
+    if func.takes_raw_args then
+        return
+    end
+    llog.info("Upgrade %s to accept args as msgpack", name)
+    box.begin()
+    local ok, err = pcall(function()
+        -- Preserve all the possible grants which could be given to the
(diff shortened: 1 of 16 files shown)


Gerold103:
@ImeevMA can you please have a look and check if this is enough for `config` module? You can now disable the user management. But I still do `func.create()` for various internal functions, since they don't seem to be related to users. Let me know if those also shouldn't be created (but then you would need to grant `execute` `universe` rights to your manual user or create those functions yourself in the same way - both sound like bad ideas).

ImeevMA:
@Gerold103 Regarding function creation and granting of privileges, the plan is to [introduce](https://www.notion.so/tarantool/_fun-dead1daee80540b4ac913435650b6eb5) granting of privileges for Lua functions not from `_func`. We then plan to introduce a new default role, `sharding`, which will have `execute` access to all necessary Lua vshard functions. So, in order for the user to perform these functions, we just need to give him this role. I believe this would solve all the problems with creating functions and granting privileges. How do you think?

The only problem is that we don't know when exactly this feature will arrive. Anyway, it looks like we can disable function creation and privilege granting in vshard. If the feature comes too late, we'll come up with something on our end.

@Totktonada, do you have anything to add?

Totktonada:
I think that it would be good to control whether to add `_func` entries. This ability doesn't look critical for our tasks, but it would be more clean to don't add the entries if they're not needed.

More on it: how about exposing a function list that are essential to expose into iproto to make rebalancer and router working? If we have a declarative information about vshard functions in the `config` logic, we can better support different vshard versions.

Totktonada:
NB: Current version of the in-progress [RFC](https://www.notion.so/tarantool/Execute-privilege-for-unregistered-Lua-functions-3191e0cf5c314349b6fff2eb6891b012) regarding granular `lua_call` privilege discards access to a function according to the `lua_call` privilege if it is present in `_func`. It means that we need to control whether `box.schema.func.create()` will be called.

Totktonada:
NB 2: @locker agreed to allow `lua_call` if a `_func` entry is present (in some cases). It is tracked by https://github.com/tarantool/tarantool/issues/9363.

Anyway, I think that control, whether to create `_func` entries in vshard, is a good feature.

Gerold103:
@Totktonada will this Lua grants feature allow to grant access to not yet created functions? It is not needed right now, but in future it might become necessary, if vshard storages would want to communicate with each other during first `vshard.storage.cfg`. Then I would need the grants to be present before the first cfg, thus before those functions are properly created. (Before the first cfg there are stubs, but they only throw an error.)

locker:
> will this Lua grants feature allow to grant access to not yet created functions?

Yes, they will (see the [design document](https://www.notion.so/tarantool/Execute-privilege-for-unregistered-Lua-functions-3191e0cf5c314349b6fff2eb6891b012?pvs=4#5905157df65a48328ad1ce494dcc8d53)).

Totktonada:
However, they're still stored in `_priv`, so not accessible before first `box.cfg()`.

Gerold103:
@Totktonada and @ImeevMA , please, check the new version out.
- You can disable user, functions, and grants management via `schema_management_mode = 'manual_access' config option;
- You can deploy those things yourself using the new module `vshard.storage.exports`. The field `exports.log` contains the history of versions. Once you've found the one you want, you can compile them for the current core version and then deploy functions and/or grants:
```Lua
local lvexports = require('vshard.storage.exports')
local exports = find_exports_for_vshard_version(lvexports.log, vshard_version_str)
exports = lvexports.compile(exports)
lvexports.deploy_funcs(exports)
lvexports.deploy_privs(exports, username_to_grant_privs_to)
```

Totktonada:
@Gerold103 It seems that it is conceptually what we need.

vshard/storage/schema.lua
```diff
@@ -256,7 +181,14 @@ local function schema_upgrade(target_version, username, password)
             if errinj == 'begin' then
                 ok, err1 = false, 'Errinj in begin'
             else
-                ok, err1 = pcall(handler.upgrade, username)
+                ok, err1 = pcall(function()
+                    local exports = schema_find_exports_for_vshard_version(
+                        next_version)
+                    exports = lvexports.compile(exports)
+                    handler.upgrade()
```
Serpentian, line 193: Why are we trying to update `_schema` several lines below, when we have already done that in `handler.upgrade()`?

```lua
            ok, err1 = pcall(_schema.replace, _schema,
                             {'vshard_version', unpack(next_version)})
            if not ok then
                llog.info("Upgraded schema to %s but couldn't update _schema "..
                          "'vshard_version' - fatal error: '%s'", next_version,
                          err1)
                os.exit(-1)
            end
```

vshard/storage/schema.lua
```diff
@@ -186,8 +191,10 @@ local function schema_upgrade(target_version, username, password)
                         next_version)
                     exports = lvexports.compile(exports)
                     handler.upgrade()
-                    lvexports.deploy_funcs(exports)
-                    lvexports.deploy_privs(exports, username)
+                    if M.management_mode == 'auto' then
```
Serpentian, line 194: The whole `pcall` should be inside if statement. Otherwise, you set `vshard_version` to `0.1.16` without any changes.

But it seems, we should exit earlier, right after `schema_init_0_1_15_0`

Serpentian (commented):
Thank you for the patch! Looks great! Some minor comments

vshard/storage/schema.lua
```diff
@@ -80,140 +94,51 @@ local function schema_init_0_1_15_0(username, password)
...
-    for _, name in ipairs(storage_api) do
-        box.schema.func.create(name, {setuid = true})
-        box.schema.user.grant(username, 'execute', 'function', name)
-    end
+    local exports = lvexports.log[1]
+    assert(exports.version == '0.1.15.0')
+    exports = lvexports.compile(exports)
+    lvexports.deploy_funcs(exports)
+    lvexports.deploy_privs(exports, username)
     box.space._schema:replace({'vshard_version', 0, 1, 15, 0})
```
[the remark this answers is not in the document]
Gerold103: > Why don't we set _schema.vshard_version automatically?

What do you mean automatically? Where is it automatic?

> Currently if we manually updated to 0.1.16 and then we try to use automatic update, we'll try to deploy everything one more time.

The "users" (Alexander and Mergen - nobody else should use this, those exports are bad) should bump the version if they want to fully deploy the entire schema themselves. If they don't do that, then it will be done again inside vshard, yes (unless schema management mode is not `manual_access`). The patch had no intention to give the "users" a full featured upgrade module for deploying the whole schema. VShard still does things on its own, like `_bucket` space, for example. It can be probably extended in the future for deploying the entire schema, but that would be ridiculous IMO. Even the current requirement to make the exports manual is a sign of something being wrong in `config` module. I can understand the need to manage users on your own, but not the vshard private and public API. I will be against exporting the entire schema management as a public module if it ever comes to that.

> Maybe we can add exports.upgrade_callback() to the returned value of exports.compile? So, that user can manually call it, when upgrade is done.

Why would the "users" need a callback for that? Can they just do whatever they want after `deploy_funcs/privs()` returns?

vshard/storage/schema.lua
Gerold103: > Why are we trying to update _schema several lines below, when we have already done that in handler.upgrade()?

In `schema_init_0_1_15_0()` it is set because this is an init-function, it is expected to create the complete schema, including the version.

In `schema_upgrade_to_0_1_16_0()` it is set, because we switch from the legacy version storage `'oncevshard:storage:1'` to `'vshard_version'`.

It might look excess, but if we would have, say, `schema_upgrade_to_0_1_25_0()`, then it wouldn't touch the version, because nothing about the version-storage would change. It is just not very clear right now because we have only 2 versions and they both need to touch the version as a part of their upgrade routine.

I do agree though, that we could drop the version setting from `schema_upgrade_to_0_1_16_0()` and probably nothing would change. I was thinking about it when touched this code and decided not to touch it, after all 😁.

vshard/storage/schema.lua
Gerold103: > Otherwise, you set vshard_version to 0.1.16 without any changes.

Yes. This is intentional (except that there will be a change for some really old vshards - removal of `'oncevshard:storage:1'`). The reason is that when you voluntary choose to manage the access (!= schema) on your own, then you do that outside of the internal version upgrades. Otherwise I would be unable to do any upgrades at all, including potential internal space changes for `_bucket` or creation of some new spaces, when the "users" just want to manage the exports on their own.

> But it seems, we should exit earlier, right after schema_init_0_1_15_0

No, there is no schema management mode called `manual`. There is only `manual_access`. The "users" only have control over the exports. The backbone of the upgrade procedure is owned by vshard. If fully manual schema is needed somewhere, then that place is behaving really wrong. This is like saying that you want to create all system spaces and WAL before Tarantool is started first time. To those users the current exports won't help, and they shouldn't help.

I was thinking though if that would be possible at all to create generic schema version log which could contain everything we have now and what we would potentially need. But it appears it can't be done as easily as the exports. Such a schema log would need to be able to describe untrivial things:
- Deletion and insertion of random keys in random spaces (to drop `'oncevshard:storage:1'` from `_schema`, to update `vshard_version`);
- Creation of spaces and indexes (to create `_bucket` and its indexes);
- Alter of those spaces and indexes (I want to change `_bucket` status field from string to unsigned), with changing their data first to prepare for the new space format (don't even know how to describe it in the form of Lua tables for space whose content is unknown at the time of upgade);

The last point means that we can't even do the schema log in the form of a list of xlog-like rows, because for `_bucket` alter we don't know the number of rows nor their exact keys (stuff changes during rebalancing). It would mean export either in the form of Lua-code wrapped into string, or SQL-code (wouldn't work on 1.10 which is still supported). That solution would make this schema log unusable for anything except just applying it right away (otherwise go parse Lua and SQL to extract details), which leads to the point why not just let vshard do the same and skip this schema log.

So my resolution was that I simply don't want to export that, because it would be too hard to use, and I see no reason why it should be used and maintained. In my tests in vtest somehow even before any function-export logs I was able to start and configure clusters which is supposedly what `config` also wants to do (I only do understand the manual user management, not exports management).

Nonetheless the patch exists because exporting just the functions and users and privs is easy enough not to bother with arguing about them.

vshard/storage/schema.lua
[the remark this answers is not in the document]
Serpentian: > VShard still does things on its own, like _bucket space, for example.

And that's pretty much it, nothing is done by vshard, when schema management mode is `manual`, except the space creation and writing `vshard_version` to `_schema`.

> Why would the "users" need a callback for that? Can they just do whatever they want after deploy_funcs/privs() returns?

I don't think we should export space creation, but `exports.upgrade_callback` could check, that the whole version was deployed (on 0.1.15 that space is also exists, on all versions - that functions were created) and write `vshard_version`.

> Even the current requirement to make the exports manual is a sign of something being wrong in config module.

Agree, I suppose, we should've stopped on the previous version, when only user wasn't created. IMO, granting privs should be done by vshard.

-----

In the end, I'm okay with the current version too

vshard/storage/schema.lua
```diff
@@ -267,7 +199,14 @@ local function schema_upgrade(target_version, username, password)
...
                 llog.info("Couldn't upgrade schema to %s: '%s'. Revert to %s",
                           next_version, err1, prev_version)
-                ok, err2 = pcall(handler.downgrade)
+                ok, err2 = pcall(function()
+                    local exports = schema_find_exports_for_vshard_version(
+                        prev_version)
+                    exports = lvexports.compile(exports)
+                    handler.downgrade()
+                    lvexports.deploy_funcs(exports)
+                    lvexports.deploy_privs(exports, username)
```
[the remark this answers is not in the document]
Serpentian: > The grants are deleted prior to the function deletion automatically by the core.

My bad, indeed they are

vshard/storage/schema.lua
Serpentian: > It might look excess, but if we would have, say, schema_upgrade_to_0_1_25_0(), then it wouldn't touch the version, because nothing about the version-storage would change. It is just not very clear right now because we have only 2 versions and they both need to touch the version as a part of their upgrade routine.

I see now, thank you for the explanation

Serpentian (approved):
Thank you for the patch! Looks great

vshard/storage/schema.lua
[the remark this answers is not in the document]
Gerold103: > when schema management mode is manual

We only have `auto` and `manual_access`, no full `manual`.

> I don't think we should export space creation, but exports.upgrade_callback could check, that the whole version was deployed 

Nah, too complicated. The stuff they want to do is feasible with the current solution too. If they come with more specific requests, we can think of it.

> Agree, I suppose, we should've stopped on the previous version, when only user wasn't created.

Yes. But I don't want to argue anymore for now. There will be another fight if they will want more.

PR #445, opened 2023-11-15
Disable box.cfg

### LINKED ISSUE #353: Drop `box.cfg` from router
That would be a breaking change, but should be done at some point anyway. There is no reason for router to have box configured. It doesn't need storage nor listening on a port.

### LINKED ISSUE #428: Allow to not invoke box.cfg() when vshard.*.cfg() is called
Currently, every time vshard.storage.cfg() or vshard.router.cfg() is called, box.cfg() is also called. This allows us to start a cluster using vshard. However, this is not necessary when vshard is run after the cluster is set up.

Serpentian:
This patchset cherry picks enum configuration type from `disable user management` PR, introduces option to disable box.cfg on storage and completely drops calling `box.cfg` on routers.

There won't be any better time for breaking change of disabling box.cfg on routers than release of Tarantool 3.0.0

Closes #353
Closes #428

### DIFF
--- a/example/router.lua
+++ b/example/router.lua
@@ -14,11 +14,13 @@ replicasets = {'cbf06940-0790-498b-948d-042b62cf3d29',
 
 -- Call a configuration provider
 cfg = dofile('localcfg.lua')
+-- vshard.router.cfg doesn't support box.cfg options.
+cfg = require('vshard.cfg').extract_vshard(cfg)
 if arg[1] == 'discovery_disable' then
     cfg.discovery_mode = 'off'
 end
-cfg.listen = 3300
 
+box.cfg{listen = 3300}
 -- Start the database with sharding
 vshard = require('vshard')
 vshard.router.cfg(cfg)
--- a/test/failover/cluster_changes.result
+++ b/test/failover/cluster_changes.result
@@ -54,7 +54,7 @@ test_run:switch('router_1')
 -- First test case: reverse weights, when only replica exists,
 -- and no replica candidate.
 --
-vshard.router.cfg(cfg)
+util.box_router_cfg(cfg)
 ---
 ...
 wait_state('All replicas are ok')
@@ -67,7 +67,7 @@ vshard.router.info().alerts
 reverse_weights()
 ---
 ...
-vshard.router.cfg(cfg)
+util.box_router_cfg(cfg)
 ---
 ...
 vshard.router.info()
@@ -274,7 +274,7 @@ test_run:switch('router_1')
 remove_some_replicas()
 ---
 ...
-vshard.router.cfg(cfg)
+util.box_router_cfg(cfg)
 ---
 ...
 info = vshard.router.info()
@@ -370,7 +370,7 @@ test_run:switch('router_1')
 add_some_replicas()
 ---
 ...
-vshard.router.cfg(cfg)
+util.box_router_cfg(cfg)
 ---
 ...
 #vshard.router.info().alerts > 1
--- a/test/lua_libs/util.lua
+++ b/test/lua_libs/util.lua
@@ -215,6 +215,23 @@ local function map_bucket_protection(test_run, cluster, value)
         [[vshard.storage.internal.is_bucket_protected = ...]], value)
 end
 
+local function box_router_cfg(cfg, router)
+    local vcfg = require('vshard.cfg')
+    local vshard_cfg = vcfg.extract_vshard(cfg)
+    local box_cfg = vcfg.extract_box(cfg, {})
+    box.cfg(box_cfg)
+    if router == nil then
+        -- Static router configuration
+        return vshard.router.cfg(vshard_cfg)
+    end
+    if type(router) == 'table' then
+        -- Named router reconfiguration
+        return router:cfg(vshard_cfg)
+    end
+    -- Named router creation
+    return vshard.router.new(router, vshard_cfg)
+end
+
 return {
     check_error = check_error,
     shuffle_masters = shuffle_masters,
@@ -231,4 +248,5 @@ return {
     git_checkout = git_checkout,
     portable_error = portable_error,
     is_timeout_error = is_timeout_error,
+    box_router_cfg = box_router_cfg,
 }
--- a/test/storage-luatest/box_cfg_mode_test.lua
+++ b/test/storage-luatest/box_cfg_mode_test.lua
@@ -1,0 +1,70 @@
+local t = require('luatest')
+local vtest = require('test.luatest_helpers.vtest')
+
+local test_group = t.group()
+
+local cfg_template = {
+    sharding = {
+        {
+            replicas = {
+                replica_1_a = {master = true},
+            },
+        },
+        {
+            replicas = {
+                replica_2_a = {master = true},
+            },
+        },
+    },
+    bucket_count = 20,
+    box_cfg_mode = 'manual',
+}
+
+local global_cfg
+
+test_group.before_all(function(g)
+    global_cfg = vtest.config_new(cfg_template)
+    vtest.cluster_new(g, global_cfg)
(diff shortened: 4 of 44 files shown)


test/lua_libs/util.lua
```diff
@@ -215,6 +215,23 @@ local function map_bucket_protection(test_run, cluster, value)
         [[vshard.storage.internal.is_bucket_protected = ...]], value)
 end
 
+local function router_cfg(cfg, router)
+    local vcfg = require('vshard.cfg')
```
Serpentian, line 219: `reload_evolution/storage` fails if `vshard.cfg` is included not inside the function:

```
[009] reload_evolution/storage.test.lua
[009]
[009] [Instance "storage_2_a" returns with non-zero exit code: 1]
[009]
[009] Last 15 lines of Tarantool Log file [Instance "storage_2_a"][/home/serpentian/Programming/tnt/vshard/test/var/009_reload_evolution/storage_2_a.log]:
[009] Starting instance storage_2_a...
[009] started
[009] Start failed: ...rd/test/var/vshard_git_tree_copy/vshard/storage/init.lua:2045: attempt to call field 'split' (a nil value)
[009] [ fail ]

```

test/storage-luatest/box_cfg_mode_test.lua
```diff
@@ -0,0 +1,62 @@
...
+--
+-- vtest.cluster_new does exactly, what user should do,
+-- when using 'manual' box_cfg_mode: call box.cfg prior
+-- to executing vshard.storage.cfg. So, just some basic test.
+--
+test_group.test_storage_basic = function(g)
+    t.assert(g.replica_1_a:grep_log('Box configuration was skipped'))
+    g.replica_1_a:exec(function(uuid)
+        local bid = _G.get_first_bucket()
+        local ok, err = ivshard.storage.bucket_send(bid, uuid)
```
Gerold103, line 57: Please, send the bucket back to its original place in the end of the test case. We need to leave the cluster state the same as when the test case was started. Here it won't break anything right now, but it is easier to simply enforce this rule everywhere.

vshard/storage/init.lua
```diff
@@ -3883,7 +3883,13 @@ local function storage_cfg_xc(cfgctx)
         error('Error injection: cfg')
     end
 
-    if not cfgctx.is_reload then
+    if cfgctx.new_cfg.box_cfg_mode == 'manual' then
+        if type(box.cfg) == 'function' then
+            local msg = "Box must be configured, when box_cfg_mode is 'manual'"
+            error(lerror.vshard(lerror.code.STORAGE_IS_DISABLED, msg))
+        end
+        log.info("Box configuration was skipped")
```
Gerold103, line 3891: Lets add "due to the manual `box_cfg_mode`" just for verbosity. I also suggest to move this entire check as early as we can. Otherwise we create the new replicaset objects just to fail here without using them. Lets do this right after `log.info("Starting configuration/reconfiguration...`.

example/router.lua
```diff
@@ -14,11 +14,13 @@ replicasets = {'cbf06940-0790-498b-948d-042b62cf3d29',
 
 -- Call a configuration provider
 cfg = dofile('localcfg.lua')
+-- vshard.router.cfg doesn't support box.cfg options.
+cfg.replication_connect_quorum = nil
```
Gerold103, line 18: I suggest to use `vshard.cfg.extract_vshard()` for this. I will make sure the config only has vshard options.

test/failover/cluster_changes.result
```diff
@@ -54,7 +54,7 @@ test_run:switch('router_1')
 -- First test case: reverse weights, when only replica exists,
 -- and no replica candidate.
 --
-vshard.router.cfg(cfg)
+util.router_cfg(cfg)
```
Gerold103, line 57: Lets make it more explicit that this function also calls `box.cfg` by naming it `util.box_router_cfg`.

test/instances/router.lua
```diff
@@ -19,6 +19,9 @@ _G.vshard = {
 }
 _G.ivshard = _G.vshard
 
+local vshard = _G.ivshard
```
Gerold103, line 22: You should be able to use the global names right away, without assigning them to a local variable. Both `vshard` and `ivshard` are already global in this file.

test/instances/router.lua
```diff
@@ -28,4 +31,21 @@ end
...
+    end
+    if type(router) == 'table' then
+        -- Named router reconfiguration
+        return router:cfg(vshard_cfg)
+    end
+    -- Named router creation
+    return vshard.router.new(router, vshard_cfg)
+end
+
+_G.router_cfg = router_cfg
```
Gerold103, line 50: Have you checked if this is really needed in luatests? If so then why? Which of them are trying to give the router `box.cfg` options? I would suggest we try to leave this file intact and try to fix the luatests so as they would call `box.cfg` explicitly for whatever they need. If vtest is doing that, then we should fix vtest.

The reasoning is that I want this instance file to be doing as few hacks as possible. This would help to catch potential issues which could be hidden by our hacks otherwise.

vshard/router/init.lua
```diff
@@ -1258,32 +1256,28 @@ end
...
 
-local function router_cfg(router, cfg, is_reload)
+local function router_cfg(router, cfg)
     cfg = lcfg.check(cfg, router.current_cfg)
     local vshard_cfg = lcfg.extract_vshard(cfg)
     local box_cfg = lcfg.extract_box(cfg, {})
+    for option, _ in pairs(box_cfg) do
+        -- Router doesn't any box.cfg options.
+        error(string.format('Unrecognized cfg option: %s', option))
+    end
```
Gerold103, line 1266: You can do it simpler:
```Lua
if next(box_cfg) then
    ... your error stuff
end
```

vshard/router/init.lua
```diff
@@ -1735,7 +1715,9 @@ end
 
 local router_mt = {
     __index = {
-        cfg = function(router, cfg) return router_cfg_fiber_safe(router, cfg, false) end,
+        cfg = function(router, cfg)
+            return router_cfg_fiber_safe(router, cfg)
+        end,
```
Gerold103, line 1720: Just do `cfg = router_cfg_fiber_safe`.

Gerold103 (commented):
Thanks for the patch, really nice work!

test/lua_libs/util.lua
Gerold103: Have you investigated why? Could be a bug in storage code.

test/storage-luatest/box_cfg_mode_test.lua
Serpentian: Sure, done

vshard/storage/init.lua
Serpentian: Done

vshard/router/init.lua
Serpentian: Done

vshard/router/init.lua
Serpentian: Done

test/instances/router.lua
Serpentian: Not needed anymore, but thanks anyway

test/lua_libs/util.lua
Serpentian: I didn't, the error looks strange, have no ideas on the reason. Created a ticket for this, https://github.com/tarantool/vshard/issues/452. Will try to investigate it later, don't have time for it right now(

test/instances/router.lua
Serpentian: It's `replication_timeout`, which is set in `vtest.config_new`. Let's not set it there. Fixed

Gerold103 (approved):
Thanks for the fixes!

PR #448, opened 2023-11-22
storage: fix assertion failure in conn_manager

### LINKED ISSUE #446: Assert in `conn_manager_collect_idle_conns()`
In its code
```Lua
                if replica == rs.master and rs.is_master_auto then
                    assert(rs ~= M.this_replicaset)
                    rs.master = nil
                end
```
the assert is not needed. It was for checking that the node never establishes connection to self. But it can actually happen. For example, the rebalancer asks all masters for data via iproto, including self if it is running on master. This is not perf critical anyway.

The patch should
* Drop the assert;
* Add a test (in the test the noactivity timeout constant could be decreased to a few milliseconds for speed);
* Make sure the errors from service functions are propagated. The error was so hard to debug, because each `<service>_f` function wraps its functional part info `pcall` and never raises the returned error. This needs to be done for all services:
```Diff
local function conn_manager_f()
    assert(not M.conn_manager_service)
    local service = lservice_info.new('conn_manager')
    M.conn_manager_service = service
    local ok, err = pcall(conn_manager_service_f, service)
    assert(M.conn_manager_service == service)
    M.conn_manager_service = nil
+    if not ok then
+        error(err)
+    end
end
```

Serpentian:
Closes #446

### DIFF
--- a/test/storage-luatest/auto_master_2_2_2_test.lua
+++ b/test/storage-luatest/auto_master_2_2_2_test.lua
@@ -287,3 +287,36 @@ test_group.test_noactivity_timeout_for_auto_master = function(g)
         _G.bucket_gc_wait()
     end, {g.replica_1_a:replicaset_uuid(), bid1, bid2})
 end
+
+test_group.test_conn_manager_connect_self = function(g)
+    vtest.cluster_rebalancer_enable(g)
+    t.assert_equals(vtest.cluster_rebalancer_find(g), 'replica_1_a')
+
+    g.replica_1_a:exec(function(uuid)
+        -- Create connections using rebalancer
+        local rebalancer = ivshard.storage.internal.rebalancer_service
+        ivtest.service_wait_for_new_ok(rebalancer,
+            {on_yield = ivshard.storage.rebalancer_wakeup})
+
+        -- Assert, that connection to self is created.
+        local replicaset = ivshard.storage.internal.replicasets[uuid]
+        ilt.assert_not_equals(replicaset.master, nil)
+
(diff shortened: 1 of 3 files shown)


test/storage-luatest/auto_master_2_2_2_test.lua
```diff
@@ -287,3 +287,37 @@ test_group.test_noactivity_timeout_for_auto_master = function(g)
...
+test_group.test_conn_manager_connect_self = function(g)
+    vtest.cluster_rebalancer_enable(g)
+    t.assert_equals(vtest.cluster_rebalancer_find(g), 'replica_1_a')
+
+    g.replica_1_a:exec(function(uuid)
+        -- Create connections using rebalancer
+        local rebalancer = ivshard.storage.internal.rebalancer_service
+        ivtest.service_wait_for_new_ok(rebalancer, {on_yield = function()
+            ivshard.storage.internal.rebalancer_fiber:wakeup()
+        end})
```
Gerold103, line 300: Lets do it more compact:
```Lua
        ivtest.service_wait_for_new_ok(rebalancer, {on_yield = ivshard.storage.rebalancer_wakeup})
```

Gerold103 (commented):
Thanks for the quick patch! Only one small comment.

test/storage-luatest/auto_master_2_2_2_test.lua
Serpentian: Done

Gerold103 (approved):
Thanks for the fix!

PR #449, opened 2023-11-22
Schema management preparation patches

Gerold103:
For #435. Those commits prepare some things needed for the schema management mode. This PR is sent early because it brings some commits useful for other PRs right now.

### DIFF
--- a/test/luatest_helpers/vtest.lua
+++ b/test/luatest_helpers/vtest.lua
@@ -75,6 +75,10 @@ local function error_is_timeout(err)
            err.message == 'Timeout exceeded') or err.type == 'TimedOut'
 end
 
+local function clear_test_cfg_options(cfg)
+    cfg.test_user_grant_range = nil
+end
+
 --
 -- Build a valid vshard config by a template. A template does not specify
 -- anything volatile such as URIs, UUIDs - these are installed at runtime.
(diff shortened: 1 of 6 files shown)


test/luatest_helpers/vtest.lua
```diff
@@ -218,15 +222,22 @@ local function cluster_new(g, cfg)
             local user = box.session.user()
             box.session.su('admin')
 
+            local grant_range = cfg.test_user_grant_range
+            ivtest.clear_test_cfg_options(cfg)
             ivshard.storage.cfg(cfg, box.info.uuid)
-            box.schema.user.grant('storage', 'super')
+
+            if grant_range ~= nil then
```
Serpentian, line 229: Nit: Maybe we should grant `super` by default, when `grant_range` is not specified? This way you won't have to change `router_test`

Serpentian (commented):
Thank you! That will be great if we merge that, too many PRs depend on these commits.

Update: feel free to merge, misclicked and submitted comment instead of approve

test/luatest_helpers/vtest.lua
Gerold103: I intentionally didn't want to do that. It should help to catch cases when we somehow miss adding a needed privilege automatically, but it won't be breaking the tests because they would always grant `super`.

test/luatest_helpers/vtest.lua
```diff
@@ -75,6 +75,10 @@ local function error_is_timeout(err)
            err.message == 'Timeout exceeded') or err.type == 'TimedOut'
 end
 
+local function clear_test_cfg_options(cfg)
+    cfg.test_user_grant_range = nil
```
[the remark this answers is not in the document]
Gerold103: Yes, this is the idea.

Serpentian (approved):
Great patch! Thank you. Sorry for such a long review, missed the notification

PR #450, opened 2023-12-01
Fix for working with ipv6 addrs in cartridge

AntiFrizz1:
Changes a needed to support ipv6 in cartridge

This changes needed for https://github.com/tarantool/cartridge/pull/2166

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -1204,7 +1204,7 @@ local replica_mt = {
         safe_uri = function(replica)
             local uri = luri.parse(replica.uri)
             uri.password = nil
-            return luri.format(uri)
+            return util.uri_format(uri)
         end,
         detach_conn = replica_detach_conn,
     },
--- a/test/replicaset-luatest/replicaset_3_test.lua
+++ b/test/replicaset-luatest/replicaset_3_test.lua
@@ -300,3 +300,13 @@ test_group.test_named_replicaset = function(g)
     t.assert_equals(ret, nil)
     vtest.storage_start(g.replica_1_b, global_cfg)
 end
+
+test_group.test_ipv6_uri = function(g)
+    local new_cfg = table.deepcopy(global_cfg)
+    local rs_uuid = g.replica_1_a:replicaset_uuid()
+    local uuid = g.replica_1_a:instance_uuid()
+    new_cfg.sharding[rs_uuid].replicas[uuid].uri = 'storage:storage@[::1]:3301'
+    local _, rs = next(vreplicaset.buildall(new_cfg))
+    local replica_string = 'replica_1_a(storage@[::1]:3301)'
+    t.assert_equals(tostring(rs.master), replica_string)
+end
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -418,6 +418,22 @@ else
     end
 end
 
+local uri_v6_str = "[::1]:80"
+local uri_format
+-- URI module format() function doesn't behave correctly for IPv6 addresses on
+-- some Tarantool versions (all <= 3.0). In fact, at the time of writing none of
+-- the versions have this bug fixed.
+if luri.format(luri.parse(uri_v6_str)) ~= uri_v6_str then
+    uri_format = function(u)
+        if u.ipv6 then
+            u.host = '[' .. u.host .. ']'
+        end
+        return luri.format(u)
+    end
+else
+    uri_format = luri.format
+end
+
 return {
     core_version = tnt_version,
     uri_eq = uri_eq,
@@ -439,4 +455,5 @@ return {
     feature = feature,
     schema_version = schema_version,
     replicaset_uuid = replicaset_uuid,
+    uri_format = uri_format,
 }


vshard/replicaset.lua
```diff
@@ -1033,6 +1033,9 @@ local replica_mt = {
         safe_uri = function(replica)
             local uri = luri.parse(replica.uri)
             uri.password = nil
+            if uri.ipv6 then
```
Serpentian, line 1036: Please, fix the commit message:

1. Commit title must have a prefix, which should state the affected subsystem. I.e. in your case you should have `replicaset: support ipv6 uri`.
2. Add meaningful commit message: describe, what was done and why.
3. At the end of the commit message, please, state, why doc and test was skipped: `NO_DOC=...`, `NO_TEST=...`.

For example please check some commits, which are already in master.

vshard/replicaset.lua
```diff
@@ -1033,6 +1033,9 @@ local replica_mt = {
...
+                uri.host = '[' .. uri.host .. ']'
```
Serpentian, line 1037: I don't understand, why this change is needed. `safe_uri` function is used only to show info in `vshard.storage/router.info` and to print info to logs, is this what you need?

1. Why can't uri be passed with brackets already to vshard config, if it's ipv6?
2. You probably wanted to change `replicaset.buildall` function, not `safe_uri`

Serpentian (changes requested):
Thank you for the patch! Please, check out my comments

vshard/replicaset.lua
AntiFrizz1: ipv6 addr in `replica.uri` passing with brackets, but after `luri.format(uri)` address will be something like `::1:3301`.

Without this changes I got `Failed to bootstrap vshard: [string "local function func(...) local cartridge = re..."]:1: ....rocks/share/tarantool/cartridge/roles/vshard-router.lua:186: Bad argument #1 to errors.netbox_eval (net.box connection expected, got nil)` during `cartridge replicasets setup --bootstrap-vshard` operation. As I understand this happens because of replicaset.master.uri address without brackets.

In instance config file (instances.yml file) we set address with brackets.

vshard/replicaset.lua
AntiFrizz1: Done

vshard/replicaset.lua
sergos: Could you please provide any sort of test.

AntiFrizz1: Done

AntiFrizz1: Plese take a look for this MRs https://github.com/tarantool/membership/pull/55, https://github.com/tarantool/cartridge/pull/2166

test/router/replicaset_master_uri_ipv6.test.lua
```diff
@@ -0,0 +1,25 @@
+test_run = require('test_run').new()
```
Serpentian, line 1: Plese, use luatest, diff-tests are legacy, they're not used anymore

AntiFrizz1: But I just modified tests in router dir. And all test there same. Where I can find corret example of test with luatest?

Serpentian: Also, the test does nothing, but checks, that the returned URI is inside `[...]`.

Does vshard cluster fails to start, when ipv6 are used?

I see, that you're trying to make the returned value not `::1:3301` but `[::1:3301]`, when ipv6 is used.
If vshard starts vshard cluster correctly with ipv6, then it's probably cartridge, which is needed to be fixed.
If vshard doesn't, then we should test, that after this change cluster operates with ipv6.

Serpentian: > But I just modified tests in router dir. And all test there same. Where I can find corret example of test with luatest?

You can use router for that, the tests are inside `test/router-luatest/*` dir. But I'd recommend writing a unit test, which checks replicaset explicitly, inside `test/replicaset-luatest/replicaset_3_test.lua`

AntiFrizz1: > Does vshard cluster fails to start, when ipv6 are used?

vshard cluster starts correctly with ipv6, but cartridge uses router:info()['<replicaset-uuid>'].master.uri in `cartridge/roles/vshard-router.lua` in `bootstrap_group` to connect to router. And ::1:3301 will be incorrect. Thats why I need this changes.

Serpentian: > vshard cluster starts correctly with ipv6, but cartridge uses router:info()['<replicaset-uuid>'].master.uri in cartridge/roles/vshard-router.lua in bootstrap_group to connect to router. And ::1:3301 will be incorrect. Thats why I need this changes.

Can't cartridge take `router:info()['<replicaset-uuid>'].master.uri` and add `[..]` if needed?

AntiFrizz1: Probably can, but in order to do that we need uri parser that can parse ipv6 uris without `[..]` because default `uri.parse` can't. And router still will store replicaset instance uri that inappropriate with ipv6 uri standart.

Gerold103: I have to agree with @AntiFrizz1 here. IPv6 addresses need to be wrapped into `[]` before adding `:<port>`. Otherwise the address is unparsable, because both IP parts and the port are using `:` delimiter in IPv6.

Serpentian: Indeed, according to [RFC 2732](https://www.rfc-editor.org/rfc/rfc2732) ipv6 must be wrapped into brackets. Uri module works wrong wit ipv6 now, created the ticket for that: https://github.com/tarantool/tarantool/issues/9556.

As we must support old Tarantool version, let's go with the change, proposed in this PR. However, let's make test with luatest, not difftest. Here it's:

```diff
diff --git a/test/replicaset-luatest/replicaset_3_test.lua b/test/replicaset-luatest/replicaset_3_test.lua
index 0107e7b..c8f8296 100644
--- a/test/replicaset-luatest/replicaset_3_test.lua
+++ b/test/replicaset-luatest/replicaset_3_test.lua
@@ -300,3 +300,14 @@ test_group.test_named_replicaset = function(g)
     t.assert_equals(ret, nil)
     vtest.storage_start(g.replica_1_b, global_cfg)
 end
+
+test_group.test_ipv6_uri = function(g)
+    local new_cfg = table.deepcopy(global_cfg)
+    local rs_uuid = g.replica_1_a:replicaset_uuid()
+    local uuid = g.replica_1_a:instance_uuid()
+    new_cfg.sharding[rs_uuid].replicas[uuid].uri = 'storage:storage@[::1]:3301'
+
+    local _, rs = next(vreplicaset.buildall(new_cfg))
+    local replica_string = 'replica_1_a(storage@[::1]:3301)'
+    t.assert_equals(tostring(rs.master), replica_string)
+end

```

AntiFrizz1: Added test with luatest. Removed previous test implementation.

vshard/replicaset.lua
```diff
@@ -1204,6 +1204,9 @@ local replica_mt = {
...
+            end
             return luri.format(uri)
```
Serpentian, line 1210: Nit: in the future, when https://github.com/tarantool/tarantool/issues/9556 is fixed, we'll have to make this brackets adding depend on Tarantool's version, as it's actually a bug in `uri.format` function.

So, I'd rather move this check (for brackets) alongside with `luri.format` to `utils.lua` and name it smth like `uri_format`. This way we'll have to make minimal changes in order to introduce version specific `uri_format` function. Up to you and @Gerold103, I'm ok with the current implementation too.

Moreover, we should create a ticket for introducing version dependent `uri_format` function, when this PR is merged.

Serpentian (approved):
Thank you for the patch! Great work)

vshard/replicaset.lua
AntiFrizz1: I'd like to see these changes merged as soon as possible to get them into the next release.

vshard/replicaset.lua
```diff
@@ -1204,6 +1204,9 @@ local replica_mt = {
         safe_uri = function(replica)
             local uri = luri.parse(replica.uri)
             uri.password = nil
+            if uri.ipv6 then
+                uri.host = '[' .. uri.host .. ']'
+            end
```
Gerold103, line 1209: While it works on the current Tarantool versions, it will not work once `uri.format` is fixed in the core. Your code will then produce incorrect URIs again - with `[[` and `]]`. Fixing it in vshard will not help if we will release at least one vshard version with the current solution - it simply won't work on new Tarantool cores.

The only solution I see is to make it work both for current and future behaviour together. For that you need to do what @Serpentian said: define a helper in `util` module - `uri_format`. It should add the brackets or not, depending on how the core uri module behaves.

A possible solution could be this (in `util.lua`):
```Lua
local uri_v6_str = "[::1]:80"
local uri_format
-- URI module format() function doesn't behave correctly for IPv6 addresses on
-- some Tarantool versions (all <= 3.0). In fact, at the time of writing none of
-- the versions have this bug fixed.
if uri.format(uri.parse(uri_v6_str)) ~= uri_v6_str then
    uri_format = function(u)
        -- Implement here your hack with the brackets.
    end
else
    uri_format = uri.format
end
-- ...
-- In the end of the `util.lua` file add `uri_format` function to the module exports.
```
And then make all the vshard code use `util.uri_format` instead of core's `uri.format`.

Gerold103 (changes requested):
Thanks for the patch!

vshard/replicaset.lua
AntiFrizz1: Fixed

Gerold103 (approved):
Thanks for the patch!

Gerold103:
I just realized that from what you are saying, you have code which touches `replicaset.master.uri` directly. This is internal code, not exposed to the public API and it might be changed in the future. I strongly advice you not to touch the vshard internals in your code.

PR #451, opened 2023-12-04
Introduce named config

### LINKED ISSUE #426: Support instance/replicaset/cluster names instead of UUIDs in the configuration
See #407 and https://github.com/tarantool/tarantool/pull/8289.

The key difference from the configuration using UUIDs is that instance/replicaset/cluster names are known at the configuration stage.

It is especially important in case of a cluster configuration introduced in https://github.com/tarantool/tarantool/issues/8724.

There is a workaround: use a hash function to generate UUIDs from names. However, it complicates upgrading from old snapshots, where there are no persistent instance/replicaset/cluster names.

I would appreciate hints about possible pitfalls on the road. What else we need to do except just passing the parameters down to `box.cfg()`?

Serpentian:
This patchset allows using names instead of UUIDs on storage and router. This blocks config module from dropping UUIDs in favor of names.

Closes #426


vshard/replicaset.lua
```diff
@@ -402,6 +415,40 @@ local function replica_on_success_request(replica)
     end
 end
 
+--
+-- Validate instance_name if neeed and make a call.
+--
+local function replica_conn_call(replica, func, args, opts)
```
Serpentian, line 421: Doesn't look too great, but I suppose, we really need this. Don't think, that `name` will be added to the netbox greeting in any foreseeable future. Up to you, it's not strictly required to allow configuring with names.

vshard/cfg.lua
```diff
@@ -257,17 +288,23 @@ local function cfg_check_weights(weights)
...
-        if uuids[replicaset_uuid] then
-            error(string.format('Duplicate uuid %s', replicaset_uuid))
+    local rebalancer_id
+    if ctx.is_named then
+        -- It's allowed to have the same names for replicaset and replica.
+        ids.replicasets = {}
+        ids.replicas = {}
+    end
+    for replicaset_id, replicaset in pairs(sharding) do
+        local rs_ids = ctx.is_named and ids.replicasets or ids
```
Gerold103, line 303: This line doesn't need to be in the loop. Same with `replica_ids`.

vshard/storage/init.lua
```diff
@@ -3743,20 +3744,44 @@ local function storage_cfg_build_local_box_cfg(cfgctx)
...
-        end
-        local real_rs_uuid = util.replicaset_uuid(info)
-        if cfgctx.replicaset_uuid ~= real_rs_uuid then
-            error(string.format('Replicaset UUID mismatch: already set ' ..
-                                '"%s" but "%s" in vshard config',
-                                real_rs_uuid, cfgctx.replicaset_uuid))
+        if cfgctx.new_cfg.identification_mode == 'name_as_key' then
+            -- Strictly verify names in case of reconfiguration. There's no
+            -- other way to check, that config is applied to the correct
+            -- instance. Changing and applying names must be done externally.
```
Gerold103, line 3762: Not necessarily externally. If the names are used from the beginning, then vshard will set them on bootstrap in the code block above.

vshard/storage/init.lua
```diff
@@ -3834,9 +3859,9 @@ local function storage_cfg_master_commit(cfgctx)
 end
 
 local function storage_cfg_context_extend(cfgctx)
-    local instance_uuid = cfgctx.instance_uuid
-    if instance_uuid == nil then
-        error('Usage: cfg(configuration, this_replica_uuid)')
+    local instance_id = cfgctx.instance_id
+    if instance_id == nil then
+        error('Usage: cfg(configuration, this_replica_id)')
```
Gerold103, line 3865: Can all those uuid->id renames be done in a separate commit? You did it for replicaset module, but somehow the much bigger refactoring of storage you left merged with the `name_as_key` introduction.

Same for the router.

test/router-luatest/router_test.lua
```diff
@@ -649,3 +649,50 @@ g.test_router_service_info = function(g)
     -- Restore everything back.
     vtest.router_cfg(g.router, global_cfg)
 end
+
+g.test_named_config_identification = function(g)
+    t.run_only_if(vutil.version_is_at_least(3, 0, 0, 'entrypoint', 0, 0))
```
Gerold103, line 654: `vutil.feature.persistent_names`.

vshard/replicaset.lua
```diff
@@ -225,6 +231,13 @@ local function replicaset_connect_to_replica(replicaset, replica)
         conn:on_connect(netbox_on_connect)
         conn.on_disconnect_ref = netbox_on_disconnect
         conn:on_disconnect(netbox_on_disconnect)
+        -- Whether name validation is needed.
+        conn.name = replica.id == replica.name
```
Gerold103, line 235: Please, set it before the triggers. Just to be sure that if the triggers would somehow run right away, they would see the flag.

Gerold103, line 235: This is a cool idea - getting the greeting and validating it on first access. We can piggyback here many other things, like vshard version, vshard features, etc. Lets make it more generic and introduce `conn.vconnect` - it would be a table storing the async requests for various stuff such as the name.

It would store members `is_named` (flags should have `is_` suffix or similar) and the other things that you need. Later we can add here more features, subscriptions, etc.

test/replicaset-luatest/replicaset_3_test.lua
```diff
@@ -2,6 +2,7 @@ local fiber = require('fiber')
 local t = require('luatest')
```
Gerold103, line 2: > Let's validate name only during connection establishing, not on every reconnect

Please, do it on every reconnect. It is not as expensive as it looks, but can protect from accidental URL changes with nasty consequences when the underlying instance could change to a different one if somebody messed up DNS.

vshard/replicaset.lua
Serpentian: Is it ok to use `storage_service_info`, like it's done now or do we want to introduce new call, specific for `conn.vconnect`? This means schema version bump, so I'd rather stay with `storage_service_info`.

Gerold103: No, this is fine. And it wouldn't cause the version bump - `storage_service_info` works via `vshard.storage._call` which already exists and was created specifically for cases when we don't want to bother with schema changes but do want to change functions.

test/luatest_helpers/vtest.lua
```diff
@@ -138,9 +148,11 @@ local function config_new(templ)
                     }
                 }
             end
-            replicas[replica_uuid] = replica
+            local replica_id = is_named and replica_name or replica_uuid
```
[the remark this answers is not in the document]
Serpentian: Agree, now it's called only when it's actually used

test/luatest_helpers/vtest.lua
```diff
@@ -156,7 +168,10 @@ local function cluster_new(g, cfg)
     local masters = {}
     local replicas = {}
     local master_map = {}
-    for replicaset_uuid, replicaset in pairs(cfg.sharding) do
+    local is_named = cfg.identification_mode == 'name_as_key'
+    for replicaset_id, replicaset in pairs(cfg.sharding) do
+        local replicaset_uuid, replicaset_name = vutil.cfg_extract_identifiers(
+            replicaset_id, replicaset, is_named)
```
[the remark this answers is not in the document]
Serpentian: If `is_named = true`, then we don't set automatically generated UUIDs, as in general we should assume, that no UUIDs were passed and test everything like that.

But this doesn't prevents us from specifying `replica.uuid` in the replica's template config:

https://github.com/tarantool/vshard/blob/c13da010efc7dcf3d891da88603733dd21988281/test/luatest_helpers/vtest.lua#L99

So, it's intended. UUID can be set, when `is_named = true`

test/luatest_helpers/vtest.lua
```diff
@@ -224,7 +244,8 @@ local function cluster_new(g, cfg)
 
             local grant_range = cfg.test_user_grant_range
             ivtest.clear_test_cfg_options(cfg)
-            ivshard.storage.cfg(cfg, box.info.uuid)
+            ivshard.storage.cfg(cfg, cfg.identification_mode == 'name_as_key'
+                                and box.info.name or box.info.uuid)
```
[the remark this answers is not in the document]
Serpentian: Introduced `_G.box_vstorage_id` in `instances/storage.lua`

test/unit-luatest/config_test.lua
```diff
@@ -290,3 +290,85 @@ g.test_enum = function()
         vcfg.check, config)
```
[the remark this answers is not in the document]
Serpentian: My bad, forgot to wrap config examples in ```. Suppose, you meant that, fixed

test/unit-luatest/config_test.lua
```diff
@@ -290,3 +290,85 @@ g.test_enum = function()
...
+    config.identification_mode = 'bad'
+    t.assert_error_msg_content_equals(
+        "Config identification mode must be enum " ..
+        "{'uuid_as_key', 'name_as_key', nil}",
+        vcfg.check, config)
+    config.identification_mode = nil
+end
+
+g.test_identification_mode_name_as_key = function()
+    t.run_only_if(vutil.version_is_at_least(3, 0, 0, 'entrypoint', 0, 0))
```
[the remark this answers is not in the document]
Serpentian: Sure, it's used everywhere now

vshard/cfg.lua
Serpentian: Indeed, this logic is located before the loop now.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -530,3 +530,63 @@ test_group.test_noactivity_timeout_for_explicit_master = function(g)
...
+    -- Test, that sending by name works properly
+    local rs_name_2 = g.replica_2_a:replicaset_name()
+    t.assert_equals(rs_name_2, 'replicaset_2')
+    g.replica_1_a:exec(function(name)
+        local bid = _G.get_first_bucket()
+        local ok, err = ivshard.storage.bucket_send(
+            bid, name, {timeout = _G.iwait_timeout})
+        ilt.assert_equals(err, nil)
+        ilt.assert(ok)
+    end, {rs_name_2})
```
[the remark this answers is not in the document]
Serpentian: Sure, now it's validated, that bucket is sent back, when rebalancer works

vshard/storage/init.lua
Serpentian: Yeah, fixed the commit message

vshard/storage/init.lua
Serpentian: Sure, both storage and router commits were divided to several ones.

test/failover/cluster_changes.result
```diff
@@ -78,34 +78,34 @@ vshard.router.info()
         status: missing
```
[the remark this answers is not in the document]
Serpentian: Both commit messages `storage/router: export name to info` were changed. It's not a backward compatibility question, UUID can be specified alongside with names, we should show both of them.

vshard/replicaset.lua
Serpentian: Sure, done

Serpentian: > And it wouldn't cause the version bump

My bad.

> Lets make it more generic and introduce conn.vconnect

Done

test/replicaset-luatest/replicaset_3_test.lua
Serpentian: Now it's validated on every reconnect

vshard/replicaset.lua
```diff
@@ -402,6 +415,40 @@ local function replica_on_success_request(replica)
...
 end
 
+--
+-- Validate instance_name if neeed and make a call.
+--
+local function replica_conn_call(replica, func, args, opts)
+    local conn = replica.conn
+    if conn.name then
+        if conn.name == true and
+           not fiber_cond_wait(conn.name_cond, opts.timeout) then
```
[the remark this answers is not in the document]
Serpentian: Great idea! Everything is done.

vshard/storage/init.lua
```diff
@@ -3468,6 +3468,7 @@ end
 local function storage_service_info()
     return {
         is_master = this_is_master(),
+        name = box.info.name
```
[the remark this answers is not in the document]
Serpentian: Added

vshard/util.lua
```diff
@@ -417,6 +418,26 @@ else
     end
 end
 
+--
+-- Extract uuid and name from the replicaset/replica config. If is_named = true,
+-- then identification_mode is 'name_as_key', new type of config is used:
+-- names becomes identifiers instead of UUIDs, replica/replicaset.name
+-- is not specified, UUIDs are specified as replica/replicaset.uuid.
+--
+local function cfg_extract_identifiers(cfg_key, cfg_value, is_named)
```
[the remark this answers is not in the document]
Serpentian: It's in `cfg` now

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -531,3 +531,79 @@ test_group.test_noactivity_timeout_for_explicit_master = function(g)
         _G.bucket_gc_wait()
     end, {g.replica_1_a:replicaset_uuid(), bid})
 end
+
```
Gerold103, line 534: Need a storage test which boots a new cluster with names. Otherwise this is uncovered now, whether vshard is able to set the names for the new cluster. I suggest to do it in an new file `storage-luatest/persistent_names_test.lua`. (The test `test_group.test_named_config_identification` can remain in storage_1_1_test.lua, because requires the cluster to be booted without names.)

Also in that new file you can test the switch from names to UUIDs.

test/storage-luatest/storage_1_1_test.lua
```diff
@@ -531,3 +531,79 @@ test_group.test_noactivity_timeout_for_explicit_master = function(g)
...
+        ilt.assert_equals(err, nil)
+        ilt.assert(ok)
+        return bid
+    end, {rs_name_2})
+
+    -- Test, that rebalancer is also ok.
+    vtest.cluster_rebalancer_enable(g)
+    g.replica_1_a:exec(function(bid)
+        local internal = ivshard.storage.internal
+        ivtest.service_wait_for_ok(internal.rebalancer_service,
```
Gerold103, line 599: `service_wait_for_new_ok`. We don't want to accidentally catch the moment when the service was ok because it didn't see the bucket being moved yet.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+        -- Whether the connection is done to the named replica.
+        is_named = conn.replica.id == conn.replica.name,
+        -- Used to wait for the appearance of vconnect.future object,
+        -- if call is done before the connection is established.
+        future_cond = fiber.cond(),
+    }
+    local vconn = conn.vconnect
+    if not vconn.is_named then
+        -- Skip validation without any waiting.
+        conn.vconnect.is_validated = true
```
Gerold103, line 97: I suggest to simply do `conn.vconnect = nil`.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    end
+end
+
+--
+-- Initialize future object. Should be done, when connection is established
+-- (inside netbox_on_connect).
+--
+local function conn_init_vconnect(conn)
```
Gerold103, line 105: `conn_vconnect_start()`.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    end
+end
+
+--
+-- Check, that future is ready, and its result is expected, close the connection
+-- if the result is not correct. The function doesn't yield.
+-- @retval true; The correct response is received.
+-- @retval nil, ...; Response is not received or validation error happened.
+--
+local function conn_validate_vconnect(conn)
```
Gerold103, line 122: `conn_vconnect_check()` (assuming that `conn_vconnect_` is our object-prefix, and that we usually call such validation procedures just `check`).

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
 local gsc = util.generate_self_checker
 
+--
+-- vconnect is asynchronous vshard greeting, saved inside netbox connection.
+-- It stores future object and additional info, needed for its work.
+-- Future is initialized, when the connection is established (inside
+-- netbox_on_connect). The connection cannot be considered "connected"
+-- until vconnect is properly validated.
+--
+local function conn_set_vconnect(conn)
```
Gerold103, line 85: `conn_vconnect_set()`.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+local function conn_validate_vconnect(conn)
+    local vconn = conn.vconnect
+    -- conn.vconnect may be nil, if connection was created on old version
+    -- and the storage was reloaded to a new one.
+    if not vconn or vconn.is_validated then
+        return true
+    end
+    -- Nothing to do, but wait in such case.
+    if not vconn.future or not vconn.future:is_ready() then
+        return nil, lerror.vshard(lerror.code.CONNECTION_IS_NOT_ESTABLISHED,
```
Gerold103, line 131: Lets either set the same error as netbox when it is not connected, or call this error code `VHANDSHAKE_NOT_COMPLETE` to emphasise that the TCP connection is done, but the handshake is not.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    vconn.is_validated = true
+    return true
+end
+
+--
+-- Wait until the future object is ready. Returns remaining timeout.
+-- @retval timeout; Future boject is ready.
+-- @retval nil, err; Timeout passed.
+--
+local function conn_wait_vconnect(conn, timeout)
```
Gerold103, line 156: `conn_vconnect_wait()`.

vshard/replicaset.lua
```diff
@@ -421,6 +525,26 @@ local function replica_call(replica, func, args, opts)
...
+    else
+        local timeout, err = conn_wait_vconnect(conn, opts.timeout)
+        if not timeout then
+            return false, nil, lerror.make(err)
+        end
+        local ok
+        ok, err = conn_validate_vconnect(conn)
+        if not ok then
+            return false, nil, lerror.make(err)
+        end
```
Gerold103, line 545: This block can be done outside of the big `if`. It is the same both for async and sync requests.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    -- Nothing to do, but wait in such case.
+    if not vconn.future or not vconn.future:is_ready() then
+        return nil, lerror.vshard(lerror.code.CONNECTION_IS_NOT_ESTABLISHED,
+                                  conn.replica.id)
+    end
+    -- Critical errors. Connection should be closed after these ones.
+    local result, err = vconn.future:result()
+    if not result then
+        conn:close()
+        -- Failed to get response. E.g. access error.
```
Gerold103, line 138: Can you add a test on that? Should be easy if you just do this `vshard.storage._call = nil` on the storage side.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+        return nil, lerror.make(err)
+    end
+    if result[1].name ~= conn.replica.name then
+        conn:close()
+        return nil, lerror.vshard(lerror.code.INSTANCE_NAME_MISMATCH,
+                                  conn.replica.name, result[1].name)
+    end
+    -- Don't validate until reconnect happens.
+    vconn.is_validated = true
```
Gerold103, line 147: I suggest to add a test on when after reconnect the name is different. To check that the validation is done on each reconnect.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    -- and the storage was reloaded to a new one.
+    if not vconn or vconn.is_validated or
+       (vconn.future and vconn.future:is_ready()) then
+        return timeout
+    end
+    local deadline = fiber_clock() + timeout
+    -- Wait for connection to be established.
+    if not vconn.future and
+       not fiber_cond_wait(vconn.future_cond, timeout) then
+        return nil, lerror.timeout()
```
Gerold103, line 169: Can you add a test on that? Should be easy if you simply change `vshard.storage._call` to a function which just does a busy-loop fiber sleep.

vshard/replicaset.lua
```diff
@@ -75,6 +75,108 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    end
+    timeout = deadline - fiber_clock()
+    -- Wait for async call to return.
+    local res, err = util.future_wait(vconn.future, timeout)
+    if res == nil then
+        return nil, lerror.make(err)
```
Gerold103, line 175: Same as before - can set `storage._call` to nil and make a synchronous request to check that. The previous occurrence could be tested with an `is_async = true` request.

vshard/replicaset.lua
```diff
@@ -421,6 +525,26 @@ local function replica_call(replica, func, args, opts)
     assert(opts and opts.timeout)
     replica.activity_ts = fiber_clock()
     local conn = replica.conn
+    if opts.is_async then
```
Gerold103, line 528: Need a test that for an `is_async` request the call fails instantly without any yields if the connection is not ready.

Gerold103 (changes requested):
Thanks for the fixes!

vshard/replicaset.lua
Serpentian: https://github.com/tarantool/vshard/issues/456

test/luatest_helpers/server.lua
```diff
@@ -206,6 +206,30 @@ function Server:replicaset_uuid()
     return uuid
 end
 
+-- Return instance name. Only for Tarantool >= 3.0.0. instance_name may be
+-- changed, so opts.update_cache is accepted for updating cached name.
+function Server:instance_name(opts)
+    opts = opts or {}
+    if self.instance_name_value and not opts.update_cache then
```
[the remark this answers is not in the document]
Serpentian: > Btw, you perhaps don't even need these opts. Could simply nullify server. instance_name_value in the tests which require its update.

Let's drop it at all. Indeed

test/storage-luatest/storage_1_1_test.lua
Serpentian: Added the test. It's good, it was tested, `vtest.cluster_bootstrap` was broken, when name identification is used.

test/storage-luatest/storage_1_1_test.lua
Serpentian: Sure, done

vshard/replicaset.lua
Serpentian: Done, but it requires calling `conn_vconnect_set` inside `on_connect` trigger.

vshard/replicaset.lua
Serpentian: Renamed

vshard/replicaset.lua
Serpentian: Renamed

vshard/replicaset.lua
Serpentian: Renamed

vshard/replicaset.lua
Serpentian: I'd rather go with `VHANDSHAKE_NOT_COMPLETE`. It's easier to distinguish it from connection error and will help fixing bugs in the future

vshard/replicaset.lua
Serpentian: Renamed

vshard/replicaset.lua
Serpentian: You're right, done

vshard/replicaset.lua
Serpentian: Added tests for all cases, you asked

vshard/replicaset.lua
Serpentian: Added

vshard/replicaset.lua
Serpentian: Added

vshard/replicaset.lua
Serpentian: Added

test/storage-luatest/storage_persistent_names_test.lua
```diff
@@ -0,0 +1,118 @@
+local t = require('luatest')
```
Gerold103, line 1: No need to call the file `storage-luatest/storage_persistent_names_test.lua` - it is in the storage folder anyway. Can be called just `storage-luatest/persistent_names_test.lua`.

test/storage-luatest/storage_persistent_names_test.lua
```diff
@@ -0,0 +1,118 @@
...
+local vtest = require('test.luatest_helpers.vtest')
+local vutil = require('vshard.util')
+
+local group_config = {{engine = 'memtx'}, {engine = 'vinyl'}}
```
Gerold103, line 5: The `engine` param is not used. I suggest to drop it then. Given that you don't use any spaces here, could also drop `memtx_use_mvcc_engine`.

test/storage-luatest/storage_persistent_names_test.lua
```diff
@@ -0,0 +1,118 @@
...
+    name = g.replica_2_a:instance_name()
+    t.assert_equals(rs_name, 'replicaset_2')
+    t.assert_equals(name, 'replica_2_a')
+end
+
+--
+-- Test named config to UUID config switch.
+--
+test_group.test_switch_to_uuid = function(g)
+    t.run_only_if(vutil.feature.persistent_names)
```
Gerold103, line 75: I don't think you need it if already done the same in `before_all`, do you?

test/replicaset-luatest/replicaset_3_test.lua
```diff
@@ -247,6 +248,7 @@ test_group.test_locate_master_when_no_conn_object = function(g)
 end
 
 test_group.test_named_replicaset = function(g)
+    t.run_only_if(vutil.feature.persistent_names)
```
Gerold103, line 251: How did it work in the previous commits without this check? You are using `identification_mode` here already in older commits, but didn't have the `run_only_if` before.

vshard/replicaset.lua
```diff
@@ -75,6 +75,141 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    local res, err = util.future_wait(vconn.future, timeout)
+    if res == nil then
+        -- Either timeout error or any other. If it's not a timeout error,
+        -- then conn must be recteated, handshake must be retried.
+        return nil, lerror.make(err)
+    end
+    return deadline - fiber_clock()
+end
+
+local function conn_vconnect_wait_and_close(conn, timeout)
```
Gerold103, line 205: Given how it works, it is rather `wait_or_close`, not `and`. Same for `check_and_close`.

Gerold103 (changes requested):
Thanks for the fixes!

vshard/replicaset.lua
Gerold103: Are you sure? I don't see any new tests.

vshard/replicaset.lua
```diff
@@ -75,6 +75,141 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    -- and the storage was reloaded to a new one. It's also nil, when
+    -- all checks were already done.
+    if not vconn then
+        return true
+    end
+    -- Nothing to do, but wait in such case.
+    if not vconn.future or not vconn.future:is_ready() then
+        return nil, lerror.vshard(lerror.code.VHANDSHAKE_NOT_COMPLETE,
+                                  conn.replica.id)
+    end
```
Gerold103, line 150: You need to check the error type first. Otherwise this condition will be true also when `error.code` equals to something from `box.error.`

test/storage-luatest/storage_persistent_names_test.lua
Serpentian: Done

test/storage-luatest/storage_persistent_names_test.lua
Serpentian: Dropped `group_config`

vshard/replicaset.lua
Serpentian: Oh, sorry. It remained in untracked files and wasn't added to commit

vshard/replicaset.lua
Serpentian: Renamed

vshard/replicaset.lua
Serpentian: I don't get it. When some error happens, `vconn.future:is_ready()` equals to true and we'll filter it out below. When there's no result, it will be in this check:

```
tarantool> function test() box.error('error') end
tarantool> f = c:eval('test', {}, {is_async = true})
tarantool> f:is_ready()
---
- true
...

tarantool> f:result()
---
- null
- Execute access to universe '' is denied for user 'guest'
```

Gerold103: Sorry, I left the comment in a wrong place. I've meant this:
```Lua
local function conn_vconnect_check_or_close(conn)
    local ok, err = conn_vconnect_check(conn)
    if not ok and err.code ~= lerror.code.VHANDSHAKE_NOT_COMPLETE then
        conn:close()
    end
    return ok, err
end
```
If `err` is a `ClientError` with code `ER_WAL_IO` (= 40, like `VHANDSHAKE_NOT_COMPLETE`), then your condition will pass. So you need to check the error type to be equal to `ShardingError` before you can check the code.

test/unit-luatest/vconnect_test.lua
```diff
@@ -0,0 +1,181 @@
...
+    end)
+
+    rs.master.conn = nil
+    local opts = table.deepcopy(timeout_opts)
+    opts.is_async = true
+    local ret, err = rs:callrw('get_uuid', {}, timeout_opts)
+    t.assert_str_contains(err.message, "_call' is not defined")
+    t.assert_equals(ret, nil)
+    t.assert_equals(rs.master.conn.state, 'closed')
+
```
Gerold103, line 109: You set `opts` but never use it. This test is identical to `test_vconnect_no_result`. Something is wrong here.

test/unit-luatest/vconnect_test.lua
```diff
@@ -0,0 +1,181 @@
...
+test_group.test_async_no_yield = function(g)
+    local _, rs = next(vreplicaset.buildall(global_cfg))
+    g.replica:exec(function()
+        rawset(_G, '_call', ivshard.storage._call)
+        -- Future should not appear at all.
+        ivshard.storage._call = function()
+            while true do
+                ifiber.sleep(0.1)
+            end
+        end
```
Gerold103, line 164: Ditto as before about an unstoppable sleep.

Gerold103 (changes requested):
Thanks for the fixes!

test/unit-luatest/vconnect_test.lua
```diff
@@ -0,0 +1,181 @@
...
+}
+local global_cfg
+
+test_group.before_all(function(g)
+    t.run_only_if(vutil.feature.persistent_names)
+    global_cfg = vtest.config_new(cfg_template)
+    vtest.cluster_new(g, global_cfg)
+    vtest.cluster_bootstrap(g, global_cfg)
+    vtest.cluster_wait_vclock_all(g)
+end)
```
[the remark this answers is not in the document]
Serpentian: Sure, it's in `replicaset-luatest` now

test/unit-luatest/vconnect_test.lua
```diff
@@ -0,0 +1,181 @@
...
+test_group.test_vconnect_no_future = function(g)
+    local _, rs = next(vreplicaset.buildall(global_cfg))
+    g.replica:exec(function()
+        rawset(_G, '_call', ivshard.storage._call)
+        -- Future should not appear at all.
+        ivshard.storage._call = function()
+            while true do
+                ifiber.sleep(0.1)
+            end
+        end
```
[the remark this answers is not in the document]
Serpentian: Added `_G.do_sleep` for that

test/unit-luatest/vconnect_test.lua
Serpentian: Done

test/unit-luatest/vconnect_test.lua
Serpentian: Thank you for noticing! We must use `opts` and retry until the time, actual error is shown. Fixed

vshard/replicaset.lua
Serpentian: Now I see and agree. `err.type` is also checked now.

vshard/replicaset.lua
```diff
@@ -75,6 +75,144 @@ local fiber_cond_wait = util.fiber_cond_wait
...
+    local deadline = fiber_clock() + timeout
+    -- Wait for connection to be established.
+    if not vconn.future and
+       not fiber_cond_wait(vconn.future_cond, timeout) then
+        return nil, lerror.timeout()
+    end
+    timeout = deadline - fiber_clock()
+    -- Wait for async call to return.
+    local res, err = util.future_wait(vconn.future, timeout)
```
Gerold103, line 199: There is `future_wait` function in replicaset.lua as a variable. You don't need `util.` prefix.

vshard/replicaset.lua
```diff
@@ -222,6 +365,7 @@ local function replicaset_connect_to_replica(replicaset, replica)
         })
         conn.replica = replica
         conn.replicaset = replicaset
+        conn_vconnect_set(conn)
```
Gerold103, line 368: Why do you need to add it here? It is not connected yet anyway. And on connect the vconnect will be added in the code above.

vshard/replicaset.lua
```diff
@@ -421,6 +565,26 @@ local function replica_call(replica, func, args, opts)
...
+        -- something is wrong.
+        local ok, err = conn_vconnect_check_or_close(conn)
+        if not ok then
+            return false, nil, lerror.make(err)
+        end
+    end
+    local timeout, err = conn_vconnect_wait_or_close(conn, opts.timeout)
+    if not timeout then
+        return false, nil, lerror.make(err)
+    end
```
Gerold103, line 580: This shouldn't be on the `is_async` path, should it? For `is_async` it would either yield or return immediately which makes it either needless or a bug.

Gerold103 (changes requested):
Thanks for the fixes! We are on the finish line.

vshard/replicaset.lua
Serpentian: It's NoOp in case of async, we already checked, that it's ready. This is the reason, why all tests passses.

I changed the code in this part a little bit, hope it's ok now.

vshard/replicaset.lua
Serpentian: Fixed

Gerold103 (approved):
Thanks for the fixes!

PR #454, opened 2023-12-11
Fix and speed up `rebalancer/receiving_bucket.test.lua`

Gerold103:
See the commits for details.

### DIFF
--- a/test/rebalancer/receiving_bucket.test.lua
+++ b/test/rebalancer/receiving_bucket.test.lua
@@ -65,7 +65,9 @@ vshard.storage.internal.errinj.ERRINJ_RECEIVE_PARTIALLY = true
 _ = test_run:switch('box_2_a')
 res, err = vshard.storage.bucket_send(1, util.replicasets[1], {timeout = 10})
 res, util.portable_error(err)
-box.space._bucket:get{1}
+status = box.space._bucket:get{1}.status
+-- Could be active if recovered quickly enough.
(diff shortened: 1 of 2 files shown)


test/rebalancer/receiving_bucket.test.lua
```diff
@@ -33,8 +33,13 @@ create_simple_space('test3', {engine = 'vinyl'})
 create_simple_space('test4')
 create_simple_space('test5', {engine = 'vinyl'})
 
-for i = 1, 10000 do box.space.test:replace{i, 1, 1} box.space.test2:replace{i, 1, 2} box.space.test3:replace{i, 1, 3} end
-for i = 1, 500 do box.space.test4:replace{i, 1, 4} box.space.test5:replace{i, 1, 5} end
+box.begin() for i = 1, 10000 do box.space.test:replace{i, 1, 1} end box.commit()
+box.begin() for i = 1, 10000 do box.space.test2:replace{i, 1, 2} end box.commit()
```
Serpentian, line 37: Can be done in the same transaction, where `test` is replaced. For me it's 2.94 sec without this change vs 2.66 with it.

Serpentian (approved):
Thank you for the patch! Great work!

test/rebalancer/receiving_bucket.test.lua
Gerold103: Done.

PR #455, opened 2023-12-14
router: revert dropping box.cfg and introduce `box_cfg_mode`

Serpentian:
1. https://github.com/tarantool/doc/issues/3895 should be closed after merging. 
2. https://github.com/tarantool/vshard/issues/353 should be reopened and closed as not planned.

Note, that https://github.com/tarantool/vshard/commit/a5304557e052f95541ab44953c5aa882f9f58dce intentionally wasn't reverted. I suppose, it's correct. We should not set `box.cfg.replication_timeout` during config generation. It should be done inside template, storages need this option, routers - don't.


Gerold103 (approved):
Thanks for the patch!

PR #457, opened 2023-12-15
storage: add support for roles in deploy_privs()

ImeevMA:
The deploy_privs() function will now grant the required privileges to a user or a role.

Follow-up #435

### DIFF
--- a/test/unit-luatest/storage_exports_test.lua
+++ b/test/unit-luatest/storage_exports_test.lua
@@ -96,3 +96,42 @@ test_group.test_core_upgrade = function(g)
         ilt.assert(func_tuple.opts.takes_raw_args)
     end)
 end
+
+-- Make sure that privileges can be granted to both the user and the role.
+test_group.test_deploy_privs = function(g)
+    g.server:exec(function()
+        local vexports = require('vshard.storage.exports')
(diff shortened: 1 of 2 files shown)


test/unit-luatest/storage_exports_test.lua
```diff
@@ -96,3 +96,39 @@ test_group.test_core_upgrade = function(g)
...
+        vexports.deploy_privs(exports, 'one')
+        for func_name in pairs(exports.funcs) do
+            local func = box.space._func.index.name:get(func_name)
+            ilt.assert(func ~= nil)
+            local priv = box.space._priv:get({user.id, 'function', func.id})
+            ilt.assert_equals(priv.privilege, box.priv.X)
+        end
+
+        -- Deploy the privileges for a role.
+        box.schema.role.create('two')
```
Gerold103, line 122: Please, drop the role and the user in the end of the test.

Gerold103 (commented):
Thanks for the patch!

test/unit-luatest/storage_exports_test.lua
ImeevMA: Thank you for the review! Fixed.

Gerold103 (approved):
Thanks for the fix!

PR #458, opened 2023-12-20
replicaset: soften name validation

Serpentian:
During upgrade to Tarantool 3.0.0 there's a time, when instance name have not been set yet. If vshard strictly validates names on connection, it leads to the cluster unavailability by the time, when all names are configured.

This commit softens the check and allows name to be nil, when instance UUID is specified in configuration, only UUID is validated in such case.

In conclusion we have the following checks on connection:

    1. UUID is validated in the following cases:
        a. When `identification_mode` is `uuid_as_key`
        b. When `identification_mode` is `name_as_key` and replica.uuid
           is specified

    2. Instance name is validated only when `identification_mode` is
       `name_as_key` and can work in the following modes:
        a. Strict validation (nil instance name is not allowed), when
           replica.uuid is not specified.
        b. Soft validation (nil instance name is allowed), when
           replica.uuid is specified.

Follow-up #426

NO_DOC=internal

### DIFF
--- a/test/replicaset-luatest/replicaset_3_test.lua
+++ b/test/replicaset-luatest/replicaset_3_test.lua
@@ -269,14 +269,24 @@ test_group.test_named_replicaset = function(g)
     t.assert_equals(rs.id, rs.name)
     t.assert_equals(replica_1_a.id, replica_1_a.name)
 
-    -- Name is not set, name mismatch error.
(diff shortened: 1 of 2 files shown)


Gerold103 (approved):
Thanks for the patch!

PR #459, opened 2023-12-22
ci: bump test-run

0x501D:
In the near future we plan to test Tarantool installed from a DEB/RPM package. Since 3.0.0-alpha1 release Tarantool packages don't have the tarantoolctl utility inside.

tarantoolctl was ebmed to the test-run project inself. The submodule has been updated to the latest version.

### DIFF
--- a/example/router.lua
+++ b/example/router.lua
@@ -17,7 +17,6 @@ cfg = dofile('localcfg.lua')
 if arg[1] == 'discovery_disable' then
     cfg.discovery_mode = 'off'
 end
-cfg.listen = 3300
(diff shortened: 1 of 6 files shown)


Gerold103:
Tests don't pass on 1.10.

0x501D:
> Tests don't pass on 1.10.

Fixed.

In addition to this, Release-master tests periodically fail.
But this problem also appears in the master branch and is beyond the scope of this PR.

Gerold103 (approved):
Thanks for the patch!

PR #461, opened 2023-12-25
Add rockspec/rock publishing workflow

ylobankov:


.github/workflows/publishing.yml
```diff
@@ -0,0 +1,53 @@
...
+          sed \
+            -e "s/branch = '.\+'/tag = '${{ github.ref_name }}'/g" \
+            -e "s/version = '.\+'/version = '${{ github.ref_name }}-1'/g" \
+            ${{ env.ROCK_NAME }}-scm-1.rockspec > \
+            ${{ env.ROCK_NAME }}-${{ github.ref_name }}-1.rockspec
+      - name: Push rockspec
+        uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: ${{ env.ROCK_NAME }}-${{ github.ref_name }}-1.rockspec
```
Totktonada, line 53: I would also publish `.all.rock` to speed up the installation on the user side. See https://github.com/tarantool/expirationd/blob/bcbe8c46ceca8d82dc5bb66b1d0a40deca03ad9d/.github/workflows/publish.yml#L43-L71 for example.

ylobankov: Makes sense, thank you!

ylobankov: Fixed.

Totktonada (approved):
Thank you! I see no problems.

Serpentian (approved):
Thank you for the workflow! No objections

Gerold103 (approved):
Thanks for the patch!

PR #462, opened 2023-12-26
replicaset: validate name in replicaset_locate_master

Serpentian:
Currently, the name validation is not used, when locate_master() is called. It makes an explicit call via the connection to obtain a future object in order to figure out, whether the node is a master.

We should not make any calls to a replica until the time, we definitely know, that its name and uuid was validated. Let's use replica_call instead of conn.call.

Follow-up #426

NO_DOC=bugfix

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -1060,7 +1060,7 @@ local function replicaset_locate_master(replicaset)
     local func = 'vshard.storage._call'
     local args = {'info'}
     local const_timeout = consts.MASTER_SEARCH_TIMEOUT
-    local ok, res, err, f
+    local ok, res, err
     local master = replicaset.master
     if master then
         local sync_opts = {timeout = const_timeout}
@@ -1091,17 +1091,20 @@ local function replicaset_locate_master(replicaset)
     local futures = {}
     local timeout = const_timeout
     local deadline = fiber_clock() + timeout
-    local async_opts = {is_async = true}
+    local async_opts = {is_async = true, timeout = timeout}
     local replicaset_id = replicaset.id
     for replica_id, replica in pairs(replicaset.replicas) do
-        local conn = replicaset_connect_to_replica(replicaset, replica)
-        if conn:is_connected() then
-            ok, f = pcall(conn.call, conn, func, args, async_opts)
-            if not ok then
-                last_err = lerror.make(f)
+        replicaset_connect_to_replica(replicaset, replica)
+        ok, err = replica:check_is_connected()
+        if ok then
+            ok, res, err = replica_call(replica, func, args, async_opts)
+            if not ok and err ~= nil then
+                last_err = err
             else
-                futures[replica_id] = f
+                futures[replica_id] = res
             end
+        elseif err ~= nil then
+            last_err = err
         end
     end
     local master_id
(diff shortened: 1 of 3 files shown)


vshard/replicaset.lua
```diff
@@ -1091,17 +1091,15 @@ local function replicaset_locate_master(replicaset)
...
-                last_err = lerror.make(f)
-            else
-                futures[replica_id] = f
-            end
+        replicaset_connect_to_replica(replicaset, replica)
+        ok, res, err = replica_call(replica, func, args, async_opts)
+        if not ok then
+            last_err = err
+        else
+            futures[replica_id] = res
```
Gerold103, line 1102: Please, do not send a request if the connection is not established. This function intentionally used to skip non-connected instances, hoping that at least among the connected ones a master would be found. You can still check it with `replica:is_connected()` to take into account the vconnect thing.

Gerold103 (changes requested):
Thanks for the patch!

vshard/replicaset.lua
Serpentian: Thank you for noticing, fixed!

Note, that I changed `replica:is_connected()` function, as otherwise it doesn't return error, only true or nil:

<details><summary>Current behavior of return values</summary>

```lua
tarantool> function a() return nil, 'error' end
---
...

tarantool> return true and true and a()
---
- null
...

tarantool> return true and true and nil, 'error'
---
- null
- error
...

```

</details>

vshard/replicaset.lua
```diff
@@ -1198,8 +1201,13 @@ replicaset_mt.__index = index
...
-            return replica.conn and replica.conn:is_connected() and
-                conn_vconnect_check_or_close(replica.conn)
+            if not replica.conn then
+                return false
+            end
+            if not replica.conn:is_connected() then
+                return false
+            end
+            return conn_vconnect_check_or_close(replica.conn)
         end,
```
Gerold103, line 1211: I get your point about returning an error for reporting, but typically functions called with `is_...` prefix should only return true or false. Any other behaviour is highly unexpected. If you want an error to be returned, I would suggest  to introduce a new function `check_is_connected()` which would work the way you need. And `is_connected()` should be left with true/false result.

Gerold103 (changes requested):
Thanks for the fix!

vshard/replicaset.lua
Serpentian: Sure, added the `check_is_connected`, left `is_connected` intact.

I also return err even when there's no `conn` at all. I suppose, it's ok to overwrite prev error even if it was from `vconnect`. Both these situations just mean, that the connection is not done yet. However, returning just `last_err` is not a good approach, IMO, it should be a map of errors.

Gerold103 (approved):
Thanks for the fixes!

vshard/replicaset.lua
Gerold103: > I also return err even when there's no conn at all

I specifically wanted to avoid it as it is not technically an error. Just nobody talked to this instance yet. But I am fine with the new behaviour. If somebody would complain about too many error logs, you can do another PR to drop this error.

> However, returning just last_err is not a good approach, IMO, it should be a map of errors.

Doesn't many too much sense, I think. The reason for one error was that people would fix those errors one by one if they have multiple. Many errors are likely to be independent anyway because would be about different instances. Didn't want to complicate the code. Again, if anybody asks, we can always change that.

PR #463, opened 2023-12-26
storage: fix schema log message

Serpentian:
storage: fix schema log message

exports_deply_funcs() used to log the core version as nil. The problem
is the fact, that table.concat was used for core version string, so the
complied exports always had nil instead of core_version.

table.concat doesn't work with non-array tables, as they have no defined
order. Let's explicitly define __tostring function for version and use
it in order to log core version.

Closes https://github.com/tarantool/vshard/issues/465

NO_DOC=bugfix

### DIFF
--- a/test/unit-luatest/storage_exports_test.lua
+++ b/test/unit-luatest/storage_exports_test.lua
@@ -135,3 +135,15 @@ test_group.test_deploy_privs = function(g)
         box.schema.role.drop('two')
     end)
 end
+
+test_group.test_core_version = function(g)
+    g.server:exec(function()
+        local exports = _G.ivexports.log[#_G.ivexports.log]
+        exports = _G.ivexports.compile(exports)
+
+        local version = _TARANTOOL
+        local pos = version:find('-g')
+        version = version:sub(1, pos - 1)
+        ilt.assert_equals(exports.core_version, version)
+    end)
+end
--- a/vshard/version.lua
+++ b/vshard/version.lua
@@ -75,6 +75,17 @@ local version_mt = {
     __le = function(l, r)
         return version_cmp(l, r) <= 0
     end,
+    __tostring = function(v)
+        local str = v.id_major .. '.' .. v.id_middle .. '.' .. v.id_minor
+        if v.rel_type then
+            str = str .. '-' .. v.rel_type
+            if v.rel_num ~= 0 then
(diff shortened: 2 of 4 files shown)


test/unit-luatest/storage_exports_test.lua
```diff
@@ -135,3 +135,16 @@ test_group.test_deploy_privs = function(g)
         box.schema.role.drop('two')
     end)
 end
+
+test_group.test_core_version = function(g)
+    g.server:exec(function()
+        local vexports = require('vshard.storage.exports')
```
Gerold103, line 141: You can use `ivexports`, it is defined for this suite in `test_group.before_all`.

test/unit-luatest/storage_exports_test.lua
```diff
@@ -135,3 +135,16 @@ test_group.test_deploy_privs = function(g)
         box.schema.role.drop('two')
```
Gerold103, line 135: In the commit message: `exports_deply_funcs` -> `exports_deploy_funcs`.

vshard/version.lua
```diff
@@ -75,6 +75,17 @@ local version_mt = {
...
+        local str = v.id_major .. '.' .. v.id_middle .. '.' .. v.id_minor
+        if v.rel_type then
+            str = str .. '-' .. v.rel_type
+            if v.rel_num ~= 0 then
+                str = str .. v.rel_num
+            end
+        end
+        str = str .. '-' .. v.id_commit
+        return str
+    end,
```
Gerold103, line 88: I suggest to add some unit tests for this function too.

Gerold103 (changes requested):
Thanks for the patch!

test/unit-luatest/storage_exports_test.lua
Serpentian: Sure, didn't notice it. Fixed

vshard/version.lua
Serpentian: Sure, added

test/unit-luatest/storage_exports_test.lua
Serpentian: Fixed

Gerold103 (approved):
Thanks for the patch!

PR #469, opened 2024-01-26
test: fix flaky failover/failover.test.lua

Serpentian:
Router pings replicas and if they cannot respond in
failover_ping_timeout, it detaches connection from this replica.
The test checks, that connection is not closed, when long response is
done, even if failover_ping_timeout is small.

However, when the connection is detached, it can be garbage collected,
which leads to 'Connection closed' error. Let's stop Lua GC for the
connection by creating reference to it.

Related to https://github.com/tarantool/tarantool/issues/9629

NO_DOC=test

### DIFF
--- a/test/failover/failover.test.lua
+++ b/test/failover/failover.test.lua
@@ -181,6 +181,9 @@ t = string.rep('a', 1024 * 1024 * 500)
 rs = vshard.router.route(31)
 while rs.master ~= rs.replica do fiber.sleep(0.01) end
 future = nil
+-- Create strong reference to prevent Lua GC from closing the
+-- connection. Waiting for tarantool/tarantool#9629 to be fixed.
+conn = rs.master.conn
 while not future do fiber.sleep(0.01) future = vshard.router.callrw(31, 'echo', {t}, {is_async = true}) end
 vshard.router.static.failover_fiber:wakeup()
 res, err = future:wait_result(5)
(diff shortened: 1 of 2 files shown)


test/failover/failover.test.lua
```diff
@@ -177,6 +177,7 @@ cfg.failover_ping_timeout = 0.0000001
 vshard.router.cfg(cfg)
 while not test_run:grep_log('router_1', 'Ping error from', 1000) do fiber.sleep(0.01) end
 
+collectgarbage("stop")
```
Gerold103, line 180: I suggest to write a comment here why is it needed. Also you might want to try to save the netbox connection into a variable. Then you would keep one strong reference and prevent the GC + you wouldn't have to pause the entire GC for everything else.

Gerold103, line 180: What do you mean as an error in this test? AFAIR, netbox won't delete itself while at least one request is in progress. Which means the "long" request should complete fine even if the connection has no refs outside of itself. No?

Gerold103 (commented):
Thanks for the patch!

test/failover/failover.test.lua
Serpentian: Added the comment, ref is ok too, it also fixes the flakiness. So the problem is definitely in Lua GC.

Serpentian: This is how it works, I don't know the reason, why the connection is closed by Lua GC. This is the bug, which should be fixed, created ticket for that: https://github.com/tarantool/tarantool/issues/9629

Gerold103 (approved):
Thanks for the fix!

PR #470, opened 2024-01-29
go-vshard-1: init golang vshard library

KaymeKaydex:
since the tarantool router is a stateless application, we did a small implementation of the router in golang, we did several performance tests that showed [positive results ](https://github.com/KaymeKaydex/go-vshard-router?tab=readme-ov-file#benchmarks)

in the case of the router, we took resources from the golang service and gave them to tarntool routers

the library allows you to avoid unnecessary network trips and speed up the route map update due to golang runtime
; now we use https://github.com/KaymeKaydex/go-vhsard-router but we want to get official support

we want to hear some feedback on this topic and review, now there are several non-critical todos, as well as an unfinished readme. Before finishing I would like to know your opinion and feedback

new topology scheme: 
Снимок экрана 2024-01-29 в 13 40 11


spatecon:
Seems legit

spatecon (approved):

KaymeKaydex:
Now we have managed to come to a repository officially supported by the tarantool community
https://github.com/tarantool/go-vshard-router

PR #471, opened 2024-03-14
test: re-enable `async_no_yield` test from `vconnect` test group

### LINKED ISSUE #456: replicaset: async call yields, when connection is not established
Associated ticket: https://github.com/tarantool/tarantool/issues/9489

Vshard reproducer. Add the following to `replicaset-luatest/replicaset_3_test.lua`:

```lua
test_group.test_connect = function()
    local _, rs = next(vreplicaset.buildall(global_cfg))
    rs.master.conn = nil
    local csw1 = fiber.self():csw()
    rs:connect_replica(rs.master)
    local csw2 = fiber.self():csw()
    -- This fails, csw2 = csw1 + 1
    t.assert_equals(csw1, csw2)
end

```

CuriousGeorgiy:
This patch re-enables the test to reflect the correct behavior after the fix for tarantool/tarantool#9489 bug was merged.

Closes #456

### DIFF
--- a/test/replicaset-luatest/vconnect_test.lua
+++ b/test/replicaset-luatest/vconnect_test.lua
@@ -3,6 +3,7 @@ local t = require('luatest')
 local vreplicaset = require('vshard.replicaset')
 local vtest = require('test.luatest_helpers.vtest')
 local vutil = require('vshard.util')
+local lversion = require('vshard.version')
 local verror = require('vshard.error')
 
 local small_timeout_opts = {timeout = 0.01}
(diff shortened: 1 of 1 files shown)


Serpentian:
We have deadlock here, either Tarantool's integration tests fails with vshard or testing on master fails in vshard. PR should be merged right before https://github.com/tarantool/tarantool/pull/9803#pullrequestreview-1935994200, backporting to 2.11 should be also done immediately.

The other option is to delete `test_group.test_async_no_yield` for now and restore it, when everything will be merged in master and 2.11. I like the second option more

test/replicaset-luatest/vconnect_test.lua
```diff
@@ -176,8 +176,12 @@ test_group.test_async_no_yield = function(g)
     local csw1 = fiber.self():csw()
     local ret, err = rs:callrw('get_uuid', {}, opts)
     local csw2 = fiber.self():csw()
-    -- Waiting for #456 to be fixed.
-    t.assert_equals(csw2, csw1 + 1)
+    if ivutil.version_is_at_least(2, 11, 0, nil, 0, 0) then
```
Serpentian, line 179: Please, include `ivutil`

CuriousGeorgiy: AFAIC, `ivutil` is globally available in tests. But this test already requires `vutil`, so I just changed `ivutil` to `vutil`.

Serpentian: Thank you

Serpentian (approved):
If we'll go this way, we'll have failing CI for some time either in vshard on in master. The alternative of removing this test completely until the time, everything is merged is better IMO. @Gerold103, what do you think?

Serpentian (approved):
Thank you for the patch! No objections

Gerold103 (approved):
Thanks for the patch!

PR #473, opened 2024-03-14
test: update `async_no_yield` test from `vconnect` test group

CuriousGeorgiy:
This patch brings an update to the test to reflect the correct behavior after tarantool/tarantool#9489 is fixed. The test against correct behavior is temporarily disabled until tarantol/tarantool#9803 is merged.

This patch needs to be merged before https://github.com/tarantool/tarantool/pull/9803.

Part of #456


Gerold103 (approved):
Thanks for the patch! Just make sure it is re-enabled back when you fix the stuff in core.

PR #478, opened 2024-07-19
failover: add failover priority replica option

georgiy-belyanin:
Add the `failover_priority` config option in the] `sharding.<replicaset_id>.replicas.<replica_id>` section.

Using the option you may specify the order failover will connect to the replicas. The lower the value, the higher the priority of the replica.

### DIFF
--- a/test/failover-luatest/failover_test.lua
+++ b/test/failover-luatest/failover_test.lua
@@ -1,0 +1,59 @@
+local t = require('luatest')
+local vtest = require('test.luatest_helpers.vtest')
+
+local g = t.group('failover')
+
+g.after_all(function(g)
+    g.cluster:stop()
+end)
+
+
+g.test_failover_priorities = function(g)
+    local cfg = {
+        sharding = {
+            replicaset_1 = {
+                replicas = {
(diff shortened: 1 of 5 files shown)


georgiy-belyanin:
Instead of the extra `failover` configuration section I've implemented an approach with the extra optional shading replica configuration property. 
```lua
return {
    sharding = {
        ['cbf06940-0790-498b-948d-042b62cf3d29'] = {
            replicas = {
                ['8a274925-a26d-47fc-9e1b-af88ce939412'] = {
                    name = 'storage_1_a',
                    master = true,
                    failover_priority = 1
                },
                ['3de2e3e1-9ebe-4d0d-abb1-26d301b84633'] = {
                    name = 'storage_1_b',
                    failover_priority = 2
                },
                ['001688c3-66f8-4a31-8e19-036c17d489c2'] = {
                    name = 'storage_1_c',
                    failover_priority = 3
                }
            },
        }
    }
}
```
In my opinion it's better not to manage with the duplicates of the replica sets and replicas UUIDs and to specify the failover priorities near the definition.

Also I haven't found the corresponding issue yet.

PR #479, opened 2024-07-31
test: ignore zero lsn when waiting for vclock in router2 test

drewdzzz:
In test `router/router2` we get vclock from one instance and wait for this vclock on another one. The problem is we get vclock with local lsn which is independent on each instance - the commit fixes this mistake.


Gerold103 (approved):
Thanks for the patch!

PR #544, opened 2025-04-18
Backport storage-alerts

### LINKED ISSUE #493: Vshard shows alerts for replica, which are not in the vshard's config
The problem here is that, if we connect any kind of replica to our shard, vshard will suppose, that it's part of the current shard, master will throw errors for this replicas:

```lua
-- Use example, add replica (let's suppose, this is non anon CDC)
tarantool> box.cfg{replication = {'storage:storage@127.0.0.1:3301', 'storage:storage@127.0.0.1:3302'}}

----

-- Now on `replica_1_a` we'll see:
unix/:./data/storage_1_a.control> vshard.storage.info()
<...>
  alerts:
  - ['UNREACHABLE_REPLICA', 'Replica 5891c9d7-a7f4-4fca-b9bc-07246c040141 isn''t active']
  - ['LOW_REDUNDANCY', 'Only one replica is active']
```

Replica should not alert about replicas, which are not in vshard's config.

mrForza:
Closes #493


test/luatest_helpers/asserts.lua
```diff
@@ -40,4 +40,10 @@ function asserts:wait_fullmesh(servers, wait_time)
     end)
 end
 
+function asserts:assert_server_no_alerts(server)
```
Gerold103, line 43: This entire file seems dead in vshard. I propose to
- Drop `test/luatest_helpers/asserts.lua`.
- Your new assertion better be added to `test/luatest_helpers/server.lua`. As a server's method. You even take the server as an argument here. So it makes sense to make it a server's method.

test/storage-luatest/persistent_names_test.lua
```diff
@@ -136,3 +136,32 @@ test_group.test_no_unreachable_replica_alert = function(g)
...
+    local instance_id = named_replica:instance_id()
+
+    named_replica:stop()
+    asserts:assert_server_no_alerts(g.replica_1_a)
+
+    named_replica:drop()
+    g.replica_1_a:exec(function(id)
+        box.space._cluster:delete(id)
+    end, {instance_id})
+end
```
Gerold103, line 167: Isn't this test identical to the ones added in `storage_1_test.lua`?

test/luatest_helpers/asserts.lua
```diff
@@ -46,4 +46,13 @@ function asserts:assert_server_no_alerts(server)
     end)
 end
 
+function asserts:info_assert_alert(alerts, alert_name)
```
Gerold103, line 49: You are not using anything in this `asserts` object. So it doesn't need to be a method of it. I suggest to either just duplicate this function in the needed tests, or put it into `vtest`. This is where we store all our vshard testing helpers.

Gerold103 (commented):
Oh, thank you for fixing that!

test/luatest_helpers/asserts.lua
mrForza: fixed

test/luatest_helpers/asserts.lua
mrForza: fixed

test/luatest_helpers/asserts.lua
Gerold103: Nope, doesn't seem fixed. I still see this file and its usage.

test/luatest_helpers/asserts.lua
Gerold103: Nope, it is still not in vtest.

PR #566, opened 2025-07-10
Cherry pick service nullification

Serpentian:


mrForza (approved):
Thank you for the patch! Everything looks good!

PR #595, opened 2025-09-03
router: forbid using cdata as bucket_id

### LINKED ISSUE #594: router.call by cdata bucket_id causes fullscan and memory leak
```lua
vshard.router.bucket_count() -- 1024
```

```lua
require("fun").length(pairs(vshard.router.internal.static_router.route_map)) -- 1024
```

```lua
vshard.router.callrw(1ULL, "assert", {"OK"}) -- OK
```

```Lua
require("fun").length(pairs(vshard.router.internal.static_router.route_map)) -- 1025
```

```lua
vshard.router.callrw(1ULL, "assert", {"OK"}) -- OK
```

```Lua
require("fun").length(pairs(vshard.router.internal.static_router.route_map)) -- 1026
```

Serpentian:
Using cdata (e.g. 1ULL) as bucket id was allowed in all requests of the router. But using it caused the growth of the `route_map` table on every request and full scan of the cluster in order to find, where the bucket is. This happened due to the fact, that cdata is compared by its pointer, not the value, and consequently, as there was no such pointer key in the `route_map`, it was added and scanned every time.

It was decided to prohibit such usage, from now on we expect users to do `tonumber(bucket_id)`, if it's needed. The motivation for such solution is the fact, that all functions except calls already prohibit using cdata as bucket_id, and calls should be consistent with it. Another point is the fact, that the behavior of the `map_callrw` with cdata as bucket ids is very untrivial.

The alternative solution of allowing usage of cdata in calls and manually converting values to number in `bucket_set`, `bucket_reset` and `bucket_resolved` was due to the above mentioned reasons.

Unfortunately, the behavior of `map_callrw` with array of cdata is still untrivial, as {10ULL, 20ULL} is interpreted as {1 = 10, 2 = 20}, and the request is executed on the replicaset with 1st and 2nd bucket.

Closes #594

NO_DOC=bugfix

### DIFF
--- a/test/router-luatest/map_callrw_test.lua
+++ b/test/router-luatest/map_callrw_test.lua
@@ -647,3 +647,41 @@ g.test_map_callrw_array_encoded_as_map = function(cg)
         end
     end, nil, bids)
 end
+
+--
+-- gh-594: using cdata as bucket_id should be prohibited, as it causes fullscan
+-- of the cluster and subsequent memory leak in the route_map.
+--
+g.test_map_callrw_with_cdata_bucket_id = function(cg)
+    cg.router:exec(function()
+        -- vshard.router._buckets_group is checked for the sake of unit
+        -- testing of the map_callrw() protection.
+        local msg = 'is not a number'
+        local _, err = ivshard.router._buckets_group({1ULL, 2ULL}, 1)
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router._buckets_group({1, 2ULL}, 1)
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router._buckets_group({1ULL, 2}, 1)
+        ilt.assert_str_contains(err.message, msg)
+
+        -- Map_callrw in general forbids using cdata.
+        _, err = ivshard.router.map_callrw('assert', {'a'},
+            {bucket_ids = {1, 2ULL}})
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router.map_callrw('assert', {'a'},
+            {bucket_ids = {[1] = 'a', [2ULL] = 'b'}})
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router.map_callrw('assert', {'a'},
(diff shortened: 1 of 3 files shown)


test/router-luatest/map_callrw_test.lua
```diff
@@ -647,3 +647,40 @@ g.test_map_callrw_array_encoded_as_map = function(cg)
...
+            {bucket_ids = {1, 2ULL}})
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router.map_callrw('assert', {'a'},
+            {bucket_ids = {[1] = 'a', [2ULL] = 'b'}})
+        ilt.assert_str_contains(err.message, msg)
+        _, err = ivshard.router.map_callrw('assert', {'a'},
+            {bucket_ids = {[1ULL] = 'a', [2ULL] = 'b'}})
+
+        local res
+        -- But there's one case, where it's not possible to forbid that,
```
Serpentian, line 678: If smb sees the solution for that problem, please, share the ideas)

olegrok: Or you can ban cdata at all because I don't know how you can handle correctly following case:
```
tarantool> {[1LL] = 10, [1LL] = 20, [1LL] = 30}
---
- 1: 20
  1: 30
  1: 10
...
```

Serpentian: Yeah, I don't really wanna see cdata at all in bucket_ids. One of the reason, is `map_callrw`, another one - the majority of methods already prohibits cdata as `bucket_id`, we just need to prohibit it for remainings.

>

```
tarantool> {[1LL] = 10, [1LL] = 20, [1LL] = 30}
---
- 1: 20
  1: 30
  1: 10
...
```

This case is already prohibited, there're no problems with prohibiting, when cdata is in key of the table. There're problems with the situation ,when it's a value. The array {10LL, 20LL} is actually {1 = 10LL, 2 = 20LL} and I cannot prohibit that, since passing cdata as values to functions is ok, but user probably means, that the `map_callrw` should be executed on 10th and 20th bucket, not on the first two of them

Issue #583, opened 2025-08-12
Router's failover causes spam of `VHANDSHAKE_NOT_COMPLETE` error in logs of the storage


Serpentian:
Prioritized replicas are constantly pinged by routers to check the health of the replica and failover if necessary. The prioritized replica is the one to which vshard.router.callro requests with the highest priority are sent. When a ping arrives to a replica from a router, the replica's replication lag is checked in relation to the master, and for this, the master is searched, if it's not found already. Every 30 seconds, connections to the masters are reset and re-established during the ping, this is used so that the replica has up-to-date information about where the master is now. Reconnection causes short-term VHANDSHAKE_NOT_COMPLETE errors, since the name check is not instantaneous and is performed with each connection, this leads to spam in logs every 30 seconds:

```
2025-08-11 17:48:28.660 [163967] main/136/vshard.conn_man/vshard.storage init.lua:3532 E> Error during master discovery for core-r01: {"code":40,"type":"ShardingError","replica":"core-r01-i01","message":"Handshake with core-r01-i01 have not been completed yet","name":"VHANDSHAKE_NOT_COMPLETE"}
2025-08-11 17:48:28.661 [163967] main/136/vshard.conn_man/vshard.storage init.lua:3532 E> Error during master discovery for core-r01: {"code":40,"type":"ShardingError","replica":"core-r01-i03","message":"Handshake with core-r01-i03 have not been completed yet","name":"VHANDSHAKE_NOT_COMPLETE"}
2025-08-11 17:48:28.662 [163967] main/136/vshard.conn_man/vshard.storage init.lua:3532 E> Error during master discovery for core-r01: {"code":40,"type":"ShardingError","replica":"core-r01-i03","message":"Handshake with core-r01-i03 have not been completed yet","name":"VHANDSHAKE_NOT_COMPLETE"}
2025-08-11 17:48:28.663 [163967] main/6718/10.192.164.11:13305 (net.box)/vshard.replicaset replicaset.lua:222 I> connected to 10.192.164.11:13305
2025-08-11 17:48:28.665 [163967] main/6719/10.192.163.101:13305 (net.box)/vshard.replicaset replicaset.lua:222 I> connected to 10.192.163.101:13305
2025-08-11 17:48:28.677 [163967] main/136/vshard.conn_man/vshard.replicaset replicaset.lua:1213 I> Found master for replicaset core-r01: core-r01-i03
[...]

Issue #578, opened 2025-08-08
Do not spam `VSHANDSHAKE_NOT_COMPLETE` in logs from internal services


Serpentian:
On Tarantool 3, as soon as connection is done the async call to the remote instance is performed in order to validate the name of the instance. Until the name is validated the `VHANDSHAKE_NOT_COMPLETE` is throws for every call.

Internal services may spam that error and they scare users even though this is expected and everything is all right. We should not print that message for internal services.

```
storage-0 2025-04-30 10:34:15.239 [7] main/163/vshard.conn_man/vshard.storage init.lua:3532 E> Error during master discovery for storage: {"code":40,"type":"ShardingError","replica":"storage-2","message":"Handshake with storage-2 have not been completed yet","name":"VHANDSHAKE_NOT_COMPLETE"}
```

PR #596, opened 2025-09-17
Remove doubled error logs in services and calls 

### LINKED ISSUE #588: Internal services should not spam the same errors in the logs on every iteration
A lot of internal vshard services make network requests. If the remote node is down or disabled, then the number of logs starts to grow very fast and the storage for that logs is not infinite. If the error code is the same, as on prev iteration, then the error message should not be printed:

```
{"code":77,"base_type":"ClientError","type":"ClientError","message":"connect, called on fd 31, aka <addr>: Connection refused","trace":[{"file":".\/tarantool\/src\/box\/lua\/net_box.c","line":2276}]}
Aug 12 13:51:09 test-fix.i[8556]: main/145/vshard.discovery._static_router/bin.test-fix.i.vshard.replicaset replicaset.lua:689 E> Exception during calling 'vshard.storage.buckets_discovery' on 'uri:3045(admin@addr)': connect, called on fd 19, aka <addr>: Connection refused
``` 
```
Aug 12 13:51:19 test-fix.i[8556]: main/148/shared_storage:sandbox_update_worker/bin.test-fix.i.vshard.replicaset replicaset.lua:689 E> Exception during calling 'vshard.storage.bucket_stat' on 'uri:3043(admin@<addr>)': {"reason":"storage is disabled explicitly","code":33,"type":"ShardingError","message":"Storage is disabled: storage is disabled explicitly","name":"STORAGE_IS_DISABLED"}
```

Router services: discovery, master_search, failover
Storage services: recover, rebalancer, GC, master_search

Serpentian:
Closes #588


vshard/error.lua
```diff
@@ -333,6 +333,14 @@ local function error_is_timeout(err)
            err.message == 'Timeout exceeded') or err.type == 'TimedOut'
 end
 
+local function errors_are_same(err1, err2)
+    if (err1 and not err2) or (not err1 and err2) then
+        return false
+    end
+    return err1.code == err2.code and err1.message == err2.message and
```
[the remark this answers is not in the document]
mrForza: I discovered one test where such kind of errors may appear - `router-luatest/router_2_2_test.lua`. In this test `unexpected eof while reading` ssl errors appear.

Logs:
```
2025-09-25 10:38:00.731 [24143] iproto ssl.c:359 E> SSL_read(16336), called on fd 47, aka unix/:(socket), peer of unix/:(socket): unexpected eof while reading {"type":"SSLError","trace":[{"f
ile":"./src/lib/core/ssl.c","line":359}]}
2025-09-25 10:38:00.732 [24143] main/112/main/vshard.replicaset replicaset.lua:306 I> disconnected from unix/:/tmp/t/001_router-luatest/replica_1_a.iproto
2025-09-25 10:38:00.732 [24143] main/112/main/vshard.replicaset replicaset.lua:306 I> disconnected from unix/:/tmp/t/001_router-luatest/replica_2_a.iproto
2025-09-25 10:38:00.732 [24143] main/176/lua/vshard.replicaset replicaset.lua:1493 I> Old replicaset and replica objects are outdated.
2025-09-25 10:38:00.732 [24143] main/112/main/vshard.storage init.lua:3078 I> Wakeup the rebalancer
2025-09-25 10:38:00.733 [24143] main/112/main/vshard.storage init.lua:3078 I> Wakeup the rebalancer
2025-09-25 10:38:00.746 [24143] iproto ssl.c:359 E> SSL_read(16310), called on fd 50, aka unix/:(socket), peer of unix/:(socket): unexpected eof while reading {"type":"SSLError","trace":[{"f
ile":"./src/lib/core/ssl.c","line":359}]}
2025-09-25 10:38:00.747 [24143] main/170/unix/:/tmp/t/001_router-luatest/replica_1_a.iproto (net.box) net_box.lua:442 E> fiber is cancelled {"type":"FiberIsCancelled","trace":[{"file":"built
in/box/net_box.lua","line":442}]}
2025-09-25 10:38:00.747 [24143] main/134/unix/:/tmp/t/001_router-luatest/replica_2_a.iproto (net.box) net_box.lua:442 E> fiber is cancelled {"type":"FiberIsCancelled","trace":[{"file":"built
in/box/net_box.lua","line":442}]}
2025-09-25 10:38:00.755 [24143] iproto ssl.c:359 E> SSL_read(16336), called on fd 48, aka unix/:(socket), peer of unix/:(socket): unexpected eof while reading {"type":"SSLError","trace":[{"f
ile":"./src/lib/core/ssl.c","line":359}]}
2025-09-25 10:38:00.755 [24143] iproto ssl.c:359 E> SSL_read(16310), called on fd 52, aka unix/:(socket), peer of unix/:(socket): unexpected eof while reading {"type":"SSLError","trace":[{"f
ile":"./src/lib/core/ssl.c","line":359}]}
```

These errors are not related to the router calling functions on the storages. But there is a non-zero probability that such errors can occur, for example, with `map_callrw` / `map_callro` / e.t.c

.github/workflows/test_catch_flaky.yml
```diff
@@ -0,0 +1,46 @@
+name: catch_flaky
+
+on:
+  pull_request
+
+jobs:
+  linux:
+    # We want to run on external PRs, but not on our own internal
+    # PRs as they'll be run by the push to the branch. The job is also not
+    # launched, if he 'notest' or 'noflaky' label is set.
```
Gerold103, line 10: `he` -> `the`.

Gerold103, line 10: I didn't get this. Will it be running for PRs in vshard repo automatically or not? What are internal and external PRs? And why should it not run on one of them? Don't we want to detect flaky tests for all PRs (except having these special labels)?

test/CMakeLists.txt
```diff
@@ -8,3 +8,11 @@ add_custom_target(test-force
         --builddir=${PROJECT_BINARY_DIR}
         --vardir=${PROJECT_BINARY_DIR}/test/var
         --force)
+
+add_custom_target(test-flaky
+    COMMAND ${CMAKE_COMMAND} -E env RANGE=$$RANGE
+        git diff --relative=test --name-only $$\{RANGE:-master..HEAD\}
+        -- ./*/*test.lua
+        | paste -sd' ' - | awk '{for (i=0\; i<32\; i++) print $$0}'
```
Gerold103, line 16: Why do we need this line? Wouldn't it work already work with xargs below? And why 32 times, while the commit message says 64?

vshard/error.lua
```diff
@@ -333,6 +333,14 @@ local function error_is_timeout(err)
            err.message == 'Timeout exceeded') or err.type == 'TimedOut'
 end
 
+local function errors_are_same(err1, err2)
+    if (err1 and not err2) or (not err1 and err2) then
```
Gerold103, line 337: `if (err1 ~= nil) ~= (err2 ~= nil) then ...` might be shorter.

vshard/replicaset.lua
```diff
@@ -697,10 +699,17 @@ local function replica_call(replica, func, args, opts)
            err.type == 'ClientError') then
             err = lerror.from_string(err.message) or err
         end
-        log.error("Exception during calling '%s' on '%s': %s", func, replica,
-                  err)
+        local log_level = lerror.are_same(err, replica.last_error) and
+                          'verbose' or 'error'
+        log[log_level]("Exception during calling '%s' on '%s': %s",
+                       func, replica, err)
+        replica.last_error = err
```
Gerold103, line 706: The problem here is that if the errors are of different types or contain the called function name or are different in any other way, and are being all spammed together, then there will still be spam, because `last_error` will not be the same. It would only work if there would be one error happening repeatedly with no other errors at all.

This also forces us to check and nullify the `last_error` on the success path, which adds one more `if`.

test/luatest_helpers/server.lua
```diff
@@ -343,6 +344,31 @@ function Server:grep_log(what, bytes, opts)
     return found
 end
 
+function Server:assert_log_exactly_once(what, opts)
```
Gerold103, line 347: The name "assert" usually means a one time thing in all the tests. But the implementation actually is waiting. I would suggest to find another name, which would reflect that more explicitly. Like `ensure` or `wait`.

vshard/service_info.lua
```diff
@@ -45,6 +46,15 @@ local function service_set_status_ok(service)
     service.error = ''
 end
 
+local function service_log_error_if_needed(service, level, err, ...)
+    local err_str = string.format(err, ...)
+    if not service.is_error_set then
+        level = service.error ~= err_str and level or 'verbose'
+        log[level](err_str)
+    end
+    return err_str
```
Gerold103, line 56: It is quite counter-intuitive that the error message is returned. Lets better just take it as a ready message and return nothing. Otherwise, if a caller would just want to log an error message received from somewhere else, then it would have to be escaped from any special characters to avoid `string.format()` doing anything crazy with it.

Or merge it with `set_status_error()` into one function.

vshard/service_info.lua
```diff
@@ -44,6 +44,7 @@ local function service_set_status_ok(service)
     service_set_status(service, 'ok')
     service_next_iter(service)
     service.error = ''
+    service.last_info = ''
```
Gerold103, line 47: This seems to be unused. What is it for? Can we still keep the info messages? They might tell not less than the error logs. If they are spamming, then the nop-messages (like when 0 is discovered, 0 is recovered, etc) could be dropped. And the others somehow aggregated, perhaps.

vshard/storage/init.lua
```diff
@@ -2825,6 +2845,7 @@ local function rebalancer_service_f(service)
     local module_version = M.module_version
     while module_version == M.module_version do
         service:next_iter()
+        M.rebalancer_ratelimit:flush()
```
Gerold103: The rebalancer can sleep for an hour. It means any errors suppressed during the last 30 seconds before the rebalancer goes into idle mode are not going to be flushed. Or would be flushed much much later.

Which actually might be a problem. The flush right now doesn't print the timestamps of the original logs when they started being suppressed, right? So here when the rebalancer wakes up an hour later and prints some errors being suppressed, we won't know that they happened an hour ago.

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,132 @@
...
+    end
+    local entry_type, code
+    if (type(limiter.custom_get_entry_signature) == 'function') then
+        entry_type, code = limiter.custom_get_entry_signature(new_entry)
+        if type(entry_type) ~= 'string' or type(code) ~= 'number' then
+            return false
+        end
+    else
+        entry_type, code = new_entry.type, new_entry.code
+    end
```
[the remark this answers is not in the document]
Serpentian: Yeah, that's good proposition, but I decided to completely drop the part, which fixes the https://github.com/tarantool/vshard/issues/578 and https://github.com/tarantool/vshard/issues/583, so that we can merge it faster

test/unit-luatest/log_ratelimit_test.lua
```diff
@@ -0,0 +1,118 @@
...
+end)
+
+test_group.after_all(function(g)
+    g.server:drop()
+end)
+
+test_group.test_get_entry_signature = function()
+    -- The custom entry signature is not set.
+    local limiter = vratelimit.new()
+    t.assert(limiter:can_log(nil))
```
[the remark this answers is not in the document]
Serpentian: Fixed

vshard/replicaset.lua
```diff
@@ -1688,6 +1692,16 @@ end
 -- Replica failover
 --------------------------------------------------------------------------------
 
+local function replica_service_flush_limiter(replica)
```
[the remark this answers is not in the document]
Serpentian: > Can you instead create a new service for this? Which would run with the flush period.

Yeah, it also solves the problem with the rebalancer. But this cannot be the service for `replicaset` module, since we also have the limiters inside router and storage, which has to be flushed. So, I decided to create a fiber right in the `log_ratelimit` module

test/replicaset-luatest/replicaset_3_test.lua
```diff
@@ -923,3 +923,32 @@ test_group.test_locate_master_not_hangs_on_futures = function(g)
...
+    router:wait_log_exactly_once(msg, {timeout = 0.5, on_yield = function()
+        local _,  rs = next(ivshard.router.internal.static_router.replicasets)
+        rs:callrw('echo', nil, {timeout = 0.05})
+    end})
+
+    t.helpers.retrying({}, function()
+        router:exec(function() _G.failover_wakeup() end)
+        t.assert(router:grep_log("Suppressed .* messages " ..
+                                 "from 'vshard.replica.replica_1_a'"))
+    end)
```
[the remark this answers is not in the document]
Serpentian: Done

vshard/router/init.lua
```diff
@@ -417,9 +424,11 @@ end
 local function discovery_f(router)
     local service = lservice_info.new('discovery')
     router.discovery_service = service
+    router.discovery_ratelimit = util.new_ratelimit_for_service('discovery')
     local ok, err = pcall(discovery_service_f, router, service)
     if router.discovery_service == service then
         router.discovery_service = nil
+        router.discovery_ratelimit = nil
     end
```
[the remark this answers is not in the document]
Serpentian: Good catch, I actually don't need it

vshard/storage/init.lua
Serpentian: That's a very strong point and the main motivation, why I introduced the background fiber for flushing the ratelimiters.

vshard/router/init.lua
```diff
@@ -1217,12 +1217,7 @@ local function master_search_set(router)
             end
         end
     else
-        -- Do not make users pay for what they do not use - when the search is
-        -- disabled for all replicasets, there should not be any fiber.
         log.info('Master auto search is disabled')
-        for _, replicaset in pairs(router.replicasets) do
```
Serpentian: This was useless, because the loop goes other the newly created replicasets, which don't have `master_search` service at all. We don't need that `worker:remove_service` at all

vshard/router/init.lua
```diff
@@ -415,12 +416,17 @@ local function discovery_service_f(router, service)
 end
 
 local function discovery_f(router)
-    local service = lservice_info.new('discovery')
+    local name = 'discovery'
+    local service = lservice_info.new()
```
Gerold103: Why did you drop the service name?

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,178 @@
...
+        module_version = MODULE_VERSION,
+        -- Fiber to flush the created ratelimiters.
+        flush_fiber = nil,
+        -- The table of all ratelimiters, which are flushed automatically.
+        limiters = {},
+    }
+else
+    -- Do not forget to apply the upgrades, when M is changed.
+    return M
+end
```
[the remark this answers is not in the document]
Serpentian: You're right, this won't work properly. Fixed now, applied the same approach we use in the storage.

And I wonder, how other reloads work, they don't even make `rawset`, just recreated on every reload, the code does nothing. And if there will appear some background services, hot-reload to the old version won't work:

https://github.com/tarantool/vshard/blob/fc86f4c51a7c910ac2ba17efd7ca0465a77fca6b/vshard/storage/ref.lua#L45-L66

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,178 @@
...
+
+--
+-- Returns true, if message can be printed in logs, false otherwise. Arguments:
+--     * new_entry - used for deduplication of the log entries. Structure: {
+--         type = <string>,
+--         code = <int>,
+--         name = <string>,
+--     }
+--
+local function ratelimit_can_log(limiter, new_entry)
```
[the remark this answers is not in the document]
Serpentian: I refactored the whole part, now there's strict separation between RO and RW functions, hope, you like it

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,178 @@
...
+    if M.flush_fiber == nil then
+        M.flush_fiber = util.reloadable_fiber_create(
+            'vshard.ratelimit_flush', M, 'ratelimit_flush_f')
+    end
+    return ratelimit
+end
+
+local function ratelimit_drop(name)
+    M.limiters[name] = nil
+end
```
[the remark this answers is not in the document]
Serpentian: > Do you think it would be a good idea to make the limiters table have weak refs? 

That's actually genius. Fixed all problems with stale limiters

vshard/router/init.lua
Serpentian: Because I'm careless, fixed)

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,202 @@
...
+        elseif ratelimit_can_log(limiter, entry) then
+            -- The entry doesn't exist in the limiter, print in requested level
+            -- and add it to the limiter, so that the consequent entries are
+            -- not printed until it's flushed.
+            ratelimit_add_entry(limiter, entry)
+            level = log_lvl
+        else
+            -- The entry should be suppressed, it's in the limiter already.
+            ratelimit_suppress_entry(limiter, entry)
+        end
```
Gerold103: The signature is potentially obtained here up to 3 times per log ("check entry" passes, "can log" doesn't, then "suppress" is called). It might be wiser to extract the signature first. And then pass it everywhere instead of the `entry`. So it is calculated only once.

Gerold103: If you do this, ping me again, I will have a look.

Gerold103 (approved):

vshard/log_ratelimit.lua
```diff
@@ -0,0 +1,187 @@
+--
```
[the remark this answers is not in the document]
Serpentian: Thank you very much for noticing, fixed

Gerold103 (approved):

PR #616, opened 2025-10-31
tla/storage: initial implementation [doubled buckets problem investigation]

Serpentian:
----

The PR is here in order to have a place, where I can show, how changes from RFC make the situation better and in order to investigate traces of doubled buckets

----

Run the tests with:

```sh
JAVA_OPTS="-DTLA-Library=../../src" tlc
    DoubledBucketsSmallTest.tla -config DoubledBucketsSmallTest.cfg -workers 6
```

TODO: refactor network reorder and drop to use storage states
TODO: ctest integration

### DIFF
--- a/proofs/tla/test/storage/DoubledBigTest.cfg
+++ b/proofs/tla/test/storage/DoubledBigTest.cfg
@@ -1,0 +1,28 @@
+INIT Init
+NEXT Next
+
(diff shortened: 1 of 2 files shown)


Gerold103:
Man, this is impressive 😨. Genius idea to use TLA for this 🔥!

Issue #531, opened 2025-03-12
Introduce workflow for checking the test flakiness


Serpentian:
The idea is to introduce a workflow, which will run every changed in the PR test huge number of times in a loop:

```
./test-run.py --force -j50 $(yes "$filename" | head -100)
```

`-j50` option is questionable, since it'll load the CPU to 100%. We can consider using `cpulimit`, but it doesn't work as well as real 100% CPU load in detecting the flakiness.

PR #617, opened 2025-11-04
ci: add flaky test catching workflow

LevKats:
Add new ci workflow running all files changed in the PR commits multiple
times in parallel by default. Can be disabled by setting the noflaky
PR label.

Co-authored-by: Nikita Zheleztsov <n.zheleztsov@proton.me>

Fixes https://github.com/tarantool/vshard/issues/531

NO_DOC=ci
NO_TEST=ci

### DIFF
--- a/.github/workflows/catch_flaky.yml
+++ b/.github/workflows/catch_flaky.yml
@@ -1,0 +1,88 @@
+name: catch_flaky
+
+on:
+  pull_request:
+    paths:
+      - 'test/*/*_test.lua'
+      - 'test/*/*.test.lua'
+
+jobs:
+  catch_flaky:
+    if: github.event_name == 'pull_request' &&
+        !contains(github.event.pull_request.labels.*.name, 'noflaky')
+
+    env:
+      TNT_RELEASE_PATH: /home/runner/tnt-release
+
+    runs-on: ubuntu-22.04
+    timeout-minutes: 180
+    steps:
+      - name: Create variables for Tarantool
+        run: |
+          branch=master
+          commit_hash=$(git ls-remote \
+                        https://github.com/tarantool/tarantool.git \
+                        --branch ${branch} | head -c 8)
+          echo "TNT_BRANCH=${branch}" >> $GITHUB_ENV
+          echo "VERSION_POSTFIX=-${commit_hash}" >> $GITHUB_ENV
+        shell: bash
+
+      - name: Cache tarantool build
+        id: cache-tnt-release
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.TNT_RELEASE_PATH }}
+          key: cache-tnt-release-master${{ env.VERSION_POSTFIX }}
+
+      - name: Clone tarantool
+        if: steps.cache-tnt-release.outputs.cache-hit != 'true'
+        uses: actions/checkout@v3
+        with:
+          repository: tarantool/tarantool
+          ref: ${{ env.TNT_BRANCH }}
+          path: tarantool
+          fetch-depth: 0
+          submodules: true
+
+      - name: Build tarantool
+        if: steps.cache-tnt-release.outputs.cache-hit != 'true'
+        run: |
+          sudo apt-get -y install build-essential cmake make zlib1g-dev \
+            libreadline-dev libncurses5-dev libssl-dev libunwind-dev \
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -8,3 +8,13 @@ add_custom_target(test-force
         --builddir=${PROJECT_BINARY_DIR}
         --vardir=${PROJECT_BINARY_DIR}/test/var
         --force)
+
+add_custom_target(test-flaky
+    COMMAND git diff --relative=test/ --name-only
+        $$\{RANGE:-master..HEAD\}
+        -- "./*/*test.lua" |
+        xargs -L 128 -r ${PROJECT_SOURCE_DIR}/test/test-run.py -j -1
```
Serpentian: The `test-run` runs for me in 1 thread, the same happens on CI, the chance of catching the flakiness, when executing the tests in one thread, is near 0. The following fixes that:

```diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 00ab983..7da69d4 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -13,7 +13,7 @@ add_custom_target(test-flaky
     COMMAND git diff --relative=test/ --name-only
         $$\{RANGE:-master..HEAD\}
         -- "./*/*test.lua" |
-        xargs -L 128 -r ${PROJECT_SOURCE_DIR}/test/test-run.py -j -1
+        xargs -L 128 -r ${PROJECT_SOURCE_DIR}/test/test-run.py
         --force --retries=0
         --repeat=$$\{N_TRIALS:-32\}
         --builddir=${PROJECT_BINARY_DIR}
```

The help from test-run:

```
  -j [JOBS], --jobs [JOBS]
                        Workers count.
                        0 means 2 x CPU count.
                        -1 means everything running consistently (single process).
```

Serpentian: All other workflows use `test-force`, which run the tests in the single process, but there, I suppose, the idea was to minimize flakiness, and I'd rather leave them as is. But the idea of the new workflow to maximize the flakiness and it's done by running the tests in all threads

.github/workflows/catch_flaky.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: pip3 install --user -r test-run/requirements.txt
+
+      - name: Setup tt
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | sudo bash
+          sudo apt install -y tt
+          tt version
+
+      - name: Setup luatest
+        run: tt rocks install luatest 1.2.1
```
Serpentian: Why do we install some specific `luatest` version and not the latest? The same comment in another workflow

.github/workflows/catch_flaky.yml
```diff
@@ -0,0 +1,51 @@
...
+    env:
+      TNT_RELEASE_PATH: /home/runner/tnt-release
+
+    runs-on: ubuntu-22.04
+    timeout-minutes: 180
+    steps:
+      - name: Install tarantool
+        uses: tarantool/setup-tarantool@v3
+        with:
+          tarantool-version: '3.5'
```
Serpentian: I'd consider using master here. Yeah, it's much slower, but then we won't have to bump the version every several months here.

Serpentian (commented):
Thank you a lot!

.github/workflows/catch_flaky.yml
```diff
@@ -0,0 +1,51 @@
+name: catch_flaky
+
+on:
+  pull_request:
+    paths:
+      - 'test/*/*.lua'
```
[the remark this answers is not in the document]
LevKats: I added more precise patterns here, but it shouldn't be a problem anyway, since `test-run` selects only the files with tests. The only case when `test-run` runs all the tests is when we call it without parameters, which should not happen because of `xargs -r`.

.github/workflows/catch_flaky.yml
LevKats: Fixed

Serpentian (approved):
Thank you! That's great, that we finally can catch flaky tests, in vshard they're very frequent.

P.S. @Gerold103, please, don't merge until we drop the last 2 commits, they're for testing the CI and should not be merged

PR #622, opened 2025-11-17
storage, router: allow calling info when disabled

### LINKED ISSUE #565: Allow vshard.storage.info() on disabled storage and expose disabled status
When a storage is disabled, any call to vshard.storage.info() currently returns a `storage disabled` error.

Disabling a replica for reads/writes is a normal operational scenario—for example, Cartridge disables a storage on OperationError or before the very first apply_config() at instance start, while the roles' logic is still inactive. At the same time, modules such as crud still need to read the vshard configuration in the disabled state but cannot because storage.info() is blocked.

Moreover, to determine whether a storage is enabled, a user has to use vshard.storage.internal.is_enabled or catch the error, which is inconvenient.

**Proposal**:

- Let vshard.storage.info() work even when the storage is disabled.

- Include a status = 'disabled' | 'enabled' field in the returned table (or provide a public helper such as vshard.storage.is_enabled()) so applications can detect the state without relying on internals.

kamenkremen:
A storage or router may be marked as disabled, which previously made it difficult to obtain status information from these components.

This patch allows calling `vshard.storage.info` and `vshard.router.info` even when the component is disabled. Both functions now also return a new boolean field `is_enabled` indicating whether the component is currently enabled.

Closes #565

@TarantoolBot document
Title: vshard: info functions behavior change

vshard.storage.info()
vshard.router.info()

These functions can now be called even when the storage or router is disabled.

Both functions now also return a new boolean field `is_enabled` indicating whether the component is currently enabled.

### DIFF
--- a/vshard/router/init.lua
+++ b/vshard/router/init.lua
@@ -1546,6 +1546,7 @@ local function router_info(router, opts)
                 router.discovery_service:info(),
         }
     end
+    state.is_enabled = router.is_enabled
     return state
 end
 
@@ -1661,7 +1662,7 @@ end
 -- Public API protection
 --------------------------------------------------------------------------------
 
-local function router_api_call_safe(func, router, ...)
+local function router_api_call_safe(func, router, _opts, ...)
     return func(router, ...)
 end
 
@@ -1669,13 +1670,14 @@ end
 -- Unsafe proxy is loaded with protections. But it is used rarely and only in
 -- the beginning of instance's lifetime.
 --
-local function router_api_call_unsafe(func, router, ...)
+local function router_api_call_unsafe(func, router, opts, ...)
     -- Router can be started on instance with unconfigured box.cfg.
     if not router.is_configured then
         local msg = 'router is not configured'
         return error(lerror.vshard(lerror.code.ROUTER_IS_DISABLED, msg))
     end
-    if not router.is_enabled then
+    local is_disabled_skip = opts and opts.is_disabled_skip
+    if not router.is_enabled and not is_disabled_skip then
         local msg = 'router is disabled explicitly'
         return error(lerror.vshard(lerror.code.ROUTER_IS_DISABLED, msg))
     end
@@ -1683,9 +1685,9 @@ local function router_api_call_unsafe(func, router, ...)
     return func(router, ...)
 end
 
-local function router_make_api(func)
+local function router_make_api(func, opts)
     return function(router, ...)
-        return router.api_call_cache(func, router, ...)
+        return router.api_call_cache(func, router, opts, ...)
     end
 end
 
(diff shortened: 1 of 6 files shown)


vshard/router/init.lua
```diff
@@ -1546,6 +1546,8 @@ local function router_info(router, opts)
                 router.discovery_service:info(),
         }
     end
+
```
Serpentian: Nit: new line is not necessary here, we don't really use them, unless it's essential to split the part of code. Same in another place

test/storage-luatest/storage_1_test.lua
```diff
@@ -633,3 +633,15 @@ test_group.test_alerts_for_named_replica = function(g)
...
 end
+
+test_group.test_info_disable_consistency = function(g)
+    g.replica_1_a:exec(function()
+        local res = vshard.storage.info()
+        ilt.assert(res.is_enabled)
+        vshard.storage.disable()
+        local pcall_res, storage_info = pcall(vshard.storage.info)
+        ilt.assert(pcall_res)
+        ilt.assert(not storage_info.is_enabled)
```
Serpentian: Nit: we have `assert_not` for that

vshard/router/init.lua
```diff
@@ -1661,31 +1662,32 @@ end
...
 --
-local function router_api_call_unsafe(func, router, ...)
+local function router_api_call_unsafe(func, router, opts, ...)
     -- Router can be started on instance with unconfigured box.cfg.
     if not router.is_configured then
         local msg = 'router is not configured'
         return error(lerror.vshard(lerror.code.ROUTER_IS_DISABLED, msg))
     end
-    if not router.is_enabled then
+    local is_disabled_skip = opts and opts.is_disabled_skip == true
```
Serpentian: `== true` in the end is not needed

Serpentian (approved):
Well done! This was fast, good work

Gerold103 (commented):
Thanks for the patch 💪! I particularly like the way you managed to split it into commits, quite easy to read.

vshard/router/init.lua
```diff
@@ -1661,15 +1661,16 @@ end
 -- Public API protection
 --------------------------------------------------------------------------------
 
-local function router_api_call_safe(func, router, ...)
-    return func(router, ...)
+local function router_api_call_safe(func, router, arg1, arg2, arg3, arg4, arg5)
+    return func(router, arg1, arg2, arg3, arg4, arg5)
 end
```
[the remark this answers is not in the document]
Gerold103: I will wait for your benches. If no degradation, then we can proceed. Otherwise will need to find a way to rework it both on storage and router.

Serpentian (approved):
Thank you for the patches! Clean work! Just one small nit, sorry, didn't notice it on prev iteration of review.

kamenkremen:
CI failed due to flaky tests in router_2_2_test. The fix is submitted in PR #625

PR #628, opened 2025-12-15
router: make full map_callrw with split args

### LINKED ISSUE #559: Full `map_callrw` with split args
## Motivation

Currently, `map_callrw` has two modes: full and partial. In case of full mode a user function is executed on all masters in the cluster, if partial mode is used, then the function is executed on masters of the replicasets, on which buckets from `bucket_ids` were found. In case of partial mode user can split arguments between replicasets, as it was stated in https://github.com/tarantool/vshard/issues/539, he can even add arguments, which will be passed to all storages, where partial `map_callrw` is executed. We can extend it even further by allowing to pass some arguments to ALL masters in the cluster, and split the remainings.

## API

Now `map_callrw` has two modes explicitly: `map_callrw(<func>, <args>, {mode = <'full'/'partial'>)`. By default the mode is `full`, if `opts.bucket_ids` were not passed, and `partial` otherwise. We'll have the following behavior:

1. `mode = nil/'full', bucket_ids = nil` - this is old full `map_callrw`, which is executed on every master in the cluster.
2. `mode = partial`, bucket_ids = nil` - error, `bucket_ids` must be in `opts` for partial `map_callrw`
3. `mode = 'full', bucket_ids = {1, 2, 3} - error, no split arguments were passed
4. `mode = 'full', bucket_ids = {[1] = ..., [2] = ..., [3] = ...}` - arguments in `bucket_ids` are passed only to the replicasets, where these buckets are found, `<args>` are passed to ALL masters in replicaset
5. `mode = nil/'partial', bucket_ids = {1, 2, 3}` - old partial map_reduce, only on repliasets with buckets 1, 2, 3
6. `mode = nil/'partial', bucket_ids = {[1] = ..., [2] = ..., [3] = ...}` - split arguments in partial `map_reduce`, `<args>` are passed only to replicasets with buckets 1, 2, 3

mrForza:
This patch introduces a new way of `map_callrw` execution by which we can pass some arguments to all storages and split buckets' arguments to those storages that have at least one bucket of `bucket_ids`. To achieve this we introduce a new string option - `mode` to `map_callrw` api.

@TarantoolBot document

Title: vshard: `mode` option for `router.map_callrw()`

This string option regulates on which storages the user function will be executed via `map_callrw`. Possible values:
1) `full` - the user function will be executed on all storages in
   cluster.
2) `partial` - the user function will be executed on certain sotrages on
   which buckets from `bucket_ids` were found.

After that changes `map_callrw` works in 4 different scenarios depending on `mode` and `bucket_ids` options:
1) `<mode = 'full', bucket_ids = nil>` - user function will be executed
   with `args` on all storages.
2) `<mode = 'full', bucket_ids = {[bid_1] = {b_arg_1}, ...}>` - storages
   that have at least one bucket of `bucket_ids` will execute user
   function with `args` and additional buckets' arguments. Other storages
   will execute user function only with `args`.
3) `<mode = 'partial', bucket_ids = {bid_1, ...}>` - user function will be
   executed with `args` on storages that have at least one bucket of
   `bucket_ids`.
4) `<mode = 'partial', bucket_id = {[bid_1] = {b_arg_1}, ...}>` - the same
   as the 3rd scenario but buckets' arguments (b_arg_1, ..., b_arg_n) will
   be added as additional arguments to user function.

Also now `map_callrw` ends with error in cases of `<mode = 'full', bucket_ids = {1, 2, ...}>` and `<mode = 'partial', bucket_ids = nil>`.

Closes tarantool/vshard#559

### DIFF
--- a/test/router-luatest/map_callrw_test.lua
+++ b/test/router-luatest/map_callrw_test.lua
@@ -691,3 +691,180 @@ g.test_map_callrw_with_cdata_bucket_id = function(cg)
         ilt.assert_not(err)
     end)
 end
+
+g.test_full_map_callrw_with_numeric_bucket_ids = function(cg)
+    local res = router_do_map(cg.router, {123}, {
+        mode = 'full',
+        timeout = vtest.wait_timeout,
+        bucket_ids = {1, 2, 3}
+    })
+    t.assert(res.err)
+    t.assert_equals(res.err.message, 'Router can\'t execute map_callrw ' ..
+                    'with \'full\' mode and numeric bucket_ids')
+    t.assert_not(res.err_id)
+    t.assert_not(res.val)
+end
+
+g.test_partial_map_callrw_with_nil_bucket_ids = function(cg)
+    local res = router_do_map(cg.router, {123}, {
+        mode = 'partial',
+        timeout = vtest.wait_timeout
+    })
+    t.assert(res.err)
+    t.assert_equals(res.err.message, 'Router can\'t execute map_callrw ' ..
+                    'with \'partial\' mode and nil bucket_ids')
+    t.assert_not(res.err_id)
+    t.assert_not(res.val)
+end
+
+local function make_do_map_tracking_bucket_ids(cg)
+    -- We override 'do_map' function on storages in order to check that
+    -- default arguments and bucket arguments were successfully passed into
+    -- destination storages according to mode and bucket_ids options.
+    vtest.cluster_exec_each_master(cg, function()
+        rawset(_G, 'old_do_map', _G.do_map)
+        rawset(_G, 'do_map', function(args, bucket_args)
+            ilt.assert_gt(require('vshard.storage.ref').count, 0)
+            return {ivutil.replicaset_uuid(),
+                    {args = args, b_args = bucket_args}}
+        end)
+    end)
+end
+
+local function reset_do_map_to_old_state(cg)
+    vtest.cluster_exec_each_master(cg, function()
+        rawset(_G, 'do_map', _G.old_do_map)
+    end)
+end
+
+g.test_full_map_callrw_with_split_args = function(cg)
+    make_do_map_tracking_bucket_ids(cg)
+    local bid1 = vtest.storage_first_bucket(g.replica_1_a)
+    local bid2 = vtest.storage_first_bucket(g.replica_2_a)
+    local bid3 = vtest.storage_first_bucket(g.replica_3_a)
+
+    local res = router_do_map(cg.router, {0}, {
+        mode = 'full',
+        timeout = vtest.wait_timeout,
+        bucket_ids = {[bid1] = {111}, [bid2] = {222}, [bid3] = {333}}
+    })
+    t.assert_not(res.err)
+    t.assert_not(res.err_id)
+    t.assert_equals(res.val, {
+        [cg.rs1_uuid] = {{cg.rs1_uuid, {args = 0, b_args = {[bid1] = {111}}}}},
+        [cg.rs2_uuid] = {{cg.rs2_uuid, {args = 0, b_args = {[bid2] = {222}}}}},
+        [cg.rs3_uuid] = {{cg.rs3_uuid, {args = 0, b_args = {[bid3] = {333}}}}},
+    })
+
+    res = router_do_map(cg.router, {0}, {
+        mode = 'full',
+        timeout = vtest.wait_timeout,
+        bucket_ids = {[bid2] = {222}}
+    })
+    t.assert_not(res.err)
+    t.assert_not(res.err_id)
+    t.assert_equals(res.val, {
+        [cg.rs1_uuid] = {{cg.rs1_uuid, {args = 0}}},
+        [cg.rs2_uuid] = {{cg.rs2_uuid, {args = 0, b_args = {[bid2] = {222}}}}},
+        [cg.rs3_uuid] = {{cg.rs3_uuid, {args = 0}}},
+    })
+    reset_do_map_to_old_state(cg)
+end
+
+g.test_full_map_callrw_without_bucket_ids = function(cg)
+    make_do_map_tracking_bucket_ids(cg)
+    local res = router_do_map(cg.router, {0}, {
+        mode = 'full',
+        timeout = vtest.wait_timeout
+    })
+    t.assert_not(res.err)
+    t.assert_not(res.err_id)
+    t.assert_equals(res.val, {
+        [cg.rs1_uuid] = {{cg.rs1_uuid, {args = 0}}},
+        [cg.rs2_uuid] = {{cg.rs2_uuid, {args = 0}}},
+        [cg.rs3_uuid] = {{cg.rs3_uuid, {args = 0}}},
+    })
+    reset_do_map_to_old_state(cg)
+end
+
+g.test_partial_map_callrw_with_numeric_bucket_ids = function(cg)
+    make_do_map_tracking_bucket_ids(cg)
+    local bid1 = vtest.storage_first_bucket(g.replica_1_a)
+    local res = router_do_map(cg.router, {0}, {
+        mode = 'partial',
+        timeout = vtest.wait_timeout,
+        bucket_ids = {bid1}
+    })
+    t.assert_not(res.err)
+    t.assert_not(res.err_id)
+    t.assert_equals(res.val, {[cg.rs1_uuid] = {{cg.rs1_uuid, {args = 0}}}})
+    reset_do_map_to_old_state(cg)
+end
+
+g.test_partial_map_callrw_with_split_args = function(cg)
+    make_do_map_tracking_bucket_ids(cg)
+    local bid1 = vtest.storage_first_bucket(g.replica_1_a)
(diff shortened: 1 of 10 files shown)


vshard/router/init.lua
```diff
@@ -978,15 +980,19 @@ local function replicasets_map_reduce(replicasets, rid, func, args,
...
+            local rs_grouped_args = grouped_args[id]
+            table.insert(args, rs_grouped_args)
+            grouped_args_added = rs_grouped_args ~= nil
         end
         local res, err = rs:callrw('vshard.storage._call', func_args, opts_map)
-        if grouped_args ~= nil then
+        if grouped_args_added then
             table.remove(args)
+            grouped_args_added = false
         end
```
Gerold103: You won't need this flag I think if you would change it to something as:
```Lua
for ... do
    local rs_args = grouped_args and grouped_args[id] or nil
    if rs_args then table.insert(args, grouped_args[id]) end
    -- do the call
    if rs_args then table.remove(args) end
end
```
The code style is a bit relaxed, but you get the idea.

vshard/router/init.lua
```diff
@@ -1083,14 +1089,46 @@ local function router_group_map_callrw_args(router, bucket_ids, bucket_args)
...
 
+--
+-- Set the appropriate mode according to bucket_ids option for backward
+-- compatibility (in case of opts_mode is nil) and check the given opts_mode
+-- correctness in other cases.
+--
+local function router_set_map_callrw_mode(opts_mode, bucket_ids)
+    if opts_mode == nil then
+        local mode = bucket_ids and MAP_CALLRW_PARTIAL or MAP_CALLRW_FULL
+        return mode, nil
```
Gerold103: You don't need to return trailing `nil`s in Lua. Can be simplified to `return bucket_ids and MAP_CALLRW_PARTIAL or MAP_CALLRW_FULL`.

vshard/router/init.lua
```diff
@@ -1137,7 +1177,12 @@ end
 local function router_map_callrw(router, func, args, opts)
     local replicasets_to_map, err, err_id, map, rid
     local timeout, do_return_raw, bucket_ids, plain_bucket_ids, grouped_args
+    local mode = MAP_CALLRW_FULL
     if opts then
+        mode, err = router_set_map_callrw_mode(opts.mode, opts.bucket_ids)
+        if err then
+            return nil, err, nil
```
Gerold103: Same here. No need for a trailing `nil`.

vshard/router/init.lua
```diff
@@ -1137,7 +1177,12 @@ end
...
+        end
         timeout = opts.timeout or consts.CALL_TIMEOUT_MIN
         do_return_raw = opts.return_raw
```
Gerold103: Here you assign `mode` and then overwrite it when `opts` are provided. Lets not do an overwrite. And instead simply declare the empty mode variable, and with given `opts` do what you do now, and in the `else` branch assign it to the "full" default. Then there won't be an overwrite, which in turn might save us a tiny bit of cpu.

vshard/router/init.lua
```diff
@@ -1083,14 +1089,46 @@ local function router_group_map_callrw_args(router, bucket_ids, bucket_args)
...
+--    2) <mode = 'full', bucket_ids = {[bid_1] = {b_arg_1}, ...}> - storages
+--       that have at least one bucket of bucket_ids will execute user function
+--       with args and additional buckets' arguments. Other storages will
+--       execute user function only with args.
+--    3) <mode = 'partial', bucket_ids = {bid_1, ...}> - user function will be
+--       executed with args on storages that have at least one bucket of
+--       bucket_ids.
+--    4) <mode = 'partial', bucket_id = {[bid_1] = {b_arg_1}, ...}> - the same
+--       as the 3rd scenario but buckets' arguments (b_arg_1, ..., b_arg_n)
+--       will be added as additional arguments to user function.
```
Gerold103: This here and in the commit message might be slightly misleading. It looks like the user should expect the bucket arguments to be passed as positional arguments like `func(all_args, bucket1_arg, bucket2_arg, ...)`. But instead it is more like `func(all_args, bucket_args_map)`. I suggest to clarify this.

Gerold103 (commented):
Thanks 🔥! I sure hope it will help the customers 😁.

Gerold103:
The commit message doesn't mention the ticket. You need to write `Closes #559` right before the TarantoolBot doc request.

vshard/router/init.lua
```diff
@@ -979,15 +981,12 @@ local function replicasets_map_reduce(replicasets, rid, func, args,
...
     for id, rs in pairs(replicasets) do
-        if grouped_args ~= nil then
-            -- It's cheaper to push and then pop, rather then deepcopy
-            -- arguments table for every call.
-            table.insert(args, grouped_args[id])
-        end
+        local rs_args = grouped_args and grouped_args[id] or nil
+        -- It's cheaper to push and then pop, rather then deepcopy
+        -- arguments table for every call.
+        if rs_args then table.insert(args, rs_args) end
```
Serpentian: Nit: let's preserve the old style with proper indentation of the conditions. It's easier to look at the code and minimizes the diff

vshard/router/init.lua
```diff
@@ -1136,33 +1171,37 @@ end
...
-        if timeout then
-            grouped_args = router_group_map_callrw_args(
-                router, plain_bucket_ids, bucket_ids)
-        end
     else
         timeout, err, err_id, rid, replicasets_to_map =
             router_ref_storage_all(router, timeout)
     end
     if timeout then
+        if plain_bucket_ids then
```
Serpentian: This won't work unfortunately. Previously the `router_group_map_callrw_args` was invoked right after `router_ref_storage_by_buckets`, now it can be invoked also after `router_ref_storage_all`. The first one finds the buckets in the cluster, updates the router map cache and only because of that we could use the `router_group_map_callrw_args` function. But the `router_ref_storage_all` doesn't do that.

Currently, the newly added mode (mode = 'full' + split args) doesn't work, when the cache on a router is broken, the arguments will be sent to the wrong instances. Let's won't forget to add the test case for that

Serpentian (commented):
Thank you for the patch! It looks great, but I'm afraid, it's a little bit broken now, let's fix and test the new mode with broken cache on the router

test/router-luatest/map_callrw_test.lua
```diff
@@ -685,3 +685,116 @@ g.test_map_callrw_with_cdata_bucket_id = function(cg)
         ilt.assert_not(err)
```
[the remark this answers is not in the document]
mrForza: fixed

vshard/router/init.lua
mrForza: fixed

test/router-luatest/map_callrw_test.lua
```diff
@@ -685,3 +685,164 @@ g.test_map_callrw_with_cdata_bucket_id = function(cg)
...
+                    'with \'partial\' mode and nil bucket_ids')
+    t.assert_not(res.err_id)
+    t.assert_not(res.val)
+end
+
+g.test_full_map_callrw_with_split_args = function(cg)
+    -- We override 'do_map' function on storages in order to check that
+    -- default arguments and bucket arguments were successfully passed into
+    -- destination storages according to mode and bucket_ids options.
+    vtest.cluster_exec_each_master(cg, function()
```
Serpentian: It's not guaranteed, that the tests will run in order, you defined: for example you can ran a single test case with `test-run`, and it won't work in that case, since the `do_map` is overridden in another test case only.

We're supposed to reset the cluster at the same state, as it was before the test. So, we have to overwrite the `do_map` function at the start of every test and drop it at the end of each one.

Serpentian (commented):
Thank you for the patchset. Very good! Seems like the last comments from me

test/router-luatest/map_callrw_test.lua
mrForza: fixed

Serpentian (approved):
Thank you a lot! Clean work, IMHO. Hope, it'll help TDG make requests safer

Serpentian (approved):

vshard/router/init.lua
```diff
@@ -1132,40 +1179,48 @@ end
...
-            })
-        if map then
-            return map
+    if mode == MAP_CALLRW_FULL then
+        timeout, err, err_id, replicasets_to_map =
+            router_ref_storage_all(router, timeout, replicasets_to_map, rid)
+        if not timeout then
+            goto fail
         end
     end
```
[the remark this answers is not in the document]
mrForza: No, in full map_callrw with split args we must invoke `router_ref_storage_by_buckets`, because without using this function the `router_group_map_callrw_args` will build incorrect table of form `{rs_id = {b_id = {b_args}}}` as the router's cache may be broken. `router_ref_storage_by_buckets` finds the buckets in the cluster, updates the router's cache - `router.route_map` and refs storages which relate to given buckets ids whereas the `router_ref_storage_all` doesn't do that.  See [Nikita's comment](https://github.com/tarantool/vshard/pull/628#discussion_r2643341859) related to this problem.

PR #629, opened 2025-12-18
storage: preserve original error on function fail

### LINKED ISSUE #614: `vshard.storage.call` masks the original error thrown from a function in some cases
Tarantool raises an error `Transaction is active at return from function` when a called over iproto function leaves an open transaction, but doesn't throw an error. This is the desired behavior, as it was discussed in tarantool/tarantool#7288. On the other hand, if the function raises an error, this error is returned to the user "as is".

`vshard.storage.call` calls the user function under a `pcall`, which means that if a user function raises an error, this error is not thrown out of `vshard.storage.call`. So if a user function throws an error before closing a transaction, the transaction will remain open and at the same time `vshard.storage.call` will catch this error, which will cause tarantool to raise `Transaction is active at return from function` instead.

Effectively, the original error will be hidden from the user and replaced with something else, without any traces of the original error location.

Example: apply the following diff to `storage.lua` in example cluster:
```diff
diff --git a/example/storage.lua b/example/storage.lua
index 05e5ef3..e76680e 100755
--- a/example/storage.lua
+++ b/example/storage.lua
@@ -119,9 +119,13 @@ function sleep(time)
 end
 
 function raise_luajit_error()
+    box.begin()
     assert(1 == 2)
+    box.commit()
 end
 
 function raise_client_error()
+    box.begin()
     box.error(box.error.UNKNOWN)
+    box.commit()
 end
```

Now try to call any of these functions from a router. The original error message will be replaced with `Transaction is active at return from function`:
```
unix/:./data/router_1.control> vshard.router.callrw(1, 'raise_client_error')
---
- null
- code: 30
  base_type: ClientError
  type: ClientError
  message: Transaction is active at return from function
  trace:
  - file: src/box/iproto.cc
    line: 2371
...
```

This is reproduced on vshard 0.1.37 and Tarantool 2.11.8, Tarantool 3.x isn't affected, at least I couldn't find a way to trigger the same error easily. This must be because of the way functions are called on 3.x compared to 2.x:
https://github.com/tarantool/vshard/blob/60ae7bbdf37dfaf8edae277c01728efaa190f2ec/vshard/storage/init.lua#L262-L287

kamenkremen:
For older Tarantool versions vshard uses func.call to execute functions.

If such a function raises an error during an uncommitted transaction, 'Transaction is active...' error is going to be raised, which will mask the original error.

This patch fixes that bug by checking the function call result. If the function returns an error and a transaction is still active, vshard now explicitly rolls back the transaction.

Fixes #614

NO_DOC=bugfix

### DIFF
--- a/test/storage-luatest/storage_1_test.lua
+++ b/test/storage-luatest/storage_1_test.lua
@@ -674,3 +674,28 @@ test_group.test_info_disable_consistency = function(g)
         ilt.assert(res.is_enabled)
     end, {global_cfg})
 end
+
+local function test_error_msg_is_preserved_template(g, add_to_schema)
+    g.replica_1_a:exec(function(add_to_schema)
+        rawset(_G, 'test_fail', function()
+            box.begin()
+            error('test_error')
+        end)
+        if add_to_schema then
+            box.schema.func.create('test_fail')
+        end
+        local status, err = ivshard.storage.call(1, 'read', 'test_fail', {})
+        ilt.assert_not(status)
+        ilt.assert_str_contains(err.message, 'test_error')
+        ilt.assert_not_str_contains(err.message, 'Transaction is active')
+        rawset(_G, 'test_fail', nil)
+        if add_to_schema then
+            box.schema.func.drop('test_fail')
+        end
+    end, {add_to_schema})
+end
+
+test_group.test_error_msg_is_preserved = function(g)
+    test_error_msg_is_preserved_template(g, false)
+    test_error_msg_is_preserved_template(g, true)
+end
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -261,6 +261,20 @@ else
 end
 
 --
+-- If the function is called directly, fails, and leaves an uncommitted
+-- transaction, the original error is replaced by the "Transaction is active"
+-- error. We manually check if the function returned an error. If so, we
+-- rollback the transaction to reveal the original error.
+--
+local function handle_results(status, ...)
+    if not status then
+        box.rollback()
+        return status, box.error.new(box.error.PROC_LUA, (...))
+    end
+    return status, ...
+end
+
+--
 -- Invoke a function on this instance. Arguments are unpacked into the function
 -- as arguments.
 -- The function returns pcall() as is, because is used from places where
@@ -284,7 +298,7 @@ local_call = function(func_name, args)
     if not func then
         return pcall(netbox_self_call, netbox_self, func_name, args)
     end
-    return pcall(func.call, func, args)
+    return handle_results(pcall(func.call, func, args))
 end
 
 end


vshard/storage/init.lua
```diff
@@ -260,6 +261,20 @@ else
     end
 end
 
+local function handle_results(status, ...)
```
Serpentian: Nit: that's add a comment, that this is copy-pasta from `net_box.lua`, so that we don't forget that in the future

test/storage-luatest/storage_1_test.lua
```diff
@@ -633,3 +633,20 @@ test_group.test_alerts_for_named_replica = function(g)
```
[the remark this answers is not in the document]
Serpentian: This sentence is still incorrect:

> For older Tarantool versions vshard uses func.call to execute functions registered in schema.

Let's properly write it. I'm asking you to do that, since any user or developer in the future may read this commit msg, let's not mislead them

test/storage-luatest/storage_1_test.lua
```diff
@@ -674,3 +674,56 @@ test_group.test_info_disable_consistency = function(g)
...
+end
+
+test_group.test_error_msg_is_preserved = function(g)
+    test_error_msg_is_preserved_template(false, g)
+    test_error_msg_is_preserved_template(true, g)
+end
+
+test_group.test_local_call_return_type_consistency = function(g)
+    g.replica_1_a:exec(function()
+        rawset(_G, 'test_return_tuple1', function()
```
Serpentian: And why don't we wanna use templates here? Two identical functions look strange

Serpentian (commented):
Almost there! Well done

Serpentian (approved):
Thank you for the patches! Well done

vshard/storage/init.lua
```diff
@@ -260,6 +261,23 @@ else
...
+    if not status then
+        box.rollback()
+        return status, box.error.new(box.error.PROC_LUA, (...))
+    end
+    local results = {...}
+    for i = 1, select('#', ...) do
+        if type(results[i]) == 'cdata' then
+            results[i] = msgpack.decode(msgpack.encode(results[i]))
+        end
+    end
```
Gerold103: Why do we even need that? Anyway the storage call is invoked remotely and it all will be converted into msgpack in order to be sent over the network. Also this looks like quite a performance hit. I think Crud people will come complaining, @Serpentian .

test/storage-luatest/storage_1_test.lua
```diff
@@ -674,3 +674,57 @@ test_group.test_info_disable_consistency = function(g)
...
+        rawset(_G, 'test_return_tuple', function()
+            return box.tuple.new({100, 200, 'tuple'})
+        end)
+        if add_to_schema then
+            box.schema.func.create('test_return_tuple')
+        end
+        local status, result = ivshard.storage.call(1,
+                                                      'read',
+                                                      'test_return_tuple',
+                                                      {})
```
Gerold103: The argument line wraps look malformed. Maybe make the function name shorter, or wrap all arguments to the next line like:
```

local function test_local_call_return_type_consistency_template(
    add_to_schema, g)
...
```

Same with the `ivshard.storage.call()`.

Serpentian:
Let's do only that variant:

> After only patch that preserves original error on function fail: 193125.2000 (99.7%)

The perf degradation is way too huge in the second variant. We'll leave https://github.com/tarantool/vshard/issues/630 unresolved for now, since nobody complained about that yet.

test/storage-luatest/storage_1_test.lua
```diff
@@ -674,3 +674,29 @@ test_group.test_info_disable_consistency = function(g)
...
+        rawset(_G, 'test_fail', function()
+            box.begin()
+            error('test_error')
+        end)
+        if add_to_schema then
+            box.schema.func.create('test_fail')
+        end
+        local status, err = ivshard.storage.call(1, 'read', 'test_fail', {})
+        ilt.assert_not(status)
+        local err_msg = tostring(err)
```
mrForza: We don't need to convert the whole error into a string, because it will contain an excess data: `code`, `base_type`, `type`, `trace`. It can slightly slow down the `assert_(not)_str_contains`. We need only the message of this error.

Lets simplify it:
```lua
ilt.assert_str_contains(err.message, 'test_error')
ilt.assert_not_str_contains(err.message, 'Transaction is active')
```

vshard/storage/init.lua
```diff
@@ -260,6 +260,20 @@ else
...
+--
+-- If the function is called directly, fails, and leaves an uncommitted
+-- transaction, the original error is replaced by the "Transaction is active"
+-- error. We manually check if the function returned an error. If so, we
+-- rollback the transaction to reveal the original error.
+--
+local function handle_results(status, ...)
+    if not status then
+        box.rollback()
+        return status, box.error.new(box.error.PROC_LUA, (...))
```
Gerold103: Wouldn't it replace error of any type with `ClientError`? What if it wasn't `ClientError`?

vshard/storage/init.lua
```diff
@@ -260,6 +260,20 @@ else
     end
 end
 
+--
+-- If the function is called directly, fails, and leaves an uncommitted
+-- transaction, the original error is replaced by the "Transaction is active"
+-- error. We manually check if the function returned an error. If so, we
+-- rollback the transaction to reveal the original error.
+--
+local function handle_results(status, ...)
```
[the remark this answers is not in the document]
kamenkremen: I did, results are in this [comment](https://github.com/tarantool/vshard/pull/629#issuecomment-3834089038). In short, here they are:
- Before patch: 193721.8000 (100%)
- After patch: 193125.2000 (99.7%)

PR #631, opened 2025-12-26
replicaset: log replica in connection triggers

Serpentian:
Before this patch, only the host and port were written to logs on connection or disconnection. However, it was very difficult to figure out to which replica we connected/disconnected, which made debugging the logs frustrating. Let's instead log the full replica object, which includes the replica id and its URI.

### DIFF
--- a/vshard/replicaset.lua
+++ b/vshard/replicaset.lua
@@ -270,8 +270,8 @@ end
 -- on_connect() trigger for net.box
 --
 local function netbox_on_connect(conn)
-    log.info("connected to %s:%s", conn.host, conn.port)
     local replica = conn.replica
+    log.info("connected to %s", replica)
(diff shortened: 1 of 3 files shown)


vshard/replicaset.lua
```diff
@@ -303,11 +303,12 @@ end
 -- on_disconnect() trigger for net.box
 --
 local function netbox_on_disconnect(conn)
-    log.info("disconnected from %s:%s", conn.host, conn.port)
-    assert(conn.replica)
+    local replica = conn.replica
+    log.info("disconnected from %s", replica)
```
Gerold103: URI is going to be printed without any passwords this way, right?

Gerold103 (commented):
Thanks! Nice, I hope it will help with debugging.

Issue #632, opened 2025-12-29
Connection is closed during name check on initial connection, when retryable error happens


Serpentian:
### Problem overview:

After cluster bootstrap router remains with alerts:

```
  alerts:
  - ['SUBOPTIMAL_REPLICA', 'A current read replica in replicaset storage-1 is not
      optimal']
  - ['SUBOPTIMAL_REPLICA', 'A current read replica in replicaset storage-2 is not
      optimal']
```

The connection on the router to prioritized replicas is manually closed (so `reconnect_after` doesn't work there), which means, that connection won't be restored automatically.

### Reproduce

I modified the functions in replicaset.lua as follows:

```lua
local function conn_vconnect_check_or_close(conn)
    local ok, err = conn_vconnect_check(conn)
    -- Close the connection, if error happened, but it is not
    -- VSHANDSHAKE_NOT_COMPLETE.
    if not ok and err and not (err.type == 'ShardingError' and
       err.code == lerror.code.VHANDSHAKE_NOT_COMPLETE) then
        log.warn('Closing the connection because of %s', err)
        conn:close()
    end
    return ok, err
end

<...>

local function conn_vconnect_wait_or_close(conn, timeout)
    local ok, err = conn_vconnect_wait(conn, timeout)
    if not ok and not lerror.is_timeout(err) then
        log.warn('Closing the connection because of %s (wait_or_close)', err)
        conn:close()
    end
    return ok, err
end
```

Then during bootstrap I see the following logs:

```
bus-output_market  | 2025-12-29 15:32:49.212 [7] main/132/vshard.failover._static_router/vshard.replicaset replicaset.lua:174 W> Closing the connection because of {"code":32,"base_type":"LuajitError","type":"LuajitError","message":"{\"reason\":\"storage is not configured\",\"code\":33,\"type\":\"ShardingError\",\"message\":\"Storage is disabled: storage is not configured\",\"name\":\"STORAGE_IS_DISABLED\"}","trace":[{"file":"./tarantool/src/lua/utils.c","line":700}]}
bus-output_market  | 2025-12-29 15:32:49.212 [7] main/126/bus-output_market:3313 (net.box)/vshard.replicaset replicaset.lua:265 I> disconnected from bus-output_market:3313
[...]

PR #638, opened 2026-01-28
replicaset: do not close initial conn during retryable error

mrForza:
Before this patch the replicaset closed an initial netbox connection
during name check (`vconnect` stage) when retryable error happened on the
storage side (`STORAGE_IS_DISABLED`, `NO_SUCH_PROC`, `AccessDeniedError`).
It was happening because `conn_vconnect_check_or_close` and
`conn_vconnect_wait_or_close` functions didn't have the logic with
retryable errors. This bug led to alerts on the router and as a result to
problems in high-level products which worked using vshard (e.g. TDG).

To fix this bug we introduce new replicaset helper functions -
`conn_vconnect_is_retryable_error` and `conn_vconnect_try_to_restart`.
They can help us to determine whether the error is retryable and restart
vconnect. We use it in `conn_vconnect_check_or_close` and
`conn_vconnect_wait_or_close`.

Also we delete one vconnect test - `test_vconnect_no_result` because it
tests the behavior which is changed by our new patch.

Closes tarantool/vshard#632

NO_DOC=bugfix

### DIFF
--- a/test/replicaset-luatest/vconnect_test.lua
+++ b/test/replicaset-luatest/vconnect_test.lua
@@ -71,33 +71,6 @@ test_group.test_vconnect_no_future = function(g)
 end
 
 --
--- Test, that conn_vconnect_check fails, when future's result is nil.
---
-test_group.test_vconnect_check_no_future = function(g)
-    local _, rs = next(vreplicaset.buildall(global_cfg))
-    g.replica:exec(function()
-        rawset(_G, '_call', ivshard.storage._call)
-        ivshard.storage._call = function() error('Non retryable error') end
-    end)
-
-    rs.master.conn = nil
-    local opts = table.deepcopy(timeout_opts)
-    opts.is_async = true
-    t.helpers.retrying({}, function()
-        -- It may be VHANDSHAKE_NOT_COMPLETE error, when future
-        -- is not ready. But at the end it must be the actual error.
-        local ret, err = rs:callrw('get_uuid', {}, opts)
-        t.assert_str_contains(err.message, "Non retryable error")
-        t.assert_equals(ret, nil)
-        t.assert_equals(rs.master.conn.state, 'closed')
-    end)
-
-    g.replica:exec(function()
-        ivshard.storage._call = _G._call
-    end)
-end
-
---
 -- 1. Change name and stop replica.
 -- 2. Wait for error_reconnect timeout.
 -- 3. Assert, that on reconnect name change is noticed.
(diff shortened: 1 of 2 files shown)


vshard/replicaset.lua
```diff
@@ -188,6 +188,41 @@ local function conn_vconnect_reset(conn)
     end
 end
 
+local function conn_vconnect_is_retryable_error(error)
+    -- ClientError is sent for all errors by old Tarantool versions which
```
Serpentian: After some thought, I don't think that duplicating code is good solution. I propose to create in `error` module `error.is_vshard_not_ready()` function and move the code from `can_backoff_after_error` there.  Then we'll use that new function in `can_backoff_after_error` and `conn_vconnect_is_retryable_error`

vshard/replicaset.lua
```diff
@@ -228,10 +263,14 @@ end
...
-    -- Close the connection, if error happened, but it is not
-    -- VSHANDSHAKE_NOT_COMPLETE.
+    if conn_vconnect_try_to_restart(conn, ok, err) then
+        return ok, err
+    end
+    -- Close the connection, if error happened, but it is not retryable
+    -- and not VSHANDSHAKE_NOT_COMPLETE.
     if not ok and err and not (err.type == 'ShardingError' and
        err.code == lerror.code.VHANDSHAKE_NOT_COMPLETE) then
+        log.warn('Closing the connection because of %s (check_or_close)', err)
```
Serpentian: To the separate commit, please. And let's test that, don't see tests about that right now

test/replicaset-luatest/vconnect_test.lua
```diff
@@ -109,7 +90,7 @@ test_group.test_vconnect_check_no_future = function(g)
         -- It may be VHANDSHAKE_NOT_COMPLETE error, when future
         -- is not ready. But at the end it must be the actual error.
         local ret, err = rs:callrw('get_uuid', {}, opts)
-        t.assert_str_contains(err.message, "_call' is not defined")
+        t.assert_str_contains(err.message, "PROC_LUA")
```
Serpentian: The error message is `Critical error`, type is `PROC_LUA`

vshard/replicaset.lua
```diff
@@ -270,7 +309,11 @@ end
 
 local function conn_vconnect_wait_or_close(conn, timeout)
     local ok, err = conn_vconnect_wait(conn, timeout)
+    if conn_vconnect_try_to_restart(conn, ok, err) then
+        return ok, err
+    end
     if not ok and not lerror.is_timeout(err) then
+        log.warn('Closing the connection because of %s (wait_or_close)', err)
```
Serpentian: Internals in log msg

mrForza: fixed

vshard/replicaset.lua
mrForza: fixed

vshard/replicaset.lua
```diff
@@ -242,6 +242,7 @@ local function conn_vconnect_check_or_close(conn)
            err.code == lerror.code.VHANDSHAKE_NOT_COMPLETE) then
         -- Close the connection, if error happened, but it is not retryable
         -- and not VSHANDSHAKE_NOT_COMPLETE.
+        log.warn('Closing the connection because of %s', err)
```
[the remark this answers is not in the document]
mrForza: fixed

vshard/replicaset.lua
```diff
@@ -228,10 +235,16 @@ end
 
 local function conn_vconnect_check_or_close(conn)
     local ok, err = conn_vconnect_check(conn)
-    -- Close the connection, if error happened, but it is not
-    -- VSHANDSHAKE_NOT_COMPLETE.
-    if not ok and err and not (err.type == 'ShardingError' and
-       err.code == lerror.code.VHANDSHAKE_NOT_COMPLETE) then
```
Gerold103: What was a problem to let it close the connection and let it reconnect and retry automatically? Or was it closed forever? I think the replicaset object re-creates the connections when we try to call them and they are noticed to be closed. No?

If that isn't so, then I would suggest we simply allow reconnects regardless of error code. This would be super simple, and even makes sense, because this thing exists for name checks and name can be changed. And then we would need to reconnect somehow anyway.

It looks shaky that stuff falls unless we cherry-pick some magic set of error codes which are retryable. (I also don't like that in applier, but this is a different story.)

Gerold103 (commented):
Thanks! Nicely sliced patches by the way ✍️.

I think I have just one question here from the start.

PR #649, opened 2026-05-13
Doubled buckets. Part 0 (rebalancer batching + ref inconsistencies during master switch)

### LINKED ISSUE #351: Rebalancer should firstly try to pick buckets which already have no refs
`rebalancer_worker_f()` should prefer to select buckets which have no refs. Their sending could be done without waiting (not counting competition with `map_callrw`).

### LINKED ISSUE #526: flaky test: storage-luatest/service_info_test
```

### LINKED ISSUE #528: flaky test: storage-luatest/auto_master_2_2_2_test
```
[013] storage-luatest/auto_master_2_2_2_test.lua                      [ fail ]
[013] Test failed! Output from reject file /tmp/t/rejects/storage-luatest/auto_master_2_2_2.reject:
[013] Tarantool version is 3.4.0-entrypoint-156-gcdd8bcfbc7
[013] TAP version 13
[013] 1..14
[013] # Started on Wed Mar 12 17:07:58 2025
[013] # Starting group: storage.is_sync:false
[013] ok     1	storage.is_sync:false.test_bootstrap
[013] ok     2	storage.is_sync:false.test_locate
[013] ok     3	storage.is_sync:false.test_notice_change
[013] ok     4	storage.is_sync:false.test_recovery
[013] ok     5	storage.is_sync:false.test_noactivity_timeout_for_auto_master
[013] ok     6	storage.is_sync:false.test_conn_manager_connect_self
[013] ok     7	storage.is_sync:false.test_master_discovery_on_disconnect
[013] # Starting group: storage.is_sync:true
[013] ok     8	storage.is_sync:true.test_bootstrap
[013] ok     9	storage.is_sync:true.test_locate
[013] ok     10	storage.is_sync:true.test_notice_change
[013] ok     11	storage.is_sync:true.test_recovery
[013] ok     12	storage.is_sync:true.test_noactivity_timeout_for_auto_master
[013] ok     13	storage.is_sync:true.test_conn_manager_connect_self
[013] not ok 14	storage.is_sync:true.test_master_discovery_on_disconnect
[013] #   Instance with replica id 2 was promoted first
[013] #   stack traceback:
[013] #   	...shard-ee/test/storage-luatest/auto_master_2_2_2_test.lua:87: in function 'promote_if_needed'
[013] #   	...shard-ee/test/storage-luatest/auto_master_2_2_2_test.lua:415: in function 'storage.is_sync:true.test_master_discovery_on_disconnect'
[013] #   	...
[013] #   	[C]: in function 'xpcall'
[013] #   artifacts:
[013] #   	replica_1_a -> /tmp/t/013_storage-luatest/artifacts/oCyftHygJXbV
[013] #   	replica_2_b -> /tmp/t/013_storage-luatest/artifacts/C60whb4CaJaA
[013] #   	replica_2_a -> /tmp/t/013_storage-luatest/artifacts/nkDUepZ196my
[013] # Ran 14 tests in 41.409 seconds, 13 succeeded, 1 errored
[013] 
```

### LINKED ISSUE #573: RW refs break replication when master changes
Rebalancing was started, after some time the replication got broken with the following error:

Image

I suppose this happened as follows: we had a master, which processed some RW requests, after that the master was changed (which is pretty casual thing, when rebalancing is in progress due to high load on storages), then a new master started sending bucket, which had RW ref on it, this broke the replication

### PR DESCRIPTION
**Pull request by Serpentian**
Closes #573
Closes #351
Closes #528
Closes #526

### CODE REVIEW

**Comment by Serpentian** [comment]
* Fixed not building the rocks bundle properly due to missing `route_dispenser.lua` in the `CMake`. Found in the integration testing.
* Fixed flaky `auto_master_2_2_2`, which was found in the `catch_flaky` workflow

File: test/rebalancer/receiving_bucket.result [hunk 1]
```diff
@@ -228,7 +228,7 @@ vshard.storage.internal.errinj.ERRINJ_LAST_RECEIVE_DELAY = true
 _ = test_run:switch('box_2_a')
 ---
 ...
-_, err = vshard.storage.bucket_send(101, util.replicasets[1], {timeout = 0.1})
+_, err = vshard.storage.bucket_send(101, util.replicasets[1], {timeout = 1})
```
Gerold103: My only comment is that it seems not right that we use one timeout for sync + data send. Sync is entirely bottlenecked by the replication and doesn't depend on bucket size. OTOH, the data send can be quite longer, and it does depend on the bucket size.

It seems logical and convenient, if we would add a new option `sync_timeout`. It would be used to wait for the sync. And the `timeout` would be used for sending.

Alternatively, to make it more explicit, we could make 3 options `sync_timeout` + `send_timeout` + `timeout`.
- If `sync_timeout` and `timeout` are specified, we calculate `send_timeout = timeout - sync_timeout`.
- If `send_timeout` and `timeout` are specified, we calculate `sync_timeout = timeout - send_timeout`.
- If `timeout` is not specified, we take it as default, same as now. Then we apply the rules above.
- If all 3 are specified, then we ignore `timeout`. Or raise an error.

This will help us to make the sync timeout in tests very high, and send - very small. And the tests can remain fast.

Serpentian: Reworked timeouts, introduced new config option: [storage: check for refs on replicas before sending a bucket](https://github.com/tarantool/vshard/pull/649/changes/7f3081739819784bcc152c9fe172fd324341ca36)

[storage: check for refs on replicas before sending a bucket](https://github.com/tarantool/vshard/pull/649/changes/7f3081739819784bcc152c9fe172fd324341ca36) - dropped the error, I used it in `bucket_test_send_on_replicas` to say, that `could not sync with replicas for readonly batch`, now I just set `err.reason` for that.

vshard/storage/init.lua
```diff
@@ -1986,6 +2012,11 @@ local function bucket_send(bucket_id, destination, opts)
     if type(bucket_id) ~= 'number' or type(destination) ~= 'string' then
         error('Usage: bucket_send(bucket_id, destination)')
     end
+    opts = bucket_send_build_opts(table.deepcopy(opts), {
+        ['timeout'] = consts.TIMEOUT_INFINITY,
+        ['chunk_timeout'] = consts.DEFAULT_BUCKET_CHUNK_TIMEOUT,
+        ['ref_timeout'] = consts.DEFAULT_BUCKET_REF_TIMEOUT,
+    })
```
Gerold103: I suggest to store the defaults outside of this function - the table is always the same, we don't need to create it on each call.

We also don't need `deepcopy` here, just `copy` is enough, since we only change root fields of `opts`.

vshard/storage/init.lua
```diff
@@ -2607,10 +2638,15 @@ local function rebalancer_prepare_buckets(bucket_count, opts)
...
 
-local function rebalancer_dispenser_pop(dispenser, opts)
-    assert(opts.timeout)
+local function rebalancer_dispenser_pop(dispenser)
     local ok, err, count, buckets
-    local deadline = fiber_clock() + opts.timeout
+    local opts = bucket_send_build_opts(nil, {
+        ['timeout'] = M.rebalancer_bucket_send_timeout,
+        ['ref_timeout'] = consts.DEFAULT_BUCKET_REF_TIMEOUT,
+        ['pop_timeout'] = consts.DEFAULT_BUCKET_POP_TIMEOUT,
```
Gerold103: What you call here `pop` is rather `sync`. Pop is what we do, but sync is what we actually want on the high level, and what is being limited here. Pop sounds too low-level and specific to dispenser.

vshard/storage/init.lua
```diff
@@ -2660,11 +2695,16 @@ end
...
-    local opts = {timeout = consts.REBALANCER_CHUNK_TIMEOUT}
-    local id, bid = rebalancer_dispenser_pop(dispenser, opts)
+    local id, bid = rebalancer_dispenser_pop(dispenser)
     local worker_throttle_count = 0
     while id and bid do
-        local status, ok, err = pcall(bucket_send_xc, bid, id, opts)
+        -- The opts are recreated for every bucket to allow reconfiguring
+        -- of the timeout when routes applier already works.
+        local send_opts = bucket_send_build_opts(nil, {
+            ['timeout'] = M.rebalancer_bucket_send_timeout,
```
Gerold103: It might look simpler if you write the options as `timeout = ...` instead of `['timeout'] = ...`. The latter syntax usually is only needed when the name contains non-ID chars like `-` or ` `.

vshard/storage/init.lua
```diff
@@ -1819,14 +1822,35 @@ local function buckets_discovery(opts)
...
+-- Builds the timeout options for sending a bucket. Used in manual
+-- `bucket_send` and rebalancer routes applier worker.
+--
+local function bucket_send_build_opts(opts, defaults)
+    assert(defaults['timeout'])
+    opts = opts or {}
+    local type = 'positive number'
+    for name, default in pairs(defaults) do
+        if opts[name] then
+            lcfg.check_option(name, type, nil, opts[name])
```
Gerold103: Nice, good idea to reuse the `cfg` here? Perhaps lets save `lcfg.check_option` as a `local cfg_check_option = lcfg.check_option` in the beginning of the file? Since it is used multiple times per bucket send.

vshard/storage/init.lua
```diff
@@ -1819,14 +1822,35 @@ local function buckets_discovery(opts)
...
 end
 
+--
+-- Builds the timeout options for sending a bucket. Used in manual
+-- `bucket_send` and rebalancer routes applier worker.
+--
+local function bucket_send_build_opts(opts, defaults)
+    assert(defaults['timeout'])
+    opts = opts or {}
+    local type = 'positive number'
```
Gerold103: I would recommend considering allowing here any number. Even negative. Otherwise it prevents writing linear code when the timeout was given to you from some other place, and you have to check it for `> 0` every time + raise an error manually. An example you can see in `util.fiber_cond_wait_xc()` and `util.future_wait_xc()`. Netbox future and fiber cond require timeout >= 0, and this gets quite annoying, because sometimes we just want to let it raise timeout error when they get it < 0.

vshard/storage/init.lua
Gerold103: If the third param is always `nil`, then maybe make `lcfg.check_option` just not take it, and only have 3 params?

vshard/storage/init.lua
```diff
@@ -2644,6 +2712,7 @@ local function rebalancer_dispenser_pop(dispenser)
         ['timeout'] = M.rebalancer_bucket_send_timeout,
         ['ref_timeout'] = consts.DEFAULT_BUCKET_REF_TIMEOUT,
         ['pop_timeout'] = consts.DEFAULT_BUCKET_POP_TIMEOUT,
+        ['sync_timeout'] = consts.DEFAULT_BUCKET_SYNC_TIMEOUT,
```
Gerold103: So what I am proposing is that `pop_timeout` is removed. Only `sync_timeout` remains, because it basically what pop does - either takes a ready bucket, or waits for a sync.

Gerold103 (commented):
Oh yes, these timeouts are beautiful ❤️.

vshard/storage/init.lua
Serpentian: Yeah, we can do that for `bucket_send`. Other ones - no go, since they're not static and depend on `M`

vshard/storage/init.lua
Serpentian: Ditto

vshard/storage/init.lua
Serpentian: Yeah, we can do that, done

vshard/storage/init.lua
Serpentian: Sure, done

vshard/storage/init.lua
Serpentian: Agree, and using negatives finally stabilizes the timeout test. Using even tiny positive can lead to success

vshard/storage/init.lua
Serpentian: Let's leave it as general as possible. If we'll want to use that function in the future for enums, we'll need that 3rd option. It's in the middle to be consistent with internal code of `cfg` module

vshard/storage/init.lua
Serpentian: Okaaay, we can remove it from opts. But we still need it and I don't think using the `sync_timeout` in pop is a good idea. Now it's dynamic: `M.rebalancer_worker_count *  opts.ref_timeout + opts.sync_timeout`

Gerold103: As we discussed on a call, I suppose we agree that this timeout won't be needed. One worker will do the preparation using refs and syncs, other workers will wait on it with infinity timeout. If the prepare-worker fails to prepare, then we abort the whole thing and cancel the fibers.

vshard/storage/init.lua
```diff
@@ -2558,7 +2590,7 @@ end
 local function rebalancer_worker_f(worker_id, dispenser, quit_cond)
     lfiber.name(string.format('vshard.rebalancer_worker_%d', worker_id))
     local _status = box.space._bucket.index.status
-    local opts = {timeout = consts.REBALANCER_CHUNK_TIMEOUT}
+    local opts = bucket_send_build_opts(nil, M.rebalancer_bucket_send_timeout)
```
Gerold103: You build here the opts for internals of bucket-send. And then call `bucket_send()` on these opts, in a loop. Seems like not much point in that, because `bucket_send()` will do `bucket_send_build_opts()` internally again anyway. Also this `opts` created before the loop will have one deadline set for the whole loop, for many buckets, while we probably intended to have different deadlines for each bucket, no?

You probably intended here to do `local opts = {timeout = M.rebalancer_bucket_send_timeout}`.

This is in `storage: rework timeouts for bucket sending` commit.

vshard/storage/init.lua
```diff
@@ -2580,6 +2584,94 @@ local function rebalancer_build_routes(replicasets)
...
+            goto prepare
+        elseif ok == nil then
+            -- Stop waiting, doesn't make sense anymore. Do not log error,
+            -- since it will either be logged later or is not needed to be
+            -- logged at all. Causes exit from a worker.
+            return
+        end
+        -- We're waken up, however, the buckets are not yet prepared
+        -- (e.g. all other workers are already sending them). We need
+        -- to wait again or prepare them on our own.
```
Gerold103, line 2749: Do you think it might make sense to reset the deadline here? Because this situation definitely means we are sending stuff ok, and perhaps even very fast.

vshard/storage/init.lua
```diff
@@ -2589,37 +2681,28 @@ end
...
-        end
-        if not is_found then
-            log.error('Can not find active buckets')
-            break
-        end
-        local ret, err = bucket_send(bucket_id, id, opts)
-        if ret then
+    while id and bid do
+        assert(dispenser.send_buckets_deadline)
+        opts.deadline = dispenser.send_buckets_deadline
```
Gerold103: Since you are manually changing the deadline here, then maybe it would be a good idea to recreate the opts inside the loop from scratch? Like `local opts = build....(dispenser.deadline - fiber_clock())`. Right now this works, but might get edgy if in the future the opts will have more than one 'deadline' attribute, and we will forget to update it here.

Hm. Or another idea - how about we make `rebalancer_dispenser_pop()` return the `opts` with already proper deadline? Then 1) it becomes harder to misuse it the dispenser and we won't need to touch its internal deadline member, 2) the dispenser will be actually using `send_buckets_deadline` in `route_dispenser.lua`, and not just store it like now.

vshard/storage/init.lua
Serpentian: > Seems like not much point in that, because bucket_send() will do bucket_send_build_opts() internally again anyway

Yeah, in this commit it doesn't make much sense, double call of the same function. The deadline was not general for the bunch of buckets, since we've overwritten the deadline on every call of the `bucket_send_build_opts`. It's not obvious, so I changed it as follows:

<details><summary>Details</summary>
<p>

```diff
diff --git a/vshard/storage/init.lua b/vshard/storage/init.lua
index f9ff6ff..8dfd37d 100644
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1841,7 +1841,9 @@ local function bucket_send_build_opts(opts, default_timeout)
             opts[name] = default
         end
     end
-    opts.deadline = fiber_clock() + opts.timeout
+    if not opts.deadline then
+        opts.deadline = fiber_clock() + opts.timeout
+    end
     return opts
 end
```

</p>
</details>

Removed the call for `build_opts` and fixed it

vshard/storage/init.lua
Serpentian: > Or another idea

I've thought about that. Though, placing it in `rebalancer_dispenser_pop` doesn't make `route_dispenser.lua` to use the deadline, which it commits, it's just a wrapper inside `storage/init.lua`. And I don't really like saving `opts` on commit and returning them from `dispenser:pop`, we already return `rs_id, bid`, returning the third one seems too much.

The second problem, I have absolutely no access to dispenser from the `M`, so I won't be able to change the timeouts on the fly, if smth goes wrong on production and I'll want to change them. So, let's better build the opts on every iteration, seems like the best approach to me here

vshard/storage/init.lua
Serpentian: This situation means, that we didn't manage to wait for buckets to be prepared and all of the workers start exiting. Moreover, it can happen iff `rebalancer_bucket_send_timeout` is not infinity. We definitely not sending buckets ok now and I see no motivation to reset the deadline

Serpentian:
```diff
diff --git a/vshard/storage/init.lua b/vshard/storage/init.lua
index 8cd6e4c..3d06a09 100644
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -2660,6 +2660,11 @@ local function rebalancer_dispenser_pop(dispenser)
     -- a worker.
     do return end
 ::prepare::
+    -- The deadline is reset in order to exclude the time for waiting when
+    -- another worker will prepare the buckets, since it's possible, that the
+    -- current worker has waited for the bucket to be prepared, was woken up,
+    -- there're no buckets left and we should prepare the batch.
+    opts.deadline = fiber_clock() + opts.timeout
     dispenser:prepare_begin()
     count = math.min(M.rebalancer_worker_count, dispenser.remaining_count)
     buckets, err = rebalancer_prepare_buckets(count, opts)
```

Serpentian:
Fixed another flakiness in the `auto_master` found by `catch_flaky` (last commit):

```
[3](https://github.com/tarantool/vshard/actions/runs/26105350143/job/76767495513?pr=649#step:12:144)
[002] not ok 14	storage.is_sync:true.test_master_discovery_on_disconnect
[002] #   Instance with replica id 2 was promoted first
[002] #   stack traceback:
[002] #   	...d/vshard/test/storage-luatest/auto_master_2_2_2_test.lua:87: in function <...d/vshard/test/storage-luatest/auto_master_2_2_2_test.lua:87>
[002] #   	...d/vshard/test/storage-luatest/auto_master_2_2_2_test.lua:87: in function 'promote_if_needed'
[002] #   	...d/vshard/test/storage-luatest/auto_master_2_2_2_test.lua:419: in function 'storage.is_sync:true.test_master_discovery_on_disconnect'
[002] #   artifacts:
[002] #   	replica_2_b -> /home/runner/work/vshard/vshard/test/var/002_storage-luatest/artifacts/RYCZgV-9BpyN
[002] #   	replica_1_a -> /home/runner/work/vshard/vshard/test/var/002_storage-luatest/artifacts/Y9b0JJm5BE8f
[002] #   	replica_2_a -> /home/runner/work/vshard/vshard/test/var/002_storage-luatest/artifacts/7hBUHfOwlZIA
[002] # Ran 14 tests in 12.490 seconds, 13 succeeded, 1 errored
[002]
```

Gerold103 (approved):
LESSSS GOOOOO!!!

Serpentian:
* In the readonly commit: fix flaky `receiving_bucket` test + fix rw_lock being dropped on recreation of the ref + new test for that
* Additional last commit with fix for ref not being updated in the `bucket_send`

<details><summary>Details</summary>
<p>

```diff
diff --git a/test/rebalancer/receiving_bucket.result b/test/rebalancer/receiving_bucket.result
index 0f1263d..bf2868d 100644
--- a/test/rebalancer/receiving_bucket.result
+++ b/test/rebalancer/receiving_bucket.result
@@ -303,7 +303,11 @@ vshard.storage.buckets_info(1)
 f1 = fiber.create(function() ret, err = vshard.storage.bucket_send(1, util.replicasets[1], {timeout = 0.3}) end)
 ---
 ...
-while f1:status() ~= 'suspended' do fiber.sleep(0.01) end
+-- We must wait for rw_lock to be set inside the loop: if we don't do that,
+-- it's possible to read the bucket with `READONLY` status but without
+-- `rw_lock`: reading can happen, when the write has not yet returned to the TX
+-- thread, when `on_commit` trigger has not yet worked, when mvcc is disabled.
+while not vshard.storage.buckets_info(1)[1].rw_lock do fiber.sleep(0.01) end
 ---
 ...
 vshard.storage.buckets_info(1)
diff --git a/test/rebalancer/receiving_bucket.test.lua b/test/rebalancer/receiving_bucket.test.lua
index d7e1951..a502da0 100644
--- a/test/rebalancer/receiving_bucket.test.lua
+++ b/test/rebalancer/receiving_bucket.test.lua
@@ -128,7 +128,11 @@ f = fiber.create(function() vshard.storage.call(1, 'write', 'do_long_thing') end
 while f:status() ~= 'suspended' do fiber.sleep(0.01) end
 vshard.storage.buckets_info(1)
 f1 = fiber.create(function() ret, err = vshard.storage.bucket_send(1, util.replicasets[1], {timeout = 0.3}) end)
-while f1:status() ~= 'suspended' do fiber.sleep(0.01) end
+-- We must wait for rw_lock to be set inside the loop: if we don't do that,
+-- it's possible to read the bucket with `READONLY` status but without
+-- `rw_lock`: reading can happen, when the write has not yet returned to the TX
+-- thread, when `on_commit` trigger has not yet worked, when mvcc is disabled.
+while not vshard.storage.buckets_info(1)[1].rw_lock do fiber.sleep(0.01) end
 vshard.storage.buckets_info(1)
 vshard.storage.bucket_refrw(1)
 while f1:status() ~= 'dead' do fiber.sleep(0.01) end
diff --git a/test/storage-luatest/rebalancer_test.lua b/test/storage-luatest/rebalancer_test.lua
index 7c60127..94783bd 100644
--- a/test/storage-luatest/rebalancer_test.lua
+++ b/test/storage-luatest/rebalancer_test.lua
@@ -394,6 +394,47 @@ test_group.test_pinning_before_readonly = function(g)
     end, {g.replica_1_a:replicaset_uuid()})
 end
 
+test_group.test_readonly_rw_lock_after_ref_drop = function(g)
+    g.replica_1_a:exec(function(uuid)
+        local fiber = require('fiber')
+        local bid = _G.get_first_bucket()
+        ilt.assert(ivshard.storage.bucket_refrw(bid))
+        local f = fiber.create(function()
+            fiber.self():set_joinable(true)
+            local opts = {timeout = ivconst.TIMEOUT_INFINITY}
+            return ivshard.storage.bucket_send(bid, uuid, opts)
+        end)
+        local info
+        ilt.helpers.retrying({timeout = iwait_timeout}, function()
+            info = ivshard.storage.buckets_info(bid)[bid]
+            ilt.assert_equals(info.status, ivconst.BUCKET.READONLY)
+            -- We must wait for rw_lock to be set inside the loop: if we place
+            -- the check for `rw_lock` after this loop, we'll get the error,
+            -- since it's possible to read the bucket with `READONLY` status
+            -- but without `rw_lock`: reading can happen, when the write has
+            -- not yet returned to the TX thread, when `on_commit` trigger
+            -- has not yet worked, when mvcc is disabled.
+            ilt.assert(info.rw_lock)
+        end)
+        ivshard.storage.internal.bucket_refs[bid] = nil
+        ilt.assert(ivshard.storage.bucket_refro(bid))
+        info = ivshard.storage.buckets_info(bid)[bid]
+        ilt.assert(info.rw_lock)
+        -- The rw ref was dropped with bucket.
+        ilt.assert_equals(info.ref_rw, nil)
+        f:cancel()
+        ilt.assert(f:join())
+        ilt.assert(ivshard.storage.bucket_unrefro(bid))
+        -- Again, wait for `rw_lock` to be dropped.
+        ilt.helpers.retrying({timeout = iwait_timeout}, function()
+            ivshard.storage.recovery_wakeup()
+            info = ivshard.storage.buckets_info(bid)[bid]
+            ilt.assert_equals(info.status, ivconst.BUCKET.ACTIVE)
+            ilt.assert_not(info.rw_lock)
+        end)
+    end, {g.replica_2_a:replicaset_uuid()})
+end
+
 --
 -- Test, that the worker sends buckets in batches. Sending more buckets, than
 -- the storage has leads to error. After error during preparation all buckets
@@ -656,3 +697,38 @@ test_group.test_replication_not_broken_when_ref_on_replica = function(g)
         ivshard.storage.bucket_unrefrw(bid)
     end, {bid})
 end
+
+test_group.test_ref_change_noticed_during_bucket_send = function(g)
+    -- With MVCC enabled dirty read of the `rw_lock` is not possible.
+    t.run_only_if(not global_cfg.memtx_use_mvcc_engine)
+    g.replica_1_a:exec(function(uuid)
+        local fiber = require('fiber')
+        local bid = _G.get_first_bucket()
+        ilt.assert(ivshard.storage.bucket_refrw(bid))
+        local f = fiber.create(function()
+            fiber.self():set_joinable(true)
+            local opts = {chunk_timeout = -1}
+            return ivshard.storage.bucket_send(bid, uuid, opts)
+        end)
+        local info
+        ilt.helpers.retrying({timeout = iwait_timeout}, function()
+            info = ivshard.storage.buckets_info(bid)[bid]
+            ilt.assert_equals(info.status, ivconst.BUCKET.READONLY)
+        end)
+        -- See `test_readonly_rw_lock_after_ref_drop` for explanation.
+        ilt.assert_not(info.rw_lock)
+        -- The ref is recreated.
+        ivshard.storage.internal.bucket_refs[bid] = nil
+        ivshard.storage.bucket_refrw(bid)
+        ivshard.storage.bucket_unrefrw(bid)
+        local _, ok, err = f:join()
+        ilt.assert_not(ok, err)
+        ilt.assert(iverror.is_timeout(err))
+        ilt.helpers.retrying({timeout = iwait_timeout}, function()
+            ivshard.storage.recovery_wakeup()
+            info = ivshard.storage.buckets_info(bid)[bid]
+            ilt.assert_equals(info.status, ivconst.BUCKET.ACTIVE)
+            ilt.assert_not(info.rw_lock)
+        end)
+    end, {g.replica_2_a:replicaset_uuid()})
+end
diff --git a/vshard/storage/init.lua b/vshard/storage/init.lua
index 743748e..1e03c5b 100644
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -1324,12 +1324,13 @@ end
 local function bucket_refro(bucket_id)
     local ref = M.bucket_refs[bucket_id]
     if not ref then
-        local _, err = bucket_check_state(bucket_id, 'read')
+        local bucket, err = bucket_check_state(bucket_id, 'read')
         if err then
             return nil, err
         end
         ref = bucket_ref_new()
         ref.ro = 1
+        ref.rw_lock = not bucket_status_is_writable(bucket.status)
         M.bucket_refs[bucket_id] = ref
     elseif ref.ro_lock then
         return nil, lerror.vshard(lerror.code.BUCKET_IS_LOCKED, bucket_id)
@@ -1930,6 +1931,12 @@ local function bucket_send_prepare(bid, opts)
             goto error
         end
         lfiber.testcancel()
+        ref = M.bucket_refs[bid]
+        -- The ref was dropped, even though it was here. Probably the
+        -- bucket was manually dropped, while we were waiting, or the ref
+        -- itself was manually dropped. Should not happen without user
+        -- intervention.
+        assert(ref)
     end
     do return true end
```

</p>
</details>

test/storage-luatest/rebalancer_test.lua
```diff
@@ -293,3 +301,134 @@ test_group.test_rebalancer_mode = function(g)
...
+            function() return pcall(ivshard.storage.bucket_pin, bid) end)
+    end, {g.replica_2_a:replicaset_uuid()})
+    g.replica_2_a:exec(function(uuid)
+        ilt.assert(ivshard.storage.bucket_send(_G.get_first_bucket(), uuid))
+    end, {g.replica_1_a:replicaset_uuid()})
+end
+
+test_group.test_readonly_rw_lock_after_ref_drop = function(g)
+    g.replica_1_a:exec(function(uuid)
+        local fiber = require('fiber')
```
Gerold103: Nit: we have `ifiber` on storage instances in vtest.

vshard/storage/init.lua
```diff
@@ -1931,6 +1931,12 @@ local function bucket_send_prepare(bid, opts)
             goto error
         end
         lfiber.testcancel()
+        ref = M.bucket_refs[bid]
+        -- The ref was dropped, even though it was here. Probably the
+        -- bucket was manually dropped, while we were waiting, or the ref
+        -- itself was manually dropped. Should not happen without user
+        -- intervention.
+        assert(ref)
```
Gerold103: If the assertion actually fails due to some wrong actions from a user, we will never call `bucket_transfer_end()` on this bucket. I suggest we still handle it gracefully with the error returned like `BUCKET_IS_CORRUPTED` (we already have this one).

Normally an assertion in C terminates the process via `SIGABRT` and we never continue in the wrong state, but in Lua we will continue execution in the wrong state.

test/storage-luatest/rebalancer_test.lua
Serpentian: Fixed in both tests

vshard/storage/init.lua
Serpentian: Agree, fixed

Gerold103 (approved):
🥁🥁🥁🥁🥁🥁 🔥💪⚡️🎉🎺 LESSSS GOOOOO 2!!!!!

PR #653, opened 2026-06-03
test: fix flakiness of rebalancer stress tests

### LINKED ISSUE #652: test: fix flaky of rebalancer/restart_during_rebalancing
**Environment**:
  - Tarantool version: 3.7.0-entrypoint-100-gba48c4e4e8
  - Vshard version: 0.1.41
  - OS: linux, x86-64

After the [storage: check for refs on replicas before sending a bucket](https://github.com/tarantool/vshard/commit/b495f8b34a559a9109d7471cc95bf3c9c0cbf0ca) commit, the stress test - `rebalancer/restart_during_rebalancing` started to flaks. The flakiness occurred both in [vshard test CI](https://github.com/tarantool/vshard/actions/runs/26518047126/job/78100784610) and in [tarantool integration test CI](https://github.com/tarantool/tarantool/actions/runs/26417610805/job/77766524412).

Based on the logs of the instance where the rebalancer was running:
- `Error during rebalancer routes applying: receiver ... : ... {"code":0,"base_type":"FiberIsCancelled" ...}`
- `disconnected from replica(id=box_2_a, uri=storage@127.0.0.1:3303)`
- `cannot gracefully shutdown client fibers`

It can be concluded that the rebalancer does not have time to complete its work before the `box_2_a`, `box_3_a`, and `box_4_a` instances are killed (the timeout of test passed).

<br>

This flakiness happens because:

1) After the commit [storage: check for refs on replicas before sending a bucket](https://github.com/tarantool/vshard/commit/b495f8b34a559a9109d7471cc95bf3c9c0cbf0ca), the rebalancing time increased significantly due to the introduction of `bucket_test_send` and `bucket_test_send_on_replicas` functions. I checked the test execution time before and after this commit. Before the commit, the test was performed for ~30-35 seconds, and after the commit for ~90-100 seconds.

2) Stable fluctuating test - `restart_during_rebalancing` is a stress test. It makes a constant read/write load on instance buckets in separate fibers + randomly restarts instances (including the one on which the rebalancer is located)

3) CI tests are running in parallel (heavy load on the server)

<br>

How to fix: I think that first of all we need to correct this and other rebalancer stress tests by reducing the number of buckets that the rebalancer works with. He's currently sending 200 buckets back and forth, and it may be necessary to reduce their number to 100 e.g. until CI is stable.

mrForza:
After the commit `storage: check for refs on replicas before sending a bucket`, the rebalancing time increased significantly due to the introduction of `bucket_test_send` and `bucket_test_send_on_replicas` functions. Consequently, the duration of the rebalancer tests has also increased (e.g. the average execution time for the `restart_during_rebalancing` was about 30 seconds before that, but now it's 90 seconds). As a result, some rebalancer tests started to fail in vshard and other CI (more details are given in issue ticket).

In order to fix the flakiness we reduce the amount of buckets which we create in cluster in rebalancer tests from `200` to `100`. We change this value not only in `restart_during_rebalancing` test, but also in all tests in order to prevent hypothetical flakiness of other tests.

Closes tarantool/vshard#652

NO_TEST=test
NO_DOC=test

### DIFF
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -336,6 +336,7 @@ local function master_call(replicaset, func, args, opts)
     local deadline = fiber_clock() + opts.timeout
     local did_first_attempt = false
     while true do
+        lfiber.testcancel()
         local res, err = replicaset:callrw(func, args, opts)
         if res ~= nil then
             return res
@@ -1131,7 +1132,6 @@ local function recovery_step_by_type(type, limiter)
             end
             goto continue
         end
-        lfiber.testcancel()
         remote_bucket, err = master_call(
             destination, 'vshard.storage._call',
             {'recovery_bucket_stat', bucket_id},
@@ -3230,7 +3230,6 @@ local function rebalancer_service_f(service, limiter)
         for src_id, src_routes in pairs(routes) do
             service:set_activity('applying routes')
             local rs = M.replicasets[src_id]
-            lfiber.testcancel()
             local status, err = master_call(
                 rs, 'vshard.storage.rebalancer_apply_routes', {src_routes},
                 {timeout = consts.REBALANCER_APPLY_ROUTES_TIMEOUT})
--- a/test/rebalancer/bucket_ref.result
+++ b/test/rebalancer/bucket_ref.result
@@ -38,7 +38,7 @@ _ = test_run:switch('box_1_a')
 vshard.storage.rebalancer_disable()
 ---
 ...
-vshard.storage.bucket_force_create(1, 100)
+vshard.storage.bucket_force_create(1, 50)
 ---
 - true
 ...
@@ -48,7 +48,7 @@ _ = test_run:switch('box_2_a')
 vshard.storage.rebalancer_disable()
 ---
 ...
-vshard.storage.bucket_force_create(101, 100)
+vshard.storage.bucket_force_create(101, 50)
 ---
 - true
 ...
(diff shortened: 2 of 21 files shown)


Gerold103 (approved):

vshard/storage/init.lua
```diff
@@ -3137,6 +3141,7 @@ local function rebalancer_download_states()
     local total_bucket_locked_count = 0
     local total_bucket_active_count = 0
     for id, replicaset in pairs(M.replicasets) do
+        lfiber.testcancel()
```
Serpentian: Nit: maybe into beginning of the loop inside `master_call`? It'll be better, since the `master_call` is retried inside. Moreover, we'll definitely forget to add the `testcancel`, when new `master_call` is added. And drop the existing `testcancel` before `master_call`

P.S. I'd actually stop on that. `testcancel` before `masters_map_call` doesn't make a lot of sense, there a lot of waiting inside without any testcancels, it'll be non trivial to add it there, and the `recovery` is definitely not the one, which causes failure of our tests (as you've figured out). So, I'd drop that for now

vshard/storage/init.lua
```diff
@@ -1048,6 +1048,7 @@ local function recovery_cluster_bucket_stat(bucket_id, destination)
         end
     end
     local call_opts = {timeout = consts.RECOVERY_GET_STAT_TIMEOUT}
+    lfiber.testcancel()
```
Serpentian: In the commit msg incorrect ticket number: should be #652, now we mention the PR

vshard/storage/init.lua
Serpentian: > And drop the existing testcancel before master_call

@mrForza. There're a lot of places in storage/init.lua, where it's called right before `master_call`. Let's drop it, while we're here)

Serpentian (approved):

PR #656, opened 2026-07-01
test: cover storage-first upgrade

### LINKED ISSUE #654: Test and document VShard upgrade order
[The doc](https://www.tarantool.io/en/doc/latest/admin/upgrades/upgrade_cluster/) descrbes router-first upgrade. However, we agreed on the following order of the upgrade for vshard and above-standing projects: firstly, upgrade the replicasets, then upgrade routers. The motivation for such order is that storage needs nothing from router, only router initiates communication with storage.

It's already applied in the ATE, we need to test that behavior and document it.

We already have the bunch of reload tests:
- `test/storage/reload.test.lua`: basic storage reload mechanics.
- `test/router/reload.test.lua`: basic for router.
- `test/reload_evolution/storage.test.lua`: storage reload from an old vshard version to latest. Checks bucket transfer.
- `test/upgrade/upgrade.test.lua`: restart-based storage upgrade
- `test/router-luatest/reload_test.lua`: router reload from old vshard to latest; checks calls, discovery, master search, and router work with a new master.

We need to add new tests, which cover the full upgrade process and check the basic functionality of the vshard. The hot-reload upgrade is preferred, but if possible and not costly, restart should also be tested (can be easily achieved with `test_group` config).

We should test the following scenarios:
* Upgrade from some old version to the latest commit (as it's already done in the existing tests)
* Upgrade from prev version to the latest commit.

kamenkremen:
Closes #654

### DIFF
--- a/test/cluster-luatest/reload_test.lua
+++ b/test/cluster-luatest/reload_test.lua
@@ -1,0 +1,694 @@
+local t = require('luatest')
+local vutil = require('vshard.util')
+local vtest = require('test.luatest_helpers.vtest')
+local vconsts = require('vshard.consts')
+local git_util = require('test.lua_libs.git_util')
+local upgrade = require('test.luatest_helpers.upgrade')
+local fio = require('fio')
+
+local g = t.group('reload_storage_router', {
+    { update_mode = 'hot_reload' },
+    { update_mode = 'restart' },
+})
+local cfg_template = {
+    sharding = {
+        {
+            master = 'auto',
+            replicas = {
+                replica_1_a = {
+                    read_only = false,
+                },
+                replica_1_b = {
+                    read_only = true,
+                },
+            },
+        },
+        {
+            master = 'auto',
+            replicas = {
+                replica_2_a = {
+                    read_only = false,
+                },
+                replica_2_b = {
+                    read_only = true,
+                },
+            },
+        },
+    },
+    bucket_count = 100,
+}
+local global_cfg
+local update_mode
+local storage_names = {
+    'replica_1_a',
+    'replica_1_b',
+    'replica_2_a',
+    'replica_2_b',
+}
+-- Oldest commit that supports master = 'auto' on storage.
+local legacy_storage_reload_hash = '775f5e67a02c3f71b1954c397765ed99dbf4ff7d'
+
+local function checkout_vshard(path, hash)
+    git_util.exec('checkout', { args = hash .. ' -f', dir = path })
+end
+
+g.before_all(function(cg)
+    -- Tarantool's override loader is available since 2.11.
+    t.run_only_if(vutil.version_is_at_least(2, 11, 0, nil, 0, 0))
+    update_mode = cg.params.update_mode
+
+    -- The test works in the following directory.
+    local vardir = vtest.vardir or fio.tempdir()
+    local storage_copy_root = vardir .. '/vshard_storage_copy_' .. update_mode
+    fio.rmtree(storage_copy_root)
+    g.storage_vshard_copy_path = storage_copy_root .. '/override'
+    local router_copy_root = vardir .. '/vshard_router_copy_' .. update_mode
+    fio.rmtree(router_copy_root)
+    g.router_vshard_copy_path = router_copy_root .. '/override'
+
+    --
+    -- Tarantool searches for compilation units in the following order:
+    --   1. preload --> override --> builtin
+    --   2. path.cwd.dot
+    --   3. cpath.cwd.dot
+    --   4. path.cwd.rocks
+    --   5. cpath.cwd.rocks
+    --   6. package.path
+    --   7. package.cpath
+    --   8. croot
+    --
+    -- Since the test is launched from the repository directory, newest
+    -- vshard will always be loaded, regardless of package.path or `LUA_PATH`.
+    -- So we can use package.setsearchroot() to change cwd, luatest's `chdir`
+    -- or override. The last one is used here, since it's the easiest.
+    --
+    g.storage_vshard_lua_path = upgrade.vshard_copy_new(
+        vtest.sourcedir, g.storage_vshard_copy_path)
+    g.router_vshard_lua_path = upgrade.vshard_copy_new(
+        vtest.sourcedir, g.router_vshard_copy_path)
+
+    g.latest_hash = git_util.log_hashes({ args = '-1',
+                                          dir = vtest.sourcedir })[1]
+
+    local temp_file = os.tmpname()
+    git_util.exec('describe', {
+        args = "--tags --abbrev=0 --match '[0-9]*.[0-9]*.[0-9]*' HEAD^",
+        dir = vtest.sourcedir,
+        fout = temp_file,
+    })
+    g.previous_version = io.lines(temp_file)()
+    os.remove(temp_file)
+    t.assert_not_equals(g.previous_version, nil, 'previous release tag')
+end)
+
+local function assert_versions(router, storage_versions, router_version)
+    for _, storage_version in ipairs(storage_versions) do
+        local name = storage_version[1]
+        local expected_version = storage_version[2]
+        local _, err = g[name]:exec(function(version)
+            ilt.assert_equals(ivconst.VERSION, version)
+        end, { expected_version })
+        t.assert_equals(err, nil)
+    end
+    router:exec(function(version)
+        ilt.assert_equals(ivconst.VERSION, version)
+    end, { router_version })
+end
+
+local function storage_test_functions_install()
+    rawset(_G, 'test_replace', function(space_name, tuple)
+        return box.space[space_name]:replace(tuple)
+    end)
+    rawset(_G, 'test_get', function(space_name, key)
+        return box.space[space_name]:get(key)
+    end)
+    rawset(_G, 'test_delete', function(space_name, key)
+        return box.space[space_name]:delete(key)
+    end)
+end
+
+local function storage_test_data_create()
+    local test = box.schema.space.create('test', { format = {
+        { 'id', 'unsigned' },
+        { 'bucket_id', 'unsigned' },
+    } })
+    test:create_index('primary')
+    test:create_index('bucket_id', {
+        unique = false,
+        parts = { 2 },
+    })
+
+    box.schema.func.create('test_replace')
+    box.schema.func.create('test_get')
+    box.schema.func.create('test_delete')
+end
+
+g.after_each(function()
+    for _, server in ipairs(g.cluster.servers) do
+        if server.log_file ~= nil and fio.path.exists(server.log_file) then
+            local msg = server:grep_log('[Aa]ssertion failed',
+                                        64 * 1024 * 1024, {noreset = true})
+            t.assert_equals(msg, nil)
+        end
+    end
+
+    g.cluster:drop()
+    g.cluster = nil
+end)
+
+local function storage_cfg_by_name(names, cfg)
+    for _, name in ipairs(names) do
+        vtest.storage_cfg(g[name], cfg)
+    end
+end
+
+--
+-- Upgrade test workflow:
+--     1. Start storages and router on the old vshard version;
+--     2. Run all checks;
+--     3. Update the storages in a separate replicaset;
+--     4. Run all checks;
+--     5. Update the future master replica;
+--     6. Run all checks;
+--     7. Move master to the updated replica;
+--     8. Run all checks;
+--     9. Update the previous master replica;
+--    10. Run all checks.
+--    11. Update router;
+--    12. Run all checks.
+--
+local function update_storages(names)
+    checkout_vshard(g.storage_vshard_copy_path, g.latest_hash)
+    for _, name in ipairs(names) do
+        local storage = g[name]
+        if update_mode == 'hot_reload' then
+            local _, err = storage:exec(function(lua_path)
+                package.path = lua_path
+                package.loaded['vshard.storage'] = nil
+                _G.vshard.storage = require('vshard.storage')
+                _G.ivconst = require('vshard.consts')
+            end, {
+                upgrade.vshard_lua_path(g.storage_vshard_copy_path),
+            })
+            t.assert_equals(err, nil)
+        else
+            vtest.storage_stop(storage)
+            vtest.storage_start(storage, global_cfg)
+        end
+        local _, err = storage:exec(storage_test_functions_install)
+        t.assert_equals(err, nil, 'install storage test functions')
+        _, err = storage:exec(function()
+            ivshard.storage.rebalancer_disable()
+        end)
+        t.assert_equals(err, nil)
+    end
+end
+
+local function update_router(router)
+    checkout_vshard(g.router_vshard_copy_path, g.latest_hash)
+    if update_mode == 'hot_reload' then
+        router:exec(function()
+            ilt.assert_equals(ivshard.router.module_version(), 0)
+            package.loaded['vshard.router'] = nil
+            ivshard.router = require('vshard.router')
+            _G.ivconst = require('vshard.consts')
+            ilt.assert_equals(ivshard.router.module_version(), 1)
+        end)
+    else
+        router:restart()
+        vtest.router_cfg(router, global_cfg)
+    end
+end
+
+-- Wait until router background services finish without errors.
+local function check_router_services(router)
+    local _, err = router:exec(function()
+        local static_router = ivshard.router.static
+        local function wait_worker_service_ok(worker, name)
+            local function wakeup()
+                worker:wakeup_service(name)
+            end
+            local service = worker.services[name]
+            ivtest.wait_for_not_nil(service.data, 'info',
+                                    { on_yield = wakeup })
+            ivtest.service_wait_for_ok(service.data.info,
+                                       { on_yield = wakeup })
+        end
+
+        if static_router.master_search_service ~= nil then
+            ivtest.service_wait_for_ok(static_router.master_search_service,
+                { on_yield = ivshard.router.master_search_wakeup })
+        else
+            for _, replicaset in pairs(static_router.replicasets) do
+                wait_worker_service_ok(replicaset.worker,
+                                       'replicaset_master_search')
+            end
+        end
+
+        local _, failover_replicaset = next(static_router.replicasets)
+        if failover_replicaset.worker == nil then
+            ivtest.service_wait_for_ok(static_router.failover_service, {
+                on_yield = function()
+                    static_router.failover_fiber:wakeup()
+                end,
+            })
+        else
+            for _, replicaset in pairs(static_router.replicasets) do
+                wait_worker_service_ok(replicaset.worker,
+                                       'replicaset_failover')
+                for _, replica in pairs(replicaset.replicas) do
+                    wait_worker_service_ok(replica.worker, 'replica_failover')
+                end
+            end
+        end
+    end)
+    t.assert_equals(err, nil, 'router services')
+end
+
+-- Check that the router sees the expected master.
+local function check_router_master(router, master_uuid)
+    local rs_uuid = g.replica_1_a:replicaset_uuid()
+    local _, err = router:exec(function(check_rs_uuid, check_master_uuid)
+        local static_router = ivshard.router.static
+        ilt.helpers.retrying({}, function()
+            ivshard.router.master_search_wakeup()
+            local replicaset = static_router.replicasets[check_rs_uuid]
+            ilt.assert_not_equals(replicaset.master, nil)
+            ilt.assert_equals(check_master_uuid, replicaset.master.uuid)
+        end)
+    end, { rs_uuid, master_uuid })
+    t.assert_equals(err, nil, 'router expected master')
+end
+
+local function storage_wait_master_sync_legacy_error(storage)
+    local _, err = storage:exec(function()
+        ilt.helpers.retrying({ timeout = iwait_timeout }, function()
+            ivshard.storage.master_sync_wakeup()
+            local internal = ivshard.storage.internal
+            local service = internal.master_sync_service
+            ilt.assert_not_equals(service, nil)
+            ilt.assert_equals(service.status, 'error')
+            local is_lua_error =
+                service.error:find('LuajitError', 1, true) ~= nil or
+                service.error:find('ClientError', 1, true) ~= nil
+            ilt.assert(is_lua_error)
+            ilt.assert_str_contains(service.error,
+                                    'attempt to call a nil value')
+            ilt.assert_not(internal.is_bucket_in_sync)
+        end)
+    end)
+    t.assert_equals(err, nil, 'master sync legacy error')
+end
+
+-- Wait until background discovery has found all buckets.
+local function check_router_discovery(router)
+    local _, err = router:exec(function()
+        ilt.helpers.retrying({}, function()
+            ivshard.router.discovery_wakeup()
+            ilt.assert_equals(ivshard.router.info().bucket.unknown, 0)
+        end)
+    end)
+    t.assert_equals(err, nil, 'router discovery')
+end
+
+-- Check routed write, synchronization, and read calls on both replicasets.
+local function check_router_calls(router)
+    local bucket_ids = {
+        vtest.storage_first_bucket(g.replica_1_a),
+        vtest.storage_first_bucket(g.replica_2_a),
+    }
+    t.assert_not_equals(bucket_ids[1], nil, 'replica_1_a first active bucket')
+    t.assert_not_equals(bucket_ids[2], nil, 'replica_2_a first active bucket')
+
+    local _, err = router:exec(function(check_bucket_ids)
+        for _, bucket_id in ipairs(check_bucket_ids) do
+            local tuple = { bucket_id, bucket_id }
+            local res, err = ivshard.router.callrw(bucket_id, 'test_replace',
+                                                   { 'test', tuple })
+            ilt.assert_equals(err, nil)
+            ilt.assert_equals(res, tuple)
+            local _, sync_err = ivshard.router.callrw(bucket_id,
+                                                      'vshard.storage.sync',
+                                                      {})
+            ilt.assert_equals(sync_err, nil)
+            res, err = ivshard.router.callro(bucket_id, 'test_get',
+                                             { 'test', { bucket_id } })
+            ilt.assert_equals(err, nil)
+            ilt.assert_equals(res, tuple)
+            res, err = ivshard.router.callrw(bucket_id, 'test_delete',
+                                              { 'test', { bucket_id } })
+            ilt.assert_equals(err, nil)
+            ilt.assert_equals(res, tuple)
+        end
+    end, { bucket_ids })
+    t.assert_equals(err, nil, 'router read/write calls')
+end
+
+local function router_wait_prioritized(router, rs_uuid, expected_uuid)
+    local _, err = router:exec(function(check_rs_uuid, check_uuid)
+        local static_router = ivshard.router.static
+        local replicaset = static_router.replicasets[check_rs_uuid]
+        local function failover_wakeup()
+            if replicaset.worker == nil then
+                static_router.failover_fiber:wakeup()
+                return
+            end
+            replicaset.worker:wakeup_service('replicaset_failover')
+            for _, replica in pairs(replicaset.replicas) do
+                replica.worker:wakeup_service('replica_failover')
+            end
+        end
+        ilt.helpers.retrying({}, function()
+            failover_wakeup()
+            ilt.assert_not_equals(replicaset.replica, nil)
+            ilt.assert_equals(replicaset.replica.uuid, check_uuid)
+        end)
+    end, { rs_uuid, expected_uuid })
+    t.assert_equals(err, nil, 'wait router prioritized replica')
+end
+
+local function failover_router_cfg()
+    local cfg = table.deepcopy(global_cfg)
+    local rs_uuid = g.replica_1_a:replicaset_uuid()
+    local replica_a_uuid = g.replica_1_a:instance_uuid()
+    local replica_b_uuid = g.replica_1_b:instance_uuid()
+
+    cfg.sharding[rs_uuid].replicas[replica_a_uuid].zone = 3
+    cfg.sharding[rs_uuid].replicas[replica_b_uuid].zone = 2
+    cfg.zone = 1
+    cfg.weights = {
+        [1] = {
+            [1] = 0,
+            [2] = 1,
+            [3] = 2,
+        },
+    }
+    -- Make failed ping attempts complete quickly and deterministically.
+    cfg.failover_ping_timeout = 0.1
+    return cfg
+end
+
+-- Check that router priority changes on a failure and is restored on recovery.
+-- Both priority transitions use short timeouts to keep this check lightweight.
+local function check_router_failover(router)
+    vtest.router_cfg(router, failover_router_cfg())
+    local timeout_state
+    local ok, check_err = pcall(function()
+        local rs_uuid = g.replica_1_a:replicaset_uuid()
+        -- replica_1_b has the lowest cost from the router's zone.
+        local prioritized_uuid = g.replica_1_b:instance_uuid()
+        router_wait_prioritized(router, rs_uuid, prioritized_uuid)
+        local prioritized = g.replica_1_b
+        local timeout_err
+        timeout_state, timeout_err = router:exec(function()
+            local old = {
+                down = ivconst.FAILOVER_DOWN_TIMEOUT,
+                up = ivconst.FAILOVER_UP_TIMEOUT,
+            }
+            ivconst.FAILOVER_DOWN_TIMEOUT = 0.01
+            ivconst.FAILOVER_UP_TIMEOUT = 0.01
+            return old
+        end)
+        t.assert_equals(timeout_err, nil, 'set failover timeouts')
+
+        prioritized:freeze()
+        local changed, change_err = pcall(function()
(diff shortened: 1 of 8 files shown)


test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+local cluster = require('test.luatest_helpers.cluster')
+
+local group_config = {
+    { update_mode = 'hot_reload' },
+    { update_mode = 'restart' },
+}
+local g = t.group('reload_storage_router', group_config)
+local cfg_template = {
+    sharding = {
+        { replicas = {
```
Serpentian: Nit: the formating is broken. Let's be consistent with other tests in code-style

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+        dir = vtest.sourcedir,
+        fout = temp_file,
+    })
+    g.previous_version = io.lines(temp_file)()
+    os.remove(temp_file)
+    t.assert_not_equals(g.previous_version, nil, 'previous release tag')
+end)
+
+--
+-- `vtest.router_cfg` cannot be used in this test, since
```
Serpentian: Let's make the call for that funciton optional then in `vtest`. We should reuse the same functions for all of the tests and not rewrite them for every individual one. I see, that it's copy-pasted from `reload_test`, then let's use the same vtest function there.

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+end
+
+-- Legacy vshard still expects box.info.cluster.uuid to be the
+-- replicaset UUID after box.cfg().
+local function compat_preload_new(vardir)
+    local path = fio.pathjoin(vardir, 'vshard_box_info_cluster_old_' ..
+                                      update_mode .. '.lua')
+    local file = assert(io.open(path, 'w'))
+    file:write([[
+        local ok, compat = pcall(require, 'compat')
```
Serpentian: This is not an option, we'll soon release the new Tarantool version, where this compat and the associated code will be dropped, so that it won't be possible to use old `box.info.cluster.uuid`. So this fix is only temporary. I'd propose to unconditionally set it in the test instead

```lua
box.info.cluster.uuid = box.info.replicaset.uuid
```

test/router-luatest/reload_storage_test.lua
Serpentian, line 1: I'd propose to create the new directory and call it `cluster_luatest` and put the test there. It's not really related to the router, it tests upgrade of the whole cluster

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+    },
+    bucket_count = 100,
+}
+local global_cfg
+local update_mode
+-- Old enough storage hash selected from
+-- test/reload_evolution/storage.test.lua for reload to latest.
+local legacy_storage_reload_hash = '139223269cddefe2ba4b8e9f6e44712f099f4b35'
+
+local function vshard_copy_new(vardir, dirname)
```
Serpentian: Let's move the function to the `git_util.lua` and reuse it in the `reload_test` too. Let's try to minimize the copy-pasted code between these two tests.

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+    local vshard_copy_path_load = vardir .. '/' .. dirname
+    t.assert_equals(fio.mkdir(vshard_copy_path_load), true)
+
+    local vshard_copy_path = vshard_copy_path_load .. '/override'
+    t.assert_equals(fio.mkdir(vshard_copy_path), true)
+    -- Copy source to the temporary directory.
+    t.assert_equals(fio.mkdir(vshard_copy_path .. '/.git'), true)
+    t.assert_equals(fio.copytree(vtest.sourcedir .. '/.git',
+                                 vshard_copy_path .. '/.git'), true)
+    return vshard_copy_path_load, vshard_copy_path
```
Serpentian: I'd prefer, that we return the lua_path instead of the `copy_path_load`. And I propose to pass the whole directory there with `override` as argument to that function, so that you don't have to return it from the function.

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+        local ok, compat = pcall(require, 'compat')
+        if ok and
+           compat._options and
+           compat._options().box_info_cluster_meaning then
+            compat.box_info_cluster_meaning = 'old'
+        end
+    ]])
+
+    if not vutil.version_is_at_least(3, 0, 0, nil, 0, 0) then
+        file:write([[
```
Serpentian: I don't understand that change. Could you please add the explanation comment to it

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+        local _, err = storage:exec(function(cfg)
+            ivtest.clear_test_cfg_options(cfg)
+            ivshard.storage.cfg(cfg, _G.box_vstorage_id(cfg))
+        end, { cfg })
+
+        t.assert_equals(err, nil, name .. ' storage reconfig')
+    end
+end
+
+-- On old versions storage.internal.is_master may be nil,
```
Serpentian: Let's try making the vtest to work with old version instead of introducing manual bootstrap function

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+            ilt.assert_equals(sync_err, nil)
+            res, err = ivshard.router.callro(bucket_id, 'get',
+                                             { 'test', { bucket_id } })
+            ilt.assert_equals(err, nil)
+            ilt.assert_equals(res, tuple)
+        end
+    end, { rs_uuid, master_uuid, bucket_ids })
+end
+
+storage_cfg_by_name = function(names, cfg)
```
Serpentian: Why using the forward declaration of the function?

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+    for _, args in ipairs(boot) do
+        local ok, err = args.storage:exec(function(first, count)
+            return ivshard.storage.bucket_force_create(first, count)
+        end, { args.first, args.count })
+        t.assert_equals(err, nil, 'storage bootstrap')
+        t.assert(ok, 'storage bootstrap')
+    end
+end
+
+-- vtest.storage_first_bucket fails on old versions.
```
Serpentian: Same

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+    end)
+    t.assert_equals(err, nil, name .. ' first active bucket')
+    t.assert_not_equals(bucket_id, nil, name .. ' first active bucket')
+    return bucket_id
+end
+
+local function create_cluster_at(opts)
+    global_cfg = vtest.config_new(cfg_template)
+    checkout_vshard(g.storage_vshard_copy_path, opts.hash)
+    g.cluster = cluster:new({})
```
Serpentian: Why do we create the cluster before `cluster_new`? It already does it internally

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+        { 'replica_1_b', vconsts.VERSION },
+        { 'replica_2_a', vconsts.VERSION },
+    }
+
+    -- First phase: everything is on old vshard.
+    assert_versions(router, all_old, opts.storage_version)
+    check_cluster(router, old_master_uuid, bucket_ids)
+
+    -- Second phase: two replicas are updated on new vshard.
+    update_storages({ 'replica_1_b', 'replica_2_a' })
```
Serpentian: Let's add one more phase: only one replicaset is updated, the second one is on the old version. Just split that phase to two.

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+    assert_versions(router, mixed, opts.storage_version)
+    check_cluster(router, old_master_uuid, bucket_ids)
+
+    -- Third phase: master is moved to the updated replica, previous master
+    -- is still on old vshard.
+    local master_id, master_lsn = old_master:exec(function()
```
Serpentian: You're using the manual master mode, so such master switch won't work, it requires `master = 'auto'` mode. Let's use auto, it's more widely used and is more sophisticated implementation-wise

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+            local _, err = storage:exec(function(lua_path)
+                package.path = lua_path
+                package.loaded['vshard.storage'] = nil
+                _G.vshard.storage = require('vshard.storage')
+                _G.ivconst = require('vshard.consts')
+            end, { vshard_lua_path(g.storage_vshard_copy_path) })
+            t.assert_equals(err, nil, name .. ' storage reload')
+        else
+            storage:restart()
+            storage_cfg_by_name({ name }, global_cfg)
```
Serpentian: Let's use `vtest.storage_stop/start`, it reconfigures the storage properly after restart

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+            local _, err = storage:exec(storage_test_api_create)
```
Serpentian: Let's better use persistent lua functions in order not to recreate them after every restart: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_schema/func_create/

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+            t.assert_equals(err, nil, name .. ' storage test api')
+        end
+    end
+end
+
+local function check_cluster(router, master_uuid, bucket_ids)
```
Serpentian: This is a good starting point. Let's try adding the master_search, failover tests. After every test let's grep, that logs doesn't have `assertion failed` (I remember, we had that problem in failover). Let's also check moving a bucket (manual bucket_send is enough) + its discovery on router.

test/router-luatest/reload_storage_test.lua
```diff
@@ -0,0 +1,426 @@
...
+        legacy_storage_start = true,
+    })
+end
+
+g.test_rolling_previous_storage_then_router_upgrade = function()
+    run_rolling_upgrade({
+        hash = g.previous_version,
+        storage_version = g.previous_version,
+    })
+end
```
Serpentian: I'd propose to split the commits as follows:

1. Refactoring of the `reload_test`. Moves some functions out of it to `git_utils`, so that they can be reused in the next commit.
2. Introduce the test + all the changes in the `vtest` needed for that test to work.

See, the first commit you have right now is not valuable without the second one, we should not split them like that. You add the feature to `vtest`, which is not used in the first commit

Serpentian (commented):
Solid starting point! Thank you a lot

test/router-luatest/reload_storage_test.lua
kamenkremen: I was trying to fix failures on tarantool 2.11 with this change, but it seems the problem is in the compat usage. I will remove it along with compat

test/router-luatest/reload_storage_test.lua
kamenkremen: This is an artifact of previous code changes that I forgot to remove

test/router-luatest/reload_storage_test.lua
kamenkremen: It seems that the vshard version used here as legacy does not support the master mode option in replicaset_template

Why doesn’t this work now? I tried to follow the documentation for switching the master and used the `legacy` option: https://www.tarantool.io/en/doc/latest/admin/upgrades/upgrade_cluster/#switching-the-master

Serpentian: >  It seems that the vshard version used here as legacy does not support the master mode option in replicaset_template

I propose to bump the commit hash then and use in the test the one, which has auto master available. We need to guarantee this upgrade procedure for the newest version, commit with auto master is new enough for that

It'll cover a lot more code, if we use auto master

test/cluster-luatest/reload_test.lua
```diff
@@ -0,0 +1,427 @@
...
+            _G.ivconst = require('vshard.consts')
+            ilt.assert_equals(ivshard.router.module_version(), 1)
+        end)
+    else
+        router:restart()
+        vtest.router_cfg(router, global_cfg)
+    end
+end
+
+local function check_cluster(router, master_uuid, move_src, move_dst)
```
Serpentian: Let's split the function to separate test cases:

```lua
local function check_cluster(stage, ...)
    check_basic(...)
    if rebalancing_must_work(stage) then
        check_rebalancing_and_discovery()
    else
        check_rebalancing_error()
    end
end
```

Every test case should return the cluster to the same state it was before starting the case. Rebalancing should work only on the following stages: all of the storages are old, sending bucket from the old master to the fully new replicaset, all of the storages are new. All other cases should not cause some arbitrary LuaJit error and should log `box.error.UNSUPPORTED` error instead

test/cluster-luatest/reload_test.lua
```diff
@@ -0,0 +1,427 @@
...
+                    read_only = true,
+                },
+            },
+        },
+        {
+            master = 'auto',
+            replicas = {
+                replica_2_a = {
+                    read_only = false,
+                },
```
Serpentian, line 31: Let's create the second replica, it never happens on real setups, when one replicaset has 2 replicas, the other one - 1. Let's test the widely used case: 2 replicas in every replicaset

vshard/storage/init.lua
```diff
@@ -3734,6 +3734,20 @@ local function master_sync_service_f(service, limiter)
...
             {'storage_bucket_checkpoint', call_timeout / 1.5}, call_opts)
+        if err and (err.type == 'LuajitError' or err.type == 'ClientError') and
+           err.message and err.message:match(
+               '/vshard/storage/init%.lua:%d+: attempt to call a nil value$')
+            then
+            -- Storages predating storage_bucket_checkpoint() have no
+            -- production capability endpoint. Their service dispatcher fails
+            -- with a plain Lua nil-call error when a service is missing.
+            map_res, err, err_id =
+                M.this_replicaset:map_call('vshard.storage.sync',
```
Serpentian: I don't understand that. We do not need to sync, if replica says, that the function is not defined. Instead, we should retry, until it succeeds (replica is upgraded). Let's just log the `UNSUPPORTED` with `prev` set to `LuajitError` and not the `Luajit` error only, if `attempt to call a nil value` is caught

Serpentian: Moreover, it may happen with all the places, where `vshard.storage._call` is used, so I'd propose to introduce the function and use it everywhere to replace that error with `UNSUPPORTED`.

The problem here is that we cannot distinguish `attempt to call a nil value` at the `service_call` from the same error inside any other function (only line number is available, which is useless). So, I suppose it's better to start returning the UNSUPPORTED error from the `service_call` from now on, though, it won't help on old versions, where the LuajitError is returned, so we still should process it anyway (however, at some time in point we'll be able to drop that code with Luajit parsing and suppose, that prev version always have that check for nil access.

@Gerold103, WDYT?

test/luatest_helpers/vtest.lua
```diff
@@ -687,7 +708,11 @@ end
 --
 local function router_cfg(router, cfg)
     router:exec(function(cfg)
-        ivtest.clear_test_cfg_options(cfg)
+        if ivtest.clear_test_cfg_options ~= nil then
+            ivtest.clear_test_cfg_options(cfg)
+        else
+            cfg.test_user_grant_range = nil
```
Serpentian: The same code block is repeated 5 times, can't we define the function inside  `instances/storage.lua`, it'll check for presence of the `vtest.clear_test_cfg_options`? Will it work? And I'd propose to merge it into `test: refactor router-luatest/reload_test.lua` commit alongside with `router_cfg` refactoring in the above mentioned test

vshard/router/init.lua
```diff
@@ -1831,9 +1831,9 @@ else
         if router.is_enabled == nil then
             router.is_enabled = true
         end
-        if router.api_call_cache == nil then
-            router.api_call_cache = router_api_call_unsafe
-        end
+        -- After reloading from an old vshard version, api_call_cache
+        -- can be nil or point to the old router_api_call signature.
+        router.api_call_cache = router_api_call_unsafe
```
Serpentian, line 1857: Makes sense, good. Feel free to resolve

vshard/storage/init.lua
```diff
@@ -4583,6 +4597,13 @@ if not rawget(_G, MODULE_INTERNALS) then
     rawset(_G, MODULE_INTERNALS, M)
 else
     reload_evolution.upgrade(M)
+    -- Hot reload from old vshard can leave master sync internals incompatible
+    -- with the reloaded code. Refresh the function and restart a dead fiber.
+    M.master_sync_f = master_sync_f
```
Serpentian, line 4644: This line is already executed above:

```lua
M.recovery_f = recovery_f
M.rebalancer_f = rebalancer_f
M.gc_bucket_f = gc_bucket_f
M.instance_watch_f = instance_watch_f
M.master_sync_f = master_sync_f
```

And I don't understand, how the fiber can become dead, if so, we should probably fix the service and not restart it. It must be restarted automatically on reload, since it's created via the `reloadable_fiber_new`

test/lua_libs/git_util.lua
```diff
@@ -40,7 +41,23 @@ local function log_hashes(params)
     return lines
 end
 
+local function vshard_lua_path(path)
+    return string.format('%s/?.lua;%s/?/init.lua;', path, path) ..
+           (os.getenv('LUA_PATH') or '')
+end
+
+local function vshard_copy_new(sourcedir, copy_path)
```
Serpentian: I know, I've proposed to move them here, though, I don't like that: the functions are not related to the git in any sense, can we introduce the `upgrade_utils.lua` and move them there?

test/cluster-luatest/reload_test.lua
```diff
@@ -0,0 +1,427 @@
+local t = require('luatest')
+local vutil = require('vshard.util')
+local vtest = require('test.luatest_helpers.vtest')
+local vconsts = require('vshard.consts')
+local git_util = require('test.lua_libs.git_util')
+local fio = require('fio')
+
+local group_config = {
```
Serpentian: Nit: let's just inline it, used only once

test/cluster-luatest/reload_test.lua
```diff
@@ -0,0 +1,427 @@
...
+local storage_names = { 'replica_1_a', 'replica_1_b', 'replica_2_a' }
+-- Oldest commit that supports master = 'auto' on storage.
+local legacy_storage_reload_hash = '775f5e67a02c3f71b1954c397765ed99dbf4ff7d'
+
+local function checkout_vshard(path, hash)
+    git_util.exec('checkout', { args = hash .. ' -f', dir = path })
+end
+
+g.before_all(function(cg)
+    -- Override of the built in modules is available only since 2.11.0.
```
Serpentian: Which built-in module do we override?

test/cluster-luatest/reload_test.lua
```diff
@@ -0,0 +1,427 @@
...
+    assert_versions(router, one_new, opts.storage_version)
+    check_cluster(router, old_master_uuid, g.replica_2_a, g.replica_1_a)
+
+    -- Third phase: another replica is updated on new vshard.
+    update_storages({ 'replica_1_b' })
+    assert_versions(router, mixed, opts.storage_version)
+    check_cluster(router, old_master_uuid, g.replica_2_a, g.replica_1_a)
+
+    -- Fourth phase: master is moved to the updated replica, previous master
+    -- is still on old vshard.
```
Serpentian, line 637: Hmm, it actually seems, that the rebalancing won't work at all in that upgrade process: switch master to the upgraded, this upgraded storage won't be able to sync and won't be able to send the bucket.

So, rebalancing in the middle of upgrade can only be tested in the `test_rolling_previous_storage_then_router_upgrade` test case. Up to you

vshard/storage/init.lua
Gerold103: - We certainly must not do any sort of error message matching to make any decisions in such code.
- Yep, handling missing functions gracefully is a good idea!

test/cluster-luatest/reload_test.lua
kamenkremen: Done

test/cluster-luatest/reload_test.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done. Errors from older vshard versions are normalized to UNSUPPORTED, with the original Lua error stored in prev. However, error matching is still used there

test/luatest_helpers/vtest.lua
kamenkremen: Changed this so vtest creates one sanitized configuration copy before passing it to storages or routers

vshard/storage/init.lua
kamenkremen: Removed the dead_fiber part, but M.master_sync_f = master_sync_f must be executed before storage_cfg, otherwise hot_reload from an older vshard version fails

test/lua_libs/git_util.lua
kamenkremen: Done

test/cluster-luatest/reload_test.lua
kamenkremen: Done

test/cluster-luatest/reload_test.lua
kamenkremen: I took that part of the code from the router test. Changed it in both places

test/cluster-luatest/reload_test.lua
kamenkremen: Done

test/lua_libs/git_util.lua
```diff
@@ -42,5 +42,5 @@ end
 
 return {
     exec = exec,
-    log_hashes = log_hashes
+    log_hashes = log_hashes,
```
Gerold103: Unnecessary diff.

test/lua_libs/upgrade_utils.lua
```diff
@@ -0,0 +1,24 @@
...
+local fio = require('fio')
+
+local function vshard_lua_path(path)
+    return string.format('%s/?.lua;%s/?/init.lua;', path, path) ..
+           (os.getenv('LUA_PATH') or '')
+end
+
+local function vshard_copy_new(sourcedir, copy_path)
+    local copy_path_load = fio.dirname(copy_path)
+    assert(fio.mkdir(copy_path_load) == true)
```
Gerold103: When a function returns a boolean, you don't need to compare its result with `true` to check if it is true. You can just write `assert(fio.mkdir(...))`. It is identical to `assert(fio.mkdir(...) == true)`.

test/luatest_helpers/upgrade.lua
Gerold103, line 1: These utilities seem to be only used in Luatest. Lets then move them into `luatest_helpers` folder and name the file just `upgrade`. Saying `...helpers/...utils` is a bit repetitive as utils and helpers are the same thing in this context.

test/luatest_helpers/vtest.lua
```diff
@@ -226,16 +236,13 @@ local function cluster_new(g, cfg)
     end
     for _, master in pairs(masters) do
         master:wait_for_readiness()
-        master:exec(function(cfg)
+        master:exec(function(cfg, test_user_grant_range)
```
Gerold103: Lets leave the original name of the variable `grant_range` in the function parameters, then your diff below will be smaller.

test/luatest_helpers/vtest.lua
```diff
@@ -387,8 +393,12 @@ local function cluster_bootstrap(g, cfg)
     local replicaset_count = 0
     local master_info, err = cluster_exec_each(g, function(is_named)
         local info = box.info
+        local is_master = ivshard.storage.internal.is_master
+        if is_master == nil then
+            is_master = box.cfg.read_only == false
+        end
```
Gerold103: Why did you add this? From what I see, the old code didn't have this logic.

test/router-luatest/reload_test.lua
```diff
@@ -29,14 +29,12 @@ local cfg_template = {
 local global_cfg
 
 g.before_all(function()
-    -- Override of the built in modules is available only since 2.11.0.
+    -- Tarantool's override loader is available since 2.11.
```
Gerold103, line 32: I suggest to keep the old comment, they seem identical.

test/router-luatest/reload_test.lua
```diff
@@ -29,14 +29,12 @@ local cfg_template = {
...
     t.run_only_if(vutil.version_is_at_least(2, 11, 0, nil, 0, 0))
     global_cfg = vtest.config_new(cfg_template)
 
     -- The test works in the following directory
-    local vardir = vtest.vardir or fio.tempdir()
-    g.vshard_copy_path_load =  vardir .. '/vshard_copy'
-    t.assert_equals(fio.mkdir(g.vshard_copy_path_load), true)
+    local vardir = fio.tempdir()
```
Gerold103: Why? Vardir of the test-run runner was intentionally supported so we could easily find the instance data files and logs. But now it will be always in a new temporary folder with an unknown name?

vshard/error.lua
```diff
@@ -294,6 +294,30 @@ local function make_error(e)
...
+       err.code ~= box.error.PROC_LUA or
+       type(err.message) ~= 'string' or
+       not err.message:match('attempt to call a nil value$') then
+        return err
+    end
+    local res = make_error(box.error.new(
+        box.error.UNSUPPORTED, 'vshard.storage._call', service_name))
+    res.prev = err
+    return res
+end
```
Gerold103, line 319: Usage of this function `wrap_storage_call_error()` combined with `wrap_storage_call_result()` looks extremely inconvenient, tbh. User is forced to pass function name twice, and the indentation level gets quite big - function called in params of another function and expecting to return some special order of result values while in some cases we expect the error message to match this strange pattern.

I don't get what is so bad about letting the old errors just be as they were + add context to them on the router about what function we tried to invoke.

Does this machinery carry any logic besides making routers report a bit nicer errors from old storages?

Gerold103 (commented):
Thank you! Very impressive test btw!

test/lua_libs/git_util.lua
kamenkremen: Fixed

test/lua_libs/upgrade_utils.lua
kamenkremen: Fixed

test/luatest_helpers/upgrade.lua
kamenkremen: Done

test/luatest_helpers/vtest.lua
kamenkremen: Done

test/luatest_helpers/vtest.lua
kamenkremen: I added this at some point to add support for a legacy vshard that didn't have `is_master`, but after changing the legacy vshard version in this test, this change became unnecessary

test/router-luatest/reload_test.lua
kamenkremen: I changed it because it was failing on catch_flaky job, which runs it many times in parallel. Changed it to keep vardir

vshard/error.lua
kamenkremen: No, that function just makes error report from old storages nicer. I don't understand how we could add context to them on the router, though. Could you elaborate, please?

test/router-luatest/reload_test.lua
kamenkremen: But that function overrides only vshard, which is not a built-in module. Isn't the original comment a little misleading?

test/router-luatest/reload_test.lua
```diff
@@ -53,12 +54,9 @@ g.before_all(function()
...
     -- or override. The last one is used here, since it's the easiest.
     --
-    g.vshard_copy_path =  vardir .. '/vshard_copy/override'
-    t.assert_equals(fio.mkdir(g.vshard_copy_path), true)
-    -- Copy source to the temporary directory
-    t.assert_equals(fio.mkdir(g.vshard_copy_path .. '/.git'), true)
-    t.assert_equals(fio.copytree(vtest.sourcedir .. '/.git',
-        g.vshard_copy_path .. '/.git'), true)
+    g.vshard_lua_path =
+        require('test.luatest_helpers.upgrade').vshard_copy_new(
```
Serpentian, line 58: Nit; it's better to globally require it, even if it's used only once

vshard/storage/init.lua
Serpentian: Let's drop the `wrap_storage_call_error`. We'll return that crutchy LuaJIT for now, but after some time, when old version becomes the one, where our change in the `service_call` is present,  then it'll become the proper error.

vshard/storage/init.lua
Serpentian: Please, elaborate, the same thing is done several lines below unconditionally. If it must be done before storage.cfg, maybe it's the case for all services and not just this one

Serpentian (commented):
Well done, we're almost there

vshard/storage/init.lua
kamenkremen: It seems that something like this happens:

  1. When hot-reloading from an old version that doesn't have master_sync_f, storage_cfg creates a master_sync_fiber.
  2. master_sync_fiber looks up its function by name, and since master_sync_f doesn't exist in M yet, it silently terminates.

  3. The final M.master_sync_f = master_sync_f provides the needed function but does not restart the dead fiber.

  After that, we have a dead handle to master_sync_fiber, so vshard thinks that the fiber is running. As a result, somewhere down the line, we get a visible MASTER_NOT_SYNCED error.

  It only shows with master_sync_f here because the other functions exist in the legacy vshard version chosen for that test. But it seems that if we chose an older version to hot-reload from, a similar error could arise with those functions too, so we probably should assign all of those functions here too.

test/router-luatest/reload_test.lua
Serpentian: Ping

test/luatest_helpers/vtest.lua
```diff
@@ -387,8 +393,9 @@ local function cluster_bootstrap(g, cfg)
     local replicaset_count = 0
     local master_info, err = cluster_exec_each(g, function(is_named)
         local info = box.info
+        local is_master = ivshard.storage.internal.is_master
         return {
-            is_master = ivshard.storage.internal.is_master,
+            is_master = is_master,
```
Serpentian: First commit, this change is not needed

vshard/storage/init.lua
Serpentian: Lol, that's bad. Nice for spotting such untrivial bug. I propose to move the whole block of the service functions before this if then. And we also need to move the `util.module_unload_functions(M)` before them, otherwise it won't be possible to unload the functions.

vshard/storage/init.lua
kamenkremen: Done

test/router-luatest/reload_test.lua
kamenkremen: Fixed

test/luatest_helpers/vtest.lua
kamenkremen: Fixed

vshard/storage/init.lua
kamenkremen: Right now, if reload fails, existing functions are not upgraded. But if we moved that block and `module_unload_functions` before this if then, couldn't the following happen: functions are replaced -> storage_cfg() fails -> reload fails, but functions in M remain updated?

I implemented a safer version: before storage_cfg(), assign a function only if it doesn't exist yet and otherwise leave the existing function unchanged. A failed reload can still leave a newly added function in M, but since the old version didn't have that function or a service using it, it probably shouldn't be a problem

test/luatest_helpers/vtest.lua
```diff
@@ -146,12 +152,15 @@ local function config_new(templ)
...
+-- Build a new cluster by a given config and optional server config.
 --
-local function cluster_new(g, cfg)
+local function cluster_new(g, cfg, server_config)
     if not g.cluster then
         g.cluster = cluster:new({})
     end
+    server_config = server_config or {}
+    local grant_range = cfg.test_user_grant_range
+    local vshard_cfg = cfg_without_test_options(cfg)
```
Serpentian, line 163: Nit: I'd propose the `cfg = cfg_without_test_options(cfg)` in order to minimize the diff

vshard/storage/init.lua
```diff
@@ -3732,7 +3744,13 @@ service_call_api = setmetatable({
 end})
 
 local function service_call(service_name, ...)
-    return service_call_api[service_name](...)
+    local service = service_call_api[service_name]
```
Serpentian, line 3747: Let's split all these fixes to their own small commits: one fix - one commit. And small unit test for each of them, if possible (e.g. in that place it's enough to check, that UNSUPPORTED error is thrown).

So that I don't ask u every time, why this change is needed)

The last commit will just have the test itself, with no changes to `init.lua`.

vshard/storage/init.lua
Serpentian: What I don't like in the current version is that it's easy to forget to place it in two places. Can we make one general table for these?

Serpentian (commented):
These are probably the last comments I have. Great work!

PR #657, opened 2026-07-07
Recovery point manager backend

### LINKED ISSUE #648: Implement a handle to inform backup agents about rebalancing
When collecting a cluster backup in a sharded cluster it's important not to do so in the middle of rebalancing. Otherwise the data might be lost (shard A backed up after bucket move to shard B, shard B backed up before bucket arrival) or duplicated.

Also when restoring from backups with xlogs to a particular point in time, the same may happen.

We need to export information about the times when rebalancing happened in cluster so that the user is able to rule out which points in time are safe to restore to and which aren't.

JIRA issue: https://jira.vk.team/browse/TNTP-7360

Serpentian:
Closes #648

### DIFF
--- a/roles/recovery-point-manager/backends/vshard-router.lua
+++ b/roles/recovery-point-manager/backends/vshard-router.lua
@@ -1,0 +1,111 @@
+local lcfg = require('vshard.cfg')
+local lrouter = require('vshard.router')
+local consts = require('vshard.consts')
+
+local BACKEND_TYPE = 'vshard-router'
+
+--------------------------------------------------------------------------------
+-- Config
+--------------------------------------------------------------------------------
+
+local function validate(cfg)
+    local ok, config = pcall(require, 'config')
+    if not ok then
+        -- Should not happen, but who knows.
+        error('the "config" module is not available')
+    end
+    if not config:is_router() then
+        error(('the "router" sharding role is required for the %q recovery ' ..
+               'point backend'):format(BACKEND_TYPE))
+    end
+    if cfg and cfg.router_name ~= nil then
+        lcfg.check_option('router_name', 'string', nil, cfg.router_name)
+    end
+end
+
+--------------------------------------------------------------------------------
+-- Backend instance
+--------------------------------------------------------------------------------
+
+--
+-- Create a cluster recovery point, tagged with opts.label, on every master via
+-- map_callrw. Returns {[replicaset_id] = {point}} on success, or nil + err on
+-- failure.
+--
+local function backend_create_point(backend, opts)
+    lcfg.check_option('opts', 'table', nil, opts)
+    -- The label is compulsory - point that cannot be referenced is useless.
+    lcfg.check_option('opts.label', 'string', nil, opts.label)
+    local call_opts = table.copy(opts)
+    call_opts.label = nil
+    return backend.router:create_cluster_recovery_point(opts.label, call_opts)
+end
+
+--
+-- The router alerts, which lead to error on recovery point creation. Any other
+-- alert the router reports is dropped.
+--
+local alert_whitelist = {
+    MISSING_MASTER = true,
+    UNREACHABLE_MASTER = true,
+    UNREACHABLE_REPLICASET = true,
+    UNKNOWN_BUCKETS = true,
+    INVALID_CFG = true,
+}
+
+--
+-- The info that will be shown under manager:info().backend. The alerts are
+-- merged into manager self alerts and pushed to the box.info.config.alerts.
+--
+local function backend_info(backend)
+    local router_info = backend.router:info()
+    -- Vshard alerts are {code, message} sequences, but the manager and the
+    -- config alert namespace expect {message = ...}, so convert them.
+    local alerts = {}
+    for _, alert in ipairs(router_info.alerts or {}) do
+        if alert_whitelist[alert[1]] then
+            table.insert(alerts, {message =
+                string.format('%s: %s', alert[1], alert[2])})
+        end
+    end
+    return {
+        backend_type = BACKEND_TYPE,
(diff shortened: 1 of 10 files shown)


roles/recovery-point-manager/backends/vshard-router.lua
```diff
@@ -0,0 +1,106 @@
...
+-- Create a cluster recovery point, tagged with opts.label, on every master via
+-- map_callrw. Returns {[replicaset_id] = {point}} on success, or nil + err on
+-- failure.
+--
+local function backend_create_point(backend, opts)
+    lcfg.check_option('opts', 'table', nil, opts)
+    -- The label is compulsory - point that cannot be referenced is useless.
+    lcfg.check_option('opts.label', 'string', nil, opts.label)
+    local label = opts.label
+    opts.label = nil
```
Gerold103: We are modifying here `opts` variable which the caller might still expect to own. Better make a non-deep copy.

roles/recovery-point-manager/backends/vshard-router.lua
```diff
@@ -0,0 +1,106 @@
...
+
+--
+-- The info that will be shown under manager:info().backend. The alerts are
+-- merged into manager self alerts and pushed to the box.info.config.alerts.
+--
+local function backend_info(backend)
+    local info = backend.router:info()
+    -- The router status reflects the overall sharding health, not the
+    -- backend's ability to create points, so it would be misleading here.
+    info.status = nil
```
Gerold103: Is it maybe a better idea to only select the needed keys rather than nullify the not needed ones? Then we have more protection against appearance of some new non-backup-relevant info keys which we would most likely forget to nullify here.

roles/recovery-point-manager/backends/vshard-router.lua
```diff
@@ -0,0 +1,106 @@
+local lcfg = require('vshard.cfg')
+local lrouter = require('vshard.router')
+
+local STATIC_ROUTER_NAME = '_static_router'
```
Gerold103: Maybe move this one to `consts` and reuse here and in `router/init.lua`?

vshard/util.lua
```diff
@@ -465,6 +465,7 @@ local feature = {
         return ok and res.replicaset ~= nil
     end)(),
     persistent_names = version_is_at_least(3, 0, 0, 'entrypoint', 0, 0),
+    recovery_point = version_is_at_least(3, 8, 0, 'entrypoint', 0, 159),
```
Gerold103: Can we make it more version agnostic and check this?
```Lua
return (box.backup and box.backup.recovery_point and box.backup.recovery_point.create) ~= nil
```

test/router-luatest/recovery_point_test.lua
```diff
@@ -0,0 +1,179 @@
...
+-- Router's create_cluster_recovery_point
+--------------------------------------------------------------------------------
+
+g.test_create_cluster_point = function(cg)
+    local before_1 = count_recovery_points(cg.replica_1_a)
+    local before_2 = count_recovery_points(cg.replica_2_a)
+
+    local res, err = cg.router:exec(function(label, opts)
+        return ivshard.router.create_cluster_recovery_point(label, opts)
+    end, {'point-1', {timeout = vtest.wait_timeout}})
```
Gerold103: No need for params of exec-function. You can get them both right inside the function itself. The label is a literal anyway, and the timeout on instances is available as `iwait_timeout`.

Gerold103 (commented):
Top as always 💪.

vshard/util.lua
Serpentian: We can, though, it won't look as nice as you've proposed:

```lua
    recovery_point = (function()
        -- Detected by explicit presence of the recovery point API rather than
        -- by a version check. Accessing box.backup before box.cfg() raises on
        -- older Tarantool (box guards fields until it is configured), so probe
        -- it under pcall; the box.backup.recovery_point table is installed
        -- before box.cfg() on a build that has the API.
        local ok, backup = pcall(function() return box.backup end)
        return ok and backup ~= nil and backup.recovery_point ~= nil
    end)(),
```

Accessing box.backup before box.cfg raises error on old Tarantools. It seems to me, that the version check is cleaner in that case. WDYT?

roles/recovery-point-manager/backends/vshard-router.lua
Serpentian: Good idea, done

roles/recovery-point-manager/backends/vshard-router.lua
Serpentian: Good, fixed

test/router-luatest/recovery_point_test.lua
Serpentian: Sure, done

roles/recovery-point-manager/backends/vshard-router.lua
Serpentian: Great idea, done

vshard/util.lua
Gerold103: The problem with version check is that it is more obscure. The API check, OTOH, explicitly shows what we are looking for.

Gerold103: You can make it even simpler:
```Lua
local ok1, ok2 = pcall(function() return box.backup.recovery_point ~= nil end)
return ok1 and ok2
```

Serpentian: Fixed. Also fixed the flakiness of the router-luatest/recovery_point_test.lua and dropped the UNSUPPORTED error, we don't need it, we already use the `box.error.UNSUPPORTED` in other cases

Gerold103 (approved):
🎬🔝🔥💪!

PR #660, opened 2026-07-09
storage: Add bucket GC pacing options

### LINKED ISSUE #659: Add independent pacing controls for bucket garbage collection
# Add independent pacing controls for bucket garbage collection

## Problem

VShard garbage-collects the source copy of a transferred bucket by repeatedly deleting tuples in transactions until the bucket is empty. `BUCKET_CHUNK_SIZE` bounds the size of each transaction and creates yield points, but garbage collection remains work-conserving: after committing one batch, it immediately starts the next one.

This limits the duration of an individual transaction but does not limit sustained garbage-collection pressure on Tarantool's transaction thread. On large buckets or busy storage nodes, source cleanup can consume most of the available transaction-thread capacity even though bucket transfer concurrency is already constrained.

The cost is especially visible for Vinyl spaces and spaces with secondary indexes, where each logical tuple deletion can require additional index and engine work. Smaller chunks improve scheduling fairness, but they do not reduce the sustained deletion rate.

VShard has controls for bucket movement concurrency, but it has no equivalent independent control for pacing removal of transferred bucket data. Reusing movement controls would also be undesirable because operators may need fast bucket transfer while deliberately spreading source cleanup over a longer period.

## Existing behavior

Bucket garbage collection deletes up to `BUCKET_CHUNK_SIZE` tuples in one transaction, commits, checks the bucket state, and continues immediately. The same constant is also used by bucket sending and receiving, so changing it for garbage collection also changes movement behavior.

`collect_bucket_garbage_interval` does not pace deletion batches, and `ERRINJ_BUCKET_GC_PAUSE` is an all-or-nothing testing mechanism rather than an operational rate control.

## Proposal

Add independent, dynamically reconfigurable bucket-GC settings:

```lua
bucket_gc_batch_size = 1000
bucket_gc_batch_delay = 0
```

- `bucket_gc_batch_size` defines the maximum number of tuples deleted in one bucket-GC transaction.
- `bucket_gc_batch_delay` defines the delay in seconds between non-empty bucket-GC deletion transactions.
- The defaults preserve existing behavior: transactions contain at most 1000 deletes and no pacing delay is applied.
- Any delay must happen after `box.commit()`, never while a transaction is open.
- Pacing should apply to both automatic garbage collection and `vshard.storage.bucket_delete_garbage()`.
- Pacing should continue across space and bucket boundaries so a sequence of small spaces or buckets cannot bypass the configured delay.
- Bucket state should be checked again after a pacing yield before more data is deleted.
- No trailing delay should be added after the final deletion transaction when there is no more data to remove.

## Rationale

A GC-specific batch size controls the longest uninterrupted deletion burst without changing bucket transfer transaction sizing. A GC-specific delay bounds sustained cleanup pressure independently from transfer admission and concurrency, allowing transferred buckets to move quickly while source cleanup proceeds at a rate appropriate for the storage workload.

A delay-based control is intentionally simple and backward compatible. More elaborate adaptive or deletes-per-second controls could be considered later, but a fixed inter-batch delay provides a predictable operational mechanism without requiring engine-specific load measurements.

## Tradeoffs

Pacing increases the time that transferred bucket data remains on the source node and can temporarily increase disk usage. Smaller batches also create more transactions and invoke the bucket-GC event trigger more frequently. These tradeoffs are opt-in because the default delay is zero and the default batch size matches current behavior.

## Related work

- #132 introduced batched garbage-collection transactions to avoid deleting an entire bucket in one transaction. This bounds transaction size but does not bound sustained cleanup throughput.
- #85 covers yielding while iterating over bucket metadata or contents. Yielding improves fairness but does not provide an operator-controlled cleanup rate.

andrejbranch:
# Add independent pacing controls for bucket garbage collection

## Summary

This change adds independent controls for bucket garbage-collection transaction size and inter-batch delay. It allows operators to reduce sustained garbage-collection pressure on Tarantool's transaction thread without reducing bucket transfer concurrency or changing transfer transaction sizing.

Closes #659 

## Motivation

Bucket garbage collection already processes tuples in bounded transactions, but it immediately begins another transaction after each commit. The existing chunking therefore limits individual transaction duration without limiting sustained cleanup pressure. The chunk-size constant is also shared with bucket sending and receiving, so it cannot be tuned specifically for source cleanup.

## Changes

- Add `bucket_gc_batch_size`, a positive integer that controls the maximum number of tuples deleted in one bucket-GC transaction.
- Add `bucket_gc_batch_delay`, a non-negative number that controls the delay in seconds between non-empty bucket-GC deletion transactions.
- Default `bucket_gc_batch_size` to 1000 and `bucket_gc_batch_delay` to 0, preserving existing behavior by default.
- Make both options dynamically reconfigurable.
- Apply the controls to automatic garbage collection and `vshard.storage.bucket_delete_garbage()`.
- Keep one pacing state across spaces and buckets so consecutive deletion transactions cannot bypass the configured delay at an iteration boundary.
- Sleep only after a successful commit, never while a transaction is open.
- Recheck bucket state after pacing yields before continuing deletion.
- Avoid a trailing delay when the preceding batch removed the final tuple.
- Document the options and their performance and disk-usage tradeoffs.

## Behavior

With the defaults, bucket garbage collection continues to delete at most 1000 tuples per transaction with no delay between transactions. Setting a smaller batch size creates more frequent commit and yield points. Setting a positive delay spreads deletion work over time while leaving bucket transfer controls unchanged.

For example:

```lua
bucket_gc_batch_size = 100
bucket_gc_batch_delay = 0.05
```

This configuration deletes at most 100 tuples in each bucket-GC transaction and waits at least 50 milliseconds before starting the next non-empty deletion transaction.

## Compatibility

The new options are optional and retain the previous effective batch size and no-delay behavior by default. Existing configurations do not need to change.

## Tradeoffs

A positive delay increases the time required to remove transferred bucket data from the source node and may temporarily increase disk usage. A smaller batch size creates more transactions and causes the `BUCKET_EVENT.GC` trigger to run more frequently. These costs are controlled explicitly by the new settings.

## Testing

- Validate default, valid, and invalid configuration values.
- Verify manual garbage collection commits one configured batch before waiting.
- Verify automatic garbage collection preserves pacing across bucket boundaries.
- Verify pacing works for MemTX and Vinyl, with and without MVCC.
- Verify exact batch-size multiples do not cause a trailing delay.
- Verify a bucket-state change during the delay prevents further deletion.
- Run the complete garbage-collector test suite, legacy unit tests, Luacheck, and Tarantool checkpatch.

### DIFF
--- a/vshard/cfg.lua
+++ b/vshard/cfg.lua
@@ -427,6 +427,14 @@ local cfg_template = {
         default = 'auto',
         enum = {'auto', 'manual', 'off'},
     },
+    bucket_gc_batch_size = {
+        type = 'positive integer', name = 'Bucket GC batch size',
+        is_optional = true, default = consts.DEFAULT_BUCKET_GC_BATCH_SIZE,
+    },
+    bucket_gc_batch_delay = {
+        type = 'non-negative number', name = 'Bucket GC batch delay',
+        is_optional = true, default = consts.DEFAULT_BUCKET_GC_BATCH_DELAY,
+    },
     collect_bucket_garbage_interval = {
         name = 'Garbage bucket collect interval', is_deprecated = true,
         reason = 'Has no effect anymore'
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ local cfg = {
     bucket_count = 10000,
     rebalancer_disbalance_threshold = 10,
     rebalancer_max_receiving = 100,
+    bucket_gc_batch_size = 1000,
+    bucket_gc_batch_delay = 0,
     sharding = {
         ['cbf06940-0790-498b-948d-042b62cf3d29'] = { -- replicaset #1
             replicas = {
@@ -128,6 +130,8 @@ local cfg = {
 * `bucket_count` total bucket count in a cluster. **It can not be changed after bootstrap!**;
 * `rebalancer_disbalance_threshold` maximal bucket disbalance percents. Disbalance for each replicaset is calculated by formula: `|etalon_bucket_count - real_bucket_count| / etalon_bucket_count * 100`.
 * `rebalancer_max_receiving` maximal bucket count that can be received in parallel by single replicaset. This count must be limited, because else, when a new replicaset is added to a cluster, the rebalancer would send to it very big amount of buckets from existing replicasets - it produces heavy load on a new replicaset to apply all these buckets.
+* `bucket_gc_batch_size` maximal number of tuples that bucket garbage collection deletes in one transaction. The default is 1000. The `BUCKET_EVENT.GC` trigger runs once per transaction, so smaller values increase its invocation rate.
+* `bucket_gc_batch_delay` delay in seconds between bucket garbage collection transactions that delete data. The default is 0, which disables the delay between batches. A positive value limits garbage collection pressure on the transaction thread, but makes removal of transferred bucket data take longer. The setting affects both automatic garbage collection and `vshard.storage.bucket_delete_garbage()`.
 
 Example of usage `rebalancer_max_receiving`:<br>
 Suppose it to be equal to 100, total bucket count is 1000 and there are
--- a/test/storage-luatest/garbage_collector_test.lua
+++ b/test/storage-luatest/garbage_collector_test.lua
@@ -160,6 +160,177 @@ test_group.test_basic = function(g)
     end)
 end
 
+--
+-- Bucket GC can limit transaction size and pause between delete batches.
+--
+test_group.test_pacing = function(g)
+    local batch_size = 10
+    local batch_delay = 0.5
+    local cfg = table.deepcopy(global_cfg)
+    cfg.bucket_gc_batch_size = batch_size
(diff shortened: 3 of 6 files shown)


vshard/cfg.lua
```diff
@@ -427,6 +427,14 @@ local cfg_template = {
         default = 'auto',
         enum = {'auto', 'manual', 'off'},
     },
+    bucket_gc_batch_size = {
```
Serpentian, line 430: We don't add the configuration options, when nobody asks for them.  If we add them, then it may happen, that nobody will use them. Moreover, they're heavily internal and difficult to use from my point of view.

So, here's the question: do you need them in your workload or do we add them because it'd be great to have such options (if so, then please, don't)? @andrejbranch

andrejbranch: To be honest I'm in the process of verifying if this GC pacing is even necessary at least in our use case. I'll update as soon as I have a better understanding.

PR #662, opened 2026-07-10
storage: resume bucket GC iterators after commits

### LINKED ISSUE #661: Resume bucket garbage collection after the preceding batch
# Resume bucket garbage collection after the preceding batch

## Problem

VShard garbage-collects a transferred bucket in bounded transactions. When one transaction reaches `BUCKET_CHUNK_SIZE`, `gc_bucket_in_space_xc()` commits it and jumps back to `restart`, where it opens a new iterator at `{bucket_id}`. The next transaction therefore starts scanning the bucket secondary index from the bucket prefix again instead of continuing after the last tuple deleted by the preceding transaction.

This is particularly expensive for large Vinyl buckets. Deleted index entries remain as LSM history and tombstones until compaction, and a Vinyl secondary-index iterator performs primary-index lookups to materialize and validate tuples. Each new batch can revisit an increasingly large deleted prefix before reaching the next live tuple. The resulting prefix work grows with both bucket cardinality and the number of batches, making smaller garbage-collection transactions disproportionately expensive.

The sender does not have the same behavior: it keeps one iterator while emitting multiple transfer chunks. The repeated-prefix scan is specific to bucket garbage collection across transaction boundaries.

## Existing behavior

The current loop is equivalent to:

```lua
::restart::
box.begin()
for _, tuple in bucket_index:pairs({bucket_id}) do
    space:delete(extract_primary_key(tuple))
    if the_batch_is_full then
        box.commit()
        goto restart
    end
end
box.commit()
```

For a bucket containing `N` tuples and a batch size `B`, this creates approximately `N / B` iterators at the same prefix. On an LSM index, the repeatedly encountered deleted prefix can make the amount of scan work grow approximately like `N² / (2B)`, depending on the engine state and compaction history.

## Proposal

When supported by Tarantool and the bucket index is a TREE index, retain the last tuple returned by the bucket iterator and pass it through the `after` iterator option when opening the next transaction:

```lua
local after = nil
::restart::
box.begin()
for _, tuple in bucket_index:pairs({bucket_id}, {after = after}) do
    space:delete(extract_primary_key(tuple))
    if the_batch_is_full then
        after = tuple
        box.commit()
        goto restart
    end
end
box.commit()
```

The tuple used as the position was deleted by the preceding transaction, but Tarantool pagination explicitly supports an `after` tuple that is no longer present in the index. Bucket state continues to be checked after each commit as it is today.

Tarantool added index pagination in 2.11, while VShard supports Tarantool 1.10. The optimization must therefore be feature-gated. Tarantool 2.11 and later can use the resumable cursor for TREE indexes; older versions and unsupported index types must retain the current prefix-restarting behavior.

## Correctness considerations

A bucket being garbage-collected is no longer writable through VShard. Cursor resumption relies on the same lifecycle invariant as bucket sending: tuples are not inserted into the bucket behind an iterator that has already passed their position. Direct writes that bypass VShard bucket-state checks are outside this invariant.

The implementation should preserve the current behavior when a transaction fails, when the bucket generation or state changes after a commit, and when pagination is unavailable. The cursor is local to one bucket and one space and is discarded when the function returns or raises.

## Benchmark

A representative local benchmark used a 50,000-tuple Vinyl bucket with three indexes, synchronous replication, and a garbage-collection batch size of 100. Values are medians across three fresh-cluster runs.

| Iteration behavior | Cleanup wall time | Cleanup CPU time |
|---|---:|---:|
| Restart from bucket prefix | 4.906 s | 4.757 s |
| Resume after last tuple | 0.602 s | 0.414 s |

The cursor reduced cleanup wall time by approximately 88% and cleanup CPU by approximately 91% without changing the number of tuple deletes, WAL rows, transactions, index updates, or synchronous commits.

Latency percentiles measured only for the duration of each cleanup are intentionally omitted from the table. Cursor cleanup finished in approximately 0.6 seconds and collected about 225 probe observations, while prefix-restarting cleanup ran for approximately 4.9 seconds and collected about 1,815 observations. Comparing percentiles from those unequal exposure periods is misleading: the cursor p99 is approximately the third-highest observation, whereas the prefix-restart p99 is approximately the nineteenth-highest observation.

A separate comparison kept direct source probes running for the same eight-second wall-clock window after starting cleanup. Both cases deleted the same 50,000-tuple bucket, and each case collected approximately 3,300 probe observations. Values are medians across three fresh-cluster runs.

| Iteration behavior | Cleanup duration | Probe average | Probe p50 | Probe p95 | Probe p99 |
|---|---:|---:|---:|---:|---:|
| Restart from bucket prefix | 4.906 s | 0.118 ms | 0.089 ms | 0.234 ms | 0.632 ms |
| Resume after last tuple | 0.501 s | 0.085 ms | 0.057 ms | 0.217 ms | 0.582 ms |

Over the equal wall-clock window, cursor resumption did not regress the direct probe p99 and improved the average, p50, p95, and p99. The p99 difference was modest, so these results should be interpreted as evidence against a material regression rather than as a precise latency improvement. These direct probes measure TX-thread responsiveness and are not a substitute for end-to-end client latency measurements.

In a separate batch-size matrix, cursor-based cleanup took 0.543, 0.511, 0.495, and 0.483 seconds for batch sizes 100, 250, 500, and 1000 respectively. Without the cursor, the same cases took 4.999, 2.301, 1.407, and 0.970 seconds. Resuming the iterator made cleanup throughput nearly independent of the transaction batch size while retaining smaller transactions as the control for the TX scheduling quantum.

## Testing

- Verify that Tarantool 2.11 and later pass the last deleted tuple through `after` on subsequent garbage-collection batches.
- Verify that the first batch starts at the bucket prefix.
- Verify that the fallback path never passes `after` when pagination is unavailable.
- Verify complete deletion with MemTX and Vinyl, with and without MVCC.
- Verify that a deleted cursor tuple is accepted as the next iterator position.
- Run the complete garbage-collector suite on supported Tarantool versions, including the Tarantool 1.10 compatibility job.

## Related work

- #132 introduced bounded garbage-collection transactions. Preserving iterator position completes that design by avoiding repeated work between those transactions.
- #85 addresses yielding during iteration but does not preserve scan progress across garbage-collection transactions.

andrejbranch:
# Resume bucket GC iteration across transaction batches

## Summary

This change avoids restarting bucket garbage-collection scans from the bucket prefix after every transaction. On Tarantool 2.11 and later, GC over a TREE bucket index now retains the last tuple committed by a full batch and passes it as the next iterator's `after` position. Older Tarantool versions and unsupported index types retain the existing prefix-restarting path.

Closes #661 

## Motivation

Bucket GC commits after `BUCKET_CHUNK_SIZE` deletes and then opens another iterator at `{bucket_id}`. For large Vinyl buckets, every new secondary-index iterator can revisit the tombstones and LSM history left by preceding batches before it reaches the next live tuple. Smaller transaction batches create more restarts and can therefore cause disproportionate cleanup CPU and wall time.

Tarantool pagination supports resuming after a tuple that is no longer present in the index. Retaining the last committed tuple allows the next transaction to continue at the correct secondary-index position without changing delete, WAL, replication, transaction, trigger, or bucket-state semantics.

## Changes

- Add an `index_pagination` core feature flag for Tarantool 2.11 and later.
- Resume bucket GC from the last tuple committed by the preceding full batch when the sharding index is a TREE index and pagination is supported.
- Preserve the current prefix-restarting behavior on Tarantool 1.10 and unsupported index types.
- Keep the cursor local to one bucket and one space.
- Continue checking the bucket generation and state after every full-batch commit.
- Add tests that verify the initial prefix scan, subsequent `after` positions, the pagination-disabled fallback, and the non-TREE fallback.

## Compatibility

There is no configuration or public API change. Tarantool versions before 2.11 never receive the `after` iterator option and retain the current implementation. The optimization is also disabled for non-TREE sharding indexes.

A garbage bucket is not writable through VShard while cleanup runs. Cursor resumption relies on the same lifecycle invariant as bucket sending: no tuples are inserted behind an iterator that has already passed their position.

## Performance

A representative local benchmark used a 50,000-tuple Vinyl bucket with three indexes, synchronous replication, and a garbage-collection transaction size of 100. Values are medians across three fresh-cluster runs.

| Iteration behavior | Cleanup wall time | Cleanup CPU time |
|---|---:|---:|
| Restart from bucket prefix | 4.906 s | 4.757 s |
| Resume after last tuple | 0.602 s | 0.414 s |

The optimization reduced cleanup wall time by approximately 88% and cleanup CPU by approximately 91%. It did not change the number of deletes, WAL rows, transactions, index updates, or synchronous commits.

## Testing

- New resumable-cursor test passes for MemTX and Vinyl, with and without MVCC.
- The test verifies the pagination-enabled cursor positions and forces the pagination-disabled fallback.
- The test verifies that a HASH sharding index retains the non-cursor path.
- Complete garbage-collector suite: 46 passed, 2 skipped.
- Luacheck 0.26.1: 0 warnings and 0 errors across 61 files.

### DIFF
--- a/test/storage-luatest/garbage_collector_test.lua
+++ b/test/storage-luatest/garbage_collector_test.lua
@@ -1074,3 +1074,156 @@ test_group.test_route_map_is_cleared = function(g)
     vtest.cluster_wait_vclock_all(g)
     g.replica_1_b:exec(bucket_set_protection, {true})
 end
+
+--
+-- gh-661: bucket GC resumes after the last tuple deleted by the preceding
+-- transaction on Tarantool versions with index pagination. Older versions and
+-- indexes that do not support positions keep restarting from the bucket prefix.
+--
+local function test_resumable_cursor_template(g, expected_resumes)
+    g.replica_1_a:exec(function(expected_resumes)
+        _G.bucket_gc_pause()
+        local s = box.space.test
+        local bucket_index = s.index.bucket_id
+        local bid = _G.get_first_bucket()
+        local old_chunk_size = ivconst.BUCKET_CHUNK_SIZE
+        local old_pairs = bucket_index.pairs
+        local chunk_size = 10
+        local tuple_count = chunk_size * 2 + 5
+
+        -- Two full chunks and a non-empty tail make GC open three iterators.
+        -- Record whether each one resumed from a position (1) or restarted at
+        -- the bucket prefix (0). The first iterator always starts at the
+        -- prefix. The next two resume only on cores with index pagination.
+        -- 'after' is opaque, so its presence is the observable signal.
+        local resumes = {}
+        bucket_index.pairs = function(self, key, opts)
+            local after = opts and opts.after
+            table.insert(resumes, after ~= nil and 1 or 0)
+            return old_pairs(self, key, opts)
+        end
+
+        local ok, err = pcall(function()
+            ivconst.BUCKET_CHUNK_SIZE = chunk_size
+            box.begin()
+            for i = 1, tuple_count do
+                s:replace{100000 + i, bid}
+            end
+            box.commit()
+            ivshard.storage.bucket_delete_garbage(bid, {force = true})
+            ilt.assert_equals(bucket_index:count({bid}), 0)
+            ilt.assert_equals(resumes, expected_resumes)
+        end)
+
+        bucket_index.pairs = old_pairs
+        ivconst.BUCKET_CHUNK_SIZE = old_chunk_size
+        _G.bucket_gc_continue()
+        ilt.assert(ok, err)
+    end, {expected_resumes})
+    vtest.cluster_wait_vclock_all(g)
+end
+
+--
+-- gh-661: pagination-capable cores resume the second and third iterators after
+-- the preceding chunk instead of restarting at the bucket prefix.
+--
+test_group.test_resumable_cursor_supported = function(g)
+    t.run_only_if(vutil.feature.index_pagination)
+    test_resumable_cursor_template(g, {0, 1, 1})
+end
+
+--
+-- gh-661: without index pagination (Tarantool < 2.11) every iterator restarts
+-- at the bucket prefix.
+--
+test_group.test_resumable_cursor_unsupported = function(g)
+    t.run_only_if(not vutil.feature.index_pagination)
+    test_resumable_cursor_template(g, {0, 0, 0})
+end
+
+--
+-- gh-661: functional and multikey sharding indexes report type == 'TREE' and
+-- pass find_sharded_spaces(), but reject positions with "does not support
+-- position by tuple"; hash indexes are not TREE at all. GC must fall back to
+-- restarting from the bucket prefix for them and still delete the bucket.
+--
+test_group.test_resumable_cursor_unsupported_index = function(g)
+    t.run_only_if(vutil.feature.index_pagination)
+    g.replica_1_a:exec(function()
+        _G.bucket_gc_pause()
+        local bid = _G.get_first_bucket()
+        local spaces = {}
+        local func_name = 'test_extract_bucket'
+        local old_chunk_size = ivconst.BUCKET_CHUNK_SIZE
+
+        -- Each case fills exactly one chunk, so GC has to open a second
+        -- iterator after committing the delete. Cleanup succeeds only if an
+        -- index which cannot derive a position from a tuple falls back to
+        -- restarting at the bucket prefix.
+        local function run(name, space_opts, index_opts, tuple)
+            local space = box.schema.create_space(name, space_opts)
+            table.insert(spaces, space)
+            space:create_index('pk')
+            space:create_index('bucket_id', index_opts)
+            space:replace(tuple)
+            ivshard.storage.bucket_delete_garbage(bid, {force = true})
+            ilt.assert_equals(space:count(), 0)
+            space:drop()
+            spaces[#spaces] = nil
+        end
+
+        local ok, err = pcall(function()
+            ivconst.BUCKET_CHUNK_SIZE = 1
+
+            run('test_hash', {
+                engine = 'memtx',
+                format = {{'id', 'unsigned'}, {'bid', 'unsigned'}},
+            }, {
+                type = 'hash', unique = true, parts = {2},
+            }, {300000, bid})
+
+            run('test_multikey', {
+                engine = 'vinyl',
+                format = {
+                    {'id', 'unsigned'},
+                    {'bid', 'unsigned'},
+                    {'tags', 'array'},
+                },
+            }, {
+                unique = false,
+                parts = {
+                    {field = 2, type = 'unsigned'},
+                    {field = 3, type = 'string', path = '[*]'},
+                },
+            }, {300001, bid, {'tag'}})
+
+            -- Memtx MVCC does not support functional indexes.
+            if not box.cfg.memtx_use_mvcc_engine then
+                box.schema.func.create(func_name, {
+                    body = [[function(tuple) return {tuple[2]} end]],
+                    is_deterministic = true,
+                    is_sandboxed = true,
+                })
+                run('test_functional', {
+                    engine = 'memtx',
+                    format = {{'id', 'unsigned'}, {'bid', 'unsigned'}},
+                }, {
+                    unique = false,
+                    func = func_name,
+                    parts = {{field = 1, type = 'unsigned'}},
+                }, {300002, bid})
+            end
+        end)
+
+        ivconst.BUCKET_CHUNK_SIZE = old_chunk_size
+        for _, space in ipairs(spaces) do
+            space:drop()
+        end
+        if box.func[func_name] ~= nil then
+            box.schema.func.drop(func_name)
+        end
+        _G.bucket_gc_continue()
+        ilt.assert(ok, err)
+    end)
+    vtest.cluster_wait_vclock_all(g)
+end
--- a/vshard/storage/init.lua
+++ b/vshard/storage/init.lua
@@ -2351,20 +2351,26 @@ end
 --
 local function gc_bucket_in_space_xc(space, bucket_id, status)
     local bucket_index = space.index[lschema.shard_index]
-    if #bucket_index:select({bucket_id}, {limit = 1}) == 0 then
+    local first = bucket_index:select({bucket_id}, {limit = 1})[1]
+    if first == nil then
         return
     end
     local bucket_generation = M.bucket_generation
     local pk_parts = space.index[0].parts
+    local use_cursor = util.index_can_paginate(bucket_index, first)
+    local after
 ::restart::
     local limit = consts.BUCKET_CHUNK_SIZE
     box.begin()
     M._on_bucket_event:run(consts.BUCKET_EVENT.GC, bucket_id,
                            {spaces = {space.name}})
-    for _, tuple in bucket_index:pairs({bucket_id}) do
+    for _, tuple in bucket_index:pairs({bucket_id}, {after = after}) do
         space:delete(util.tuple_extract_key(tuple, pk_parts))
         limit = limit - 1
         if limit == 0 then
+            if use_cursor then
+                after = bucket_index:tuple_pos(tuple)
+            end
             box.commit()
             bucket_generation =
                 bucket_guard_xc(bucket_generation, bucket_id, status)
--- a/vshard/util.lua
+++ b/vshard/util.lua
@@ -438,6 +438,7 @@ end
 local feature = {
     msgpack_object = lmsgpack.object ~= nil,
     netbox_return_raw = version_is_at_least(2, 10, 0, 'beta', 2, 86),
+    index_pagination = version_is_at_least(2, 11, 0, nil, 0, 0),
     multilisten = luri.parse_many ~= nil,
     -- It appeared earlier but at 2.9.0 there is a strange bug about an immortal
     -- tuple - after :delete(pk) it still stays in the space. That makes the
@@ -473,6 +474,19 @@ local feature = {
         return ok1 and ok2
     end)(),
 }
+
+--
+-- Check whether an index supports pagination using a tuple position. Pagination
+-- is a TREE index feature, but functional and multikey TREE indexes reject
+-- tuple positions. Probe the index so callers can safely fall back.
+--
+local function index_can_paginate(index, tuple)
+    if not feature.index_pagination or index.type ~= 'TREE' then
+        return false
+    end
+    local ok = pcall(index.tuple_pos, index, tuple)
+    return ok
+end
 
 local schema_version = function()
     return box.info.schema_version
@@ -587,6 +601,7 @@ return {
     fiber_is_self_canceled = fiber_is_self_canceled,
     index_min = index_min,
     index_has = index_has,
+    index_can_paginate = index_can_paginate,
     feature = feature,
     schema_version = schema_version,
     replicaset_uuid = replicaset_uuid,


vshard/storage/init.lua
```diff
@@ -2356,15 +2357,32 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
...
     M._on_bucket_event:run(consts.BUCKET_EVENT.GC, bucket_id,
                            {spaces = {space.name}})
-    for _, tuple in bucket_index:pairs({bucket_id}) do
+    local gen, param, state
+    if after == nil then
+        gen, param, state = bucket_index:pairs({bucket_id})
+    else
+        gen, param, state = bucket_index:pairs({bucket_id}, {after = after})
+    end
+    for _, tuple in gen, param, state do
```
ochaton: since the default value for `after` is `nil` then `{after=after}` is evaluated to `{after=nil}` and it's the same as `{}`
you could simply write this:

```lua
-- ...
local after = nil
::restart::

local limit = consts.BUCKET_CHUNK_SIZE
box.begin()
M._on_bucket_event:run(consts.BUCKET_EVENT.GC, bucket_id,
                       {spaces = {space.name}})

for _, tuple in bucket_index:pairs({bucket_id}, { after = after }) do
    -- ....
    if limit == 0 then
        if use_cursor then
            after = tuple
        end
        -- ...
    end
    -- ....
end
```

so it's will be just one-line fix

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -160,6 +160,89 @@ test_group.test_basic = function(g)
     end)
```
Serpentian, line 160: Please, add issue number to the commit msg like that:

```
Closes #661 
```

vshard/storage/init.lua
```diff
@@ -2356,15 +2357,28 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
...
                            {spaces = {space.name}})
-    for _, tuple in bucket_index:pairs({bucket_id}) do
+    -- 'after' is nil on the first pass, where {after = nil} is equivalent to
+    -- {}, and the last committed tuple on resumes.
+    for _, tuple in bucket_index:pairs({bucket_id}, {after = after}) do
         space:delete(util.tuple_extract_key(tuple, pk_parts))
         limit = limit - 1
         if limit == 0 then
+            if use_cursor then
+                after = tuple
```
Serpentian: Using tuple is dangerous due to bug we had with it: https://github.com/tarantool/tarantool/issues/11221. Let's instead use the `bucket_index:tuple_pos(tuple)`. You can check the perf test, I've added, in the issue you've created

vshard/storage/init.lua
```diff
@@ -2356,15 +2357,28 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
...
+                       bucket_index.type == 'TREE'
+    local after = nil
 ::restart::
     local limit = consts.BUCKET_CHUNK_SIZE
     box.begin()
     M._on_bucket_event:run(consts.BUCKET_EVENT.GC, bucket_id,
                            {spaces = {space.name}})
-    for _, tuple in bucket_index:pairs({bucket_id}) do
+    -- 'after' is nil on the first pass, where {after = nil} is equivalent to
+    -- {}, and the last committed tuple on resumes.
```
Serpentian: The comment is not needed, the code is pretty self-explanatory

vshard/storage/init.lua
```diff
@@ -2356,15 +2357,28 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
...
     local bucket_generation = M.bucket_generation
     local pk_parts = space.index[0].parts
+    -- A new Vinyl iterator started at the bucket prefix has to walk past
+    -- tombstones left by all preceding batches. Resume from the last tuple
+    -- when the core supports pagination. The tuple itself is deleted, but
+    -- 'after' positions are valid even when their tuple is no longer present.
+    -- Pagination is supported by TREE indexes only.
+    local use_cursor = util.feature.index_pagination and
+                       bucket_index.type == 'TREE'
+    local after = nil
```
Serpentian: No need to initialize with `nil`, it's already `nil` with just `local after`

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -160,6 +160,89 @@ test_group.test_basic = function(g)
...
 end
 
+--
+-- On Tarantool versions with index pagination, bucket GC resumes after the
+-- last tuple deleted by the preceding transaction. Older versions keep
+-- restarting from the bucket prefix.
+--
+test_group.test_resumable_cursor = function(g)
```
Serpentian: Nit: let's add the test to the end of the file

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -160,6 +160,89 @@ test_group.test_basic = function(g)
     end)
 end
 
+--
+-- On Tarantool versions with index pagination, bucket GC resumes after the
```
Serpentian: Nit: ticket number here, see the other tests as example:

```
--
-- gh-434: bucket_recv() shouldn't change the transfer flags if a transfer for
-- the same bucket is already in progress.
--
```

vshard/storage/init.lua
```diff
@@ -2356,15 +2357,28 @@ local function gc_bucket_in_space_xc(space, bucket_id, status)
     end
     local bucket_generation = M.bucket_generation
     local pk_parts = space.index[0].parts
+    -- A new Vinyl iterator started at the bucket prefix has to walk past
+    -- tombstones left by all preceding batches. Resume from the last tuple
+    -- when the core supports pagination. The tuple itself is deleted, but
+    -- 'after' positions are valid even when their tuple is no longer present.
+    -- Pagination is supported by TREE indexes only.
+    local use_cursor = util.feature.index_pagination and
+                       bucket_index.type == 'TREE'
```
Serpentian: This check is not enough. functional and multikey sharding indexes still have `type == 'TREE'`, but `index:pairs(..., {after = tuple})` rejects them with “does not support position by tuple”. `find_sharded_spaces()` accepts such numeric first-part indexes, so once a bucket exceeds `BUCKET_CHUNK_SIZE`, GC fails on the second batch and retries indefinitely without dropping the bucket.

<details><summary>TREE indexes which don't allow pagination</summary>

```lua
-- Exhaustive examples of TREE index categories for which Tarantool rejects
-- `after = tuple`. Each index has an unsigned first part, so VShard's
-- find_sharded_spaces() accepts it as a sharding index.

local fio = require('fio')

local data_dir = fio.tempdir()
box.cfg({
    log = fio.pathjoin(data_dir, 'tarantool.log'),
    memtx_dir = data_dir,
    wal_dir = data_dir,
    wal_mode = 'none',
})

local function check(name, space, index, prefix, expected)
    assert(index.type == 'TREE')
    assert(index.parts[1].type == 'unsigned')
    local tuple = index:select(prefix, {limit = 1})[1]
    assert(tuple ~= nil)
    space:delete({tuple[1]})
    assert(space:get({tuple[1]}) == nil)

    local pairs_ok, pairs_err = pcall(index.pairs, index, prefix,
                                      {after = tuple})
    local pos_ok, pos_err = pcall(index.tuple_pos, index, tuple)
    assert(not pairs_ok and tostring(pairs_err):find(expected, 1, true))
    assert(not pos_ok and tostring(pos_err):find(expected, 1, true))
    print(('%s: pairs=%s; tuple_pos=%s'):
          format(name, tostring(pairs_err), tostring(pos_err)))
end

-- 1. A regular JSON-path multikey index. The bucket part is scalar and a
-- later key part is multikey. This is the most plausible accidental VShard
-- schema: bucket prefix scans work normally until `after = tuple` is used.
local s1 = box.schema.space.create('json_multikey_later_part')
s1:create_index('primary')
local i1 = s1:create_index('bucket_id', {
    type = 'TREE',
    unique = false,
    parts = {
        {field = 2, type = 'unsigned'},
        {field = 3, type = 'string', path = '[*]'},
    },
})
s1:insert({1, 10, {'tag'}})
check('json_multikey_later_part', s1, i1, {10},
      'Multikey index does not support position by tuple')

-- 2. A regular JSON-path multikey index whose first (bucket) part itself is
-- multikey. VShard discovery accepts it because it checks only the part type,
-- although assigning one tuple to multiple buckets violates VShard's normal
-- data model.
local s2 = box.schema.space.create('json_multikey_bucket_part')
s2:create_index('primary')
local i2 = s2:create_index('bucket_id', {
    type = 'TREE',
    unique = false,
    parts = {
        {field = 2, type = 'unsigned', path = '[*]'},
    },
})
s2:insert({1, {10}})
check('json_multikey_bucket_part', s2, i2, {10},
      'Multikey index does not support position by tuple')

-- 3. A single-key functional index. Its declared parts describe fields in the
-- key returned by the function, so the unsigned first part passes VShard
-- discovery.
box.schema.func.create('extract_bucket', {
    body = [[function(tuple) return {tuple[2]} end]],
    is_deterministic = true,
    is_sandboxed = true,
})
local s3 = box.schema.space.create('functional')
s3:create_index('primary')
local i3 = s3:create_index('bucket_id', {
    type = 'TREE',
    unique = false,
    func = 'extract_bucket',
    parts = {{field = 1, type = 'unsigned'}},
})
s3:insert({1, 10})
check('functional', s3, i3, {10},
      'Functional index does not support position by tuple')

-- 4. A functional multikey index. It has both unsupported properties, but the
-- functional-index check runs first, so Tarantool reports it as functional.
box.schema.func.create('extract_bucket_multikey', {
    body = [[function(tuple) return {{tuple[2]}} end]],
    is_deterministic = true,
    is_sandboxed = true,
    is_multikey = true,
})
local s4 = box.schema.space.create('functional_multikey')
s4:create_index('primary')
local i4 = s4:create_index('bucket_id', {
    type = 'TREE',
    unique = false,
    func = 'extract_bucket_multikey',
    parts = {{field = 1, type = 'unsigned'}},
})
s4:insert({1, 10})
check('functional_multikey', s4, i4, {10},
      'Functional index does not support position by tuple')

os.exit(0)
```

</details>

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -160,6 +160,89 @@ test_group.test_basic = function(g)
...
+        local util = require('vshard.util')
+        local old_chunk_size = ivconst.BUCKET_CHUNK_SIZE
+        local old_pagination = util.feature.index_pagination
+        local old_pairs = bucket_index.pairs
+        local chunk_size = 10
+        local tuple_count = chunk_size * 2 + 5
+        local hash_space
+
+        local function run(use_cursor, id_base)
+            util.feature.index_pagination = use_cursor
```
Serpentian: Could we extract the common setup into test_resumable_cursor_template() and define separate feature-gated tests? Pagination availability is fixed by the Tarantool version, so overriding util.feature.index_pagination creates an artificial runtime state and makes the test harder to follow.

 On pagination-capable cores, the template should expect {0, 100010, 100020}; otherwise it should expect {0, 0, 0}. The non-TREE fallback can remain a separate test gated on pagination availability.


<details><summary>Smth like that (LLM generated, but way more readable)</summary>
<p>

```lua
local function test_resumable_cursor_template(g, expected_positions)
    g.replica_1_a:exec(function(expected_positions)
        _G.bucket_gc_pause()
        local s = box.space.test
        local bucket_index = s.index.bucket_id
        local bid = _G.get_first_bucket()
        local old_chunk_size = ivconst.BUCKET_CHUNK_SIZE
        local old_pairs = bucket_index.pairs
        local chunk_size = 10
        local tuple_count = chunk_size * 2 + 5

        -- Two full chunks and a non-empty tail make GC open three iterators.
        -- Record the position passed to each iterator. The first one always
        -- starts at the bucket prefix. Subsequent iterators either resume from
        -- the last tuple of the preceding chunk or restart at the prefix when
        -- pagination is unavailable.
        local function run()
            ivconst.BUCKET_CHUNK_SIZE = chunk_size
            box.begin()
            for i = 1, tuple_count do
                s:replace{100000 + i, bid}
            end
            box.commit()

            local positions = {}
            bucket_index.pairs = function(self, key, opts)
                local after = opts and opts.after
                -- Tuple IDs are positive, so 0 denotes a prefix restart.
                table.insert(positions, after and after.id or 0)
                if opts == nil then
                    return old_pairs(self, key)
                end
                return old_pairs(self, key, opts)
            end
            local ok, err = pcall(
                ivshard.storage.bucket_delete_garbage, bid, {force = true})
            bucket_index.pairs = old_pairs
            ilt.assert(ok, err)
            ilt.assert_equals(bucket_index:count({bid}), 0)
            return positions
        end

        local ok, err = pcall(function()
            ilt.assert_equals(run(), expected_positions)
        end)
        bucket_index.pairs = old_pairs
        ivconst.BUCKET_CHUNK_SIZE = old_chunk_size
        _G.bucket_gc_continue()
        ilt.assert(ok, err)
    end, {expected_positions})
    vtest.cluster_wait_vclock_all(g)
end

test_group.test_resumable_cursor_supported = function(g)
    t.run_only_if(vutil.feature.index_pagination)
    test_resumable_cursor_template(g, {0, 100010, 100020})
end

test_group.test_resumable_cursor_unsupported = function(g)
    t.run_only_if(not vutil.feature.index_pagination)
    test_resumable_cursor_template(g, {0, 0, 0})
end

test_group.test_resumable_cursor_unsupported_index = function(g)
    t.run_only_if(vutil.feature.index_pagination)
    g.replica_1_a:exec(function()
        local bid = _G.get_first_bucket()
        local spaces = {}
        local func_name = 'test_extract_bucket'
        local old_chunk_size = ivconst.BUCKET_CHUNK_SIZE

        -- Each case fills exactly one chunk, so GC has to open a second
        -- iterator after committing the delete. Cleanup succeeds only if an
        -- index which cannot derive an 'after' position from a tuple falls back
        -- to restarting at the bucket prefix.
        local function run(name, space_opts, index_opts, tuple)
            local space = box.schema.create_space(name, space_opts)
            table.insert(spaces, space)
            space:create_index('pk')
            space:create_index('bucket_id', index_opts)
            space:replace(tuple)
            ivshard.storage.bucket_delete_garbage(bid, {force = true})
            ilt.assert_equals(space:count(), 0)
            space:drop()
            spaces[#spaces] = nil
        end

        local ok, err = pcall(function()
            ivconst.BUCKET_CHUNK_SIZE = 1

            run('test_hash', {
                engine = 'memtx',
                format = {{'id', 'unsigned'}, {'bid', 'unsigned'}},
            }, {
                type = 'hash', unique = true, parts = {2},
            }, {300000, bid})

            run('test_multikey', {
                engine = 'vinyl',
                format = {
                    {'id', 'unsigned'},
                    {'bid', 'unsigned'},
                    {'tags', 'array'},
                },
            }, {
                unique = false,
                parts = {
                    {field = 2, type = 'unsigned'},
                    {field = 3, type = 'string', path = '[*]'},
                },
            }, {300001, bid, {'tag'}})

            box.schema.func.create(func_name, {
                body = [[function(tuple) return {tuple[2]} end]],
                is_deterministic = true,
                is_sandboxed = true,
            })
            run('test_functional', {
                engine = 'memtx',
                format = {{'id', 'unsigned'}, {'bid', 'unsigned'}},
            }, {
                unique = false,
                func = func_name,
                parts = {{field = 1, type = 'unsigned'}},
            }, {300002, bid})
        end)
        ivconst.BUCKET_CHUNK_SIZE = old_chunk_size
        for _, space in ipairs(spaces) do
            space:drop()
        end
        if box.func[func_name] ~= nil then
            box.schema.func.drop(func_name)
        end
        ilt.assert(ok, err)
    end)
    vtest.cluster_wait_vclock_all(g)
end
```

</details

Serpentian (commented):
Great work digging into the iterator behavior, measuring its impact, and carrying the solution. This is an impressive first patch!

vshard/storage/init.lua
andrejbranch: Great catch, confirmed all four categories from your script reject positions. The gate now probes the index once (pcall(bucket_index.tuple_pos, ...)) and falls back to the prefix-restart scan when it fails, so functional/multikey TREE indexes (and hash) keep working instead of looping on the second batch. Covered by test_resumable_cursor_unsupported_index, which exercises hash + multikey + functional and asserts the bucket still drains completely

test/storage-luatest/garbage_collector_test.lua
andrejbranch: Refactored to your structure: test_resumable_cursor_template() + version-gated test_resumable_cursor_supported / _unsupported (run_only_if on vutil.feature.index_pagination), plus _unsupported_index for the non-paginating indexes. No more overriding util.feature.index_pagination at runtime. One deviation from your draft: since we now pass tuple_pos() (an opaque position) rather than the tuple, the recorder can't read after.id, so it asserts restart-vs-resume ({0, 1, 1} supported / {0, 0, 0} unsupported) instead of the exact tuple ids. Same intent, first iterator at the prefix, the next two resuming just compatible with positions.

vshard/storage/init.lua
andrejbranch: Switched to after = bucket_index:tuple_pos(tuple). This is also the faster "position" mode from your benchmark, so it's a nice win on both counts. Thanks for the pointer to #11221.

vshard/util.lua
```diff
@@ -438,6 +438,8 @@ end
 local feature = {
     msgpack_object = lmsgpack.object ~= nil,
     netbox_return_raw = version_is_at_least(2, 10, 0, 'beta', 2, 86),
+    -- The 'after' iterator option was added in Tarantool 2.11.
```
Serpentian: Nit: let's drop the comment, obvious from the code

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -1074,3 +1074,156 @@ test_group.test_route_map_is_cleared = function(g)
     vtest.cluster_wait_vclock_all(g)
     g.replica_1_b:exec(bucket_set_protection, {true})
 end
+
```
Serpentian, line 1077: The newly added test fails on some versions, let's fix that and also let's fix the nits from `checkpatch`

Serpentian (commented):
We're almost there, let's fix the CI and please, share the perf test source u did (preferably, in the issue, not in the PR)

vshard/storage/init.lua
```diff
@@ -2351,20 +2351,36 @@ end
...
+    -- tombstones left by all preceding batches, which is expensive on Vinyl.
+    -- Resume from the position of the last committed tuple instead. The tuple
+    -- is deleted, but its position stays valid once the tuple is gone.
+    -- Pagination is a TREE-index feature, but functional and multikey TREE
+    -- indexes still reject positions with "does not support position by tuple"
+    -- even though find_sharded_spaces() accepts them. Probe the index once and
+    -- fall back to restarting from the bucket prefix when it is unsupported.
+    local use_cursor = util.feature.index_pagination and
+                       bucket_index.type == 'TREE' and
+                       pcall(bucket_index.tuple_pos, bucket_index, first)
```
Gerold103: I like the compatibility taken into account! This seems working. I would suggest to make the usage easier though. Imagine that we will have more than one place where this kind of functionality would be needed? Then it would be quite inconvenient to write this code block again and all these checks and the huge comment.

Perhaps there is a way to encapsulate the entire thing inside `util`. So usage would look like this:
```Lua
local use_cursor = util.index_can_paginate(bucket_index, first)
```

And `util.index_can_paginate()` internally would do these checks with the detailed comment.

What do you think?

test/storage-luatest/garbage_collector_test.lua
```diff
@@ -1074,3 +1074,159 @@ test_group.test_route_map_is_cleared = function(g)
...
+        -- prefix. The next two resume only on cores with index pagination.
+        -- 'after' is opaque, so its presence is the observable signal.
+        local resumes = {}
+        bucket_index.pairs = function(self, key, opts)
+            local after = opts and opts.after
+            table.insert(resumes, after ~= nil and 1 or 0)
+            if opts == nil then
+                return old_pairs(self, key)
+            end
+            return old_pairs(self, key, opts)
```
Gerold103: This shouldn't be needed. When `opts` is nil, calling `old_pairs(self, key, opts)` is equivalent to calling `old_pairs(self, key)`.

Gerold103 (commented):
Thanks for the contribution! Very nicely done!

vshard/storage/init.lua
andrejbranch: Agreed. Moved the compatibility checks and explanation into util.index_can_paginate(). The GC path is now a single helper call.

test/storage-luatest/garbage_collector_test.lua
andrejbranch: Simplified as suggested; old_pairs(self, key, opts) is now used directly.

Gerold103 (approved):
Thanks! Top! 🔥

(A note for future - do not resolve comments of reviewers. The reviewers will close them on their own when they double-check that the problem was addressed)

Serpentian (approved):
Looks good to me, thank you a lot for finding the issue and solving it cleanly!

andrejbranch:
Thanks all for you're prompt responses. Much appreciated!

PR #667, opened 2026-08-26
Wait for master sync during bootstrap

### LINKED ISSUE #650: Bootstrap can lead to doubled buckets
It's not a good idea to allow bootstrapping the cluster without any synchronization. Consider the case:

    1. User bootstrapps the cluster, only on 1 replicaset buckets are created.
    2. Master is changed in that replicaset. Old master still considers itself a master (due to misconfiguration), it considers itself syncrhonized, since it did that between 1 and 2.
    3. User calls boostrap one more time, it passes, since the new leader doesn't have the buckets, created on the old one.
    4. Old master considers itself in sync, sends the buckets to another replicaset
    5. We have doubled buckets.

Instead, I propose to make map call over masters in cluster_bootstrap and wait there for masters to be in sync. After that we make map call one more time and create the buckets (with protection for is_sync). Though, the bootstrap can still fail, if master changes between these two map calls, though, it seems we're ok with it now, since partial boostrap is anyway possible now, if master changes and router doesn't know about that yet (see  (see https://github.com/tarantool/vshard/issues/499).

kamenkremen:
A new master may still have no buckets while it is catching up with the old master. In this state, bootstrap may consider the cluster empty and create duplicate buckets.

Closes #650


test/luatest_helpers/vtest.lua
```diff
@@ -372,8 +372,21 @@ local function cluster_exec_each_master(g, func, args)
     end)
 end
 
+--
```
Serpentian, line 375: This is not the correct commit to add this to. The first commit adds the `bucket_create/drop`, it doesn't change the algorithm of bootstrap. Let's move it to the second commit or even to the separate one

vshard/router/init.lua
```diff
@@ -1336,8 +1336,41 @@ end
 -- Bootstrap
 --------------------------------------------------------------------------------
 
+local function cluster_wait_masters_sync(replicasets, timeout)
+    local deadline = fiber_clock() + timeout
+    local sync_err
+    repeat
```
Serpentian: Let's move the loop to the storage. There's no need to make `timeout / MASTER_ENABLE_WAIT_INTERVAL` network calls to each master, when we can do only one (per master).

vshard/router/init.lua
```diff
@@ -1336,8 +1336,41 @@ end
 -- Bootstrap
 --------------------------------------------------------------------------------
 
+local function cluster_wait_masters_sync(replicasets, timeout)
```
Serpentian: Please, let's leave the comment, that this call for masters still doesn't guarantee us, that we won't encounter th `MASTER_NOT_SYNCED` error during `bucket_create` call below. And the reason, why this can happen in the comment too.

vshard/storage/init.lua
```diff
@@ -1658,6 +1688,22 @@ local function bucket_force_drop(bucket_id)
     return true
 end
 
+--
+-- Drop bucket manually for tests or emergency cases. The operation verifies
+-- that this instance is a synchronized master.
+--
+local function bucket_drop(bucket_id)
```
Serpentian: Huh, this is a funny function. It's impossible to use it: it deletes the bucket iff master is synchronized and iff bucket is already in the `GARBAGE` state (due to `bucket_state_edges`). It doesn't even drop the data of the bucket. 

`bucket_force_drop` is also unusable, but we cannot drop it that easily due to backward compatibility. I propose not to introduce `bucket_drop` and state why in the commit msg

vshard/storage/init.lua
```diff
@@ -4642,6 +4690,8 @@ return {
     --
     -- Bucket methods.
     --
+    bucket_create = storage_make_api(bucket_create),
```
Serpentian, line 4704: Let's add some unit tests for the `bucket_create`: success, non-masterrejection, non-synced rejection

vshard/storage/init.lua
```diff
@@ -1607,8 +1615,10 @@ end
 -- are inserted.
 -- @param first_bucket_id Identifier of a first bucket in a range.
 -- @param count Bucket range length to insert. By default is 1.
+-- @param is_safe Whether to use bucket_space_* helpers instead of direct
```
Serpentian: Nit: The `@param is_safe` comment describes the implementation rather than the behavior. Perhaps something like "refuse each write unless the instance is a synchronized master"?

test/router/boot_replica_first.test.lua
```diff
@@ -32,6 +32,9 @@ util.map_evals(test_run, {REPLICASET_1}, 'bootstrap_storage(\'memtx\')')
 test_run:switch('box_1_a')
 vshard.storage.cfg(cfg, instance_uuid)
 
+test_run:switch('box_1_c')
+vshard.storage.cfg(cfg, instance_uuid)
```
Serpentian, line 36: The reason for that change in the commit msg. And let's also add there, how the doubled bucket can happen if unsafe bootstrap is used

vshard/storage/init.lua
```diff
@@ -1631,21 +1652,30 @@ local function bucket_force_create_impl(first_bucket_id, count)
...
     end
     return true
 end
 
+local function bucket_force_create(first_bucket_id, count)
+    return bucket_create_common('bucket_force_create', first_bucket_id, count,
+                                false)
+end
+
+local function bucket_create(first_bucket_id, count)
```
Serpentian, line 1702: In the doc request of the first commit msg: let's say that using force_create is unsafe and `bucket_create` should be prefered. And why

vshard/router/init.lua
```diff
@@ -1375,15 +1407,23 @@ local function cluster_bootstrap(router, opts)
     if last_err then
         return nil, err
     end
+
+    -- Wait for all masters to synchronize before creating buckets.
+    local timeout = opts and opts.timeout or consts.CALL_TIMEOUT_MAX
+    ok, err = cluster_wait_masters_sync(router.replicasets, timeout)
```
Serpentian: Please, let's merge the call with the `vshard.storage.buckets_count` above. Moreover, collecting the `bucket_count` before `cluster_wait_sync` is unsafe: a catching-up master reports 0 buckets and passes the emptiness check, then the sync wait succeeds, the master reveals its real buckets, and `bucket_create` fails with a raw duplicate-key (TUPLE_FOUND) errror

vshard/router/init.lua
```diff
@@ -1375,15 +1407,23 @@ local function cluster_bootstrap(router, opts)
     if last_err then
         return nil, err
     end
+
+    -- Wait for all masters to synchronize before creating buckets.
+    local timeout = opts and opts.timeout or consts.CALL_TIMEOUT_MAX
```
Serpentian: The sync wait defaults to CALL_TIMEOUT_MAX, while the other bootstrap calls without opts use `master.net_timeout` (~0.5s, adaptive). Please, use the `CALL_TIMEOUT_MIN` as default (it's the same e.g. for the router_call_impl)

test/router-luatest/router_2_2_test.lua
```diff
@@ -1,4 +1,5 @@
 local t = require('luatest')
+local verror = require('vshard.error')
 local vtest = require('test.luatest_helpers.vtest')
 local vutil = require('vshard.util')
```
Serpentian, line 4: Commit msg of the second one:

The doc request specifies the implementation details, which are not needed, we mention only user related stuff in doc (see the doc for router.bootstrap e.g.), it seems to me we should not add anything to it for now (it may change in the subsequent review iterations)

Serpentian (commented):
Great patch! It's very well done in general, we just should work with it a little bit

test/luatest_helpers/vtest.lua
kamenkremen: Done

vshard/router/init.lua
kamenkremen: Done

vshard/router/init.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Fixed

test/router/boot_replica_first.test.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done

vshard/router/init.lua
kamenkremen: Done

vshard/router/init.lua
kamenkremen: Done

test/router-luatest/router_2_2_test.lua
kamenkremen: Fixed

vshard/storage/init.lua
```diff
@@ -385,6 +385,33 @@ local function bucket_check_is_synced()
     return true
 end
 
+--
+-- Wait until this instance is a synchronized master and return its bucket
+-- count.
+--
+-- This does not guarantee that a later bucket_create() call succeeds. The
```
Serpentian, line 392: Not the right place for that comment. This function is abstract, the comment only makes sense in bootstrap, which calls `wait_bucket_sync` and then `bucket_create`.

So, either the comment should be moved out to the `cluster_bootstrap`, or this fucntion should be renamed to smth like `storage_bootstrap_prepare`.

I also find a little bit confusing, that the function named `storage_wait_bucket_sync` returns `bucket_count` for some reason (and yeah, it's required), but It seems we do that pretty often (i mean returning the `bucket_count`), so nevermind

vshard/storage/init.lua
Serpentian: Nah, I cannot understand from the commit msg, why the `bucket_drop` is not introduced. We should say, that it'll be unusable, the same way `bucket_force_drop` is

test/router/boot_replica_first.test.lua
Serpentian: > A new master may still have no buckets while it is catching up with the
old master. In this state, unsafe bootstrap can consider the cluster
empty and create duplicate buckets. Once the old master's records
are replicated, the cluster has duplicate buckets.

It's not the correct description. We cannot get the doubled buckets after replication)

vshard/storage/init.lua
Serpentian: > The function creates buckets only on a synchronized master.

User doens't know, what the sycnhronized master is, and he doesn't need to know that. Let's just say, that it creates the bucket in a way, which is safe with bucket rebalancing, and consequently should be preferred to `force_create`

test/storage-luatest/bucket_triggers_test.lua
```diff
@@ -102,6 +102,58 @@ local function bucket_set_protection(value)
     ivshard.storage.internal.is_bucket_protected = value
 end
 
+test_group.test_bucket_create = function(g)
```
Serpentian, line 105: The test case just checks, that the `bucket_create` creates the bucket in `ACTIVE` state, pretty useless from my point of view, can be dropped.

test/storage-luatest/bucket_triggers_test.lua
```diff
@@ -102,6 +102,58 @@ local function bucket_set_protection(value)
...
+        local ok, err = ivshard.storage.bucket_create(bid)
+        return ok, err, box.space._bucket:get(bid)
+    end, {bid})
+    t.assert_equals(ok, nil)
+    t.assert_equals(err.code, verror.code.NON_MASTER)
+    t.assert_equals(bucket, nil)
+end
+
+test_group.test_bucket_create_on_unsynced_master = function(g)
+    local bid = cfg_template.bucket_count + 1
```
Serpentian: Let's consider reproducing the real non-sync state, this just replaces internal variable.

vshard/storage/init.lua
```diff
@@ -1619,8 +1629,19 @@ local function bucket_force_create_impl(first_bucket_id, count)
         -- work of vshard. So, as we don't want to disable the protection
         -- of the buckets for the whole replicaset for bootstrap, a bucket's
         -- status must go the following way: none -> RECEIVING -> ACTIVE.
-        _bucket:insert({i, BRECEIVING})
-        _bucket:replace({i, BACTIVE})
+        if is_safe then
+            local bucket, err = bucket_space_insert({i, BRECEIVING})
```
Serpentian, line 1633: Nit: I'd consider introducing the `bucket_space_insert/replace_xc`

vshard/storage/init.lua
```diff
@@ -1631,21 +1652,30 @@ local function bucket_force_create_impl(first_bucket_id, count)
     box.commit()
 end
 
-local function bucket_force_create(first_bucket_id, count)
+local function bucket_create_common(name, first_bucket_id, count, is_safe)
```
Serpentian, line 1655: Nit: we don't use the `_comon` names at all, maybe `_internal`? And I'd probably prefer, that the name is figured out from the `is_safe` arg, but I'm not sure, do as u want)

test/luatest_helpers/vtest.lua
```diff
@@ -372,8 +372,21 @@ local function cluster_exec_each_master(g, func, args)
...
+-- Wait for _bucket space being synchronized on master. Without synchronization
+-- the buckets cannot be sent or received, rebalancer and recovery doesn't work.
+--
+local function storage_wait_bucket_sync(storage)
+    storage:exec(function()
+        local count, err = ivshard.storage._call('storage_wait_bucket_sync',
```
Serpentian, line 381: Nit: `count` can be replaced with `_`, you don't need it

test/luatest_helpers/vtest.lua
```diff
@@ -419,6 +432,9 @@ local function cluster_bootstrap(g, cfg)
         replicaset_count = replicaset_count + 1
     end
     t.assert_not_equals(masters, {}, 'have masters')
+    local _, sync_err = cluster_for_each_in(masters,
+                                            storage_wait_bucket_sync)
```
Serpentian, line 436: Can be in one line

test/storage-luatest/bucket_triggers_test.lua
```diff
@@ -102,6 +102,58 @@ local function bucket_set_protection(value)
...
+        internal.is_bucket_in_sync = false
+        local ok, err = ivshard.storage.bucket_create(bid)
+        internal.is_bucket_in_sync = was_in_sync
+        return ok, err, box.space._bucket:get(bid)
+    end, {bid})
+    t.assert_equals(ok, nil)
+    t.assert_equals(err.code, verror.code.MASTER_NOT_SYNCED)
+    t.assert_equals(bucket, nil)
+end
+
```
Serpentian, line 156: Nit: let's add new tests to the end of the file, when possible

vshard/router/init.lua
```diff
@@ -1352,9 +1351,13 @@ local function cluster_bootstrap(router, opts)
         if_not_bootstrapped = false
     end
 
+    local timeout = opts and opts.timeout or consts.CALL_TIMEOUT_MIN
+    local wait_timeout = math.max(
+        timeout - consts.MASTER_ENABLE_WAIT_INTERVAL, 0)
```
Serpentian, line 1356: Let's use the same approach as in the `bucket_test_send_on_replicas`, division and not the subtraction

vshard/router/init.lua
```diff
@@ -1352,9 +1351,13 @@ local function cluster_bootstrap(router, opts)
         if_not_bootstrapped = false
     end
 
+    local timeout = opts and opts.timeout or consts.CALL_TIMEOUT_MIN
```
Serpentian, line 1354: Nit: Hmm, by default there was `net_timeout`, now it's `CALL_TIMEOUT_MIN`, they're not always the same, probably the comment is enough, too minor to think about, and too difficult to figure out the correct `net_timoeut` (the master may be missing, old master can be replica already and so on)

vshard/storage/init.lua
```diff
@@ -385,6 +385,33 @@ local function bucket_check_is_synced()
...
+        end
+        if err.code ~= lerror.code.MASTER_NOT_SYNCED then
+            return nil, err
+        end
+
+        timeout = deadline - fiber_clock()
+        if timeout <= 0 then
+            return nil, err
+        end
+        lfiber.sleep(math.min(consts.MASTER_ENABLE_WAIT_INTERVAL, timeout))
```
Serpentian, line 411: Let's use the fiber_cond instead

test/router-luatest/router_2_2_test.lua
```diff
@@ -599,6 +600,103 @@ local function get_auto_master_global_cfg()
...
+    res, err = router_bootstrap(g.router, 0.1)
+    t.assert_equals(res, nil)
+    t.assert_equals(err and err.code, verror.code.MASTER_NOT_SYNCED)
+    t.assert_equals(storage_bucket_count(g.replica_2_a), 0)
+
+    g.replica_1_b:exec(function(replication)
+        box.cfg({replication = replication})
+    end, {replication})
+    storage_set_master(g.replica_1_a, false)
+    storage_set_master(g.replica_1_b, true)
```
Serpentian, line 682: `1_b` is already a master at this point, so this one is not needed, or am I missing smth?

Serpentian (commented):
Well done! Only nits, no major comments from me

test/luatest_helpers/vtest.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Fixed

vshard/storage/init.lua
kamenkremen: Done

test/storage-luatest/bucket_triggers_test.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done

test/router/boot_replica_first.test.lua
kamenkremen: Fixed

vshard/storage/init.lua
kamenkremen: Fixed

test/luatest_helpers/vtest.lua
kamenkremen: Fixed

test/storage-luatest/bucket_triggers_test.lua
kamenkremen: Fixed

vshard/router/init.lua
kamenkremen: Done

vshard/router/init.lua
kamenkremen: Done

vshard/storage/init.lua
kamenkremen: Done, added a dedicated bucket synchronization condition to use here

test/router-luatest/router_2_2_test.lua
kamenkremen: Yeah, this one is not needed. Removed

vshard/storage/init.lua
kamenkremen: Moved the comment

PR #668, opened 2026-09-07
Encode error objects into json before logging them

### LINKED ISSUE #624: Malformed replication status log message
Right now I see in logs following message:
```
{"time": "2025-11-26T13:30:57.882+0300", "level": "WARN", "message": "replication_status on '*.*.*.*:*' is unknown: table: 0x4e5b5148", "pid": 17372 , "cord_name": "main", "fiber_id": 2930124, "fiber_name": "vshard.replica.*.*.*.*:*", "file": "...rantool\/tdg\/.rocks\/share\/tarantool\/vshard\/replicaset.lua", "line": 1343, "module": "tdg.vshard.replicaset"}
```

Seems we should deserialize `table: 0x4e5b5148` for logs in a different way.

### LINKED ISSUE #651: Printing errors should be wrapped with `json_encode` function
Current master works ok with error objects in `log`. However, it seems that old ones print the table pointer instead of error object, e.g.

```
2026-05-23 18:50:53.396 [402275] main/128/vshard.recovery log_ratelimit.lua:136 E> Error during searching in cluster for recovery of 9 bucket: table: 0x402b2f78
```

It can be fixed by logging as follows:

```lua
limiter:log_error(err,  'Error during searching in cluster for recovery of ' ..
      '%d bucket: %s', bucket_id, json_encode(err))
```

Related https://github.com/tarantool/vshard/issues/624

jadhavgaurav:
Closes #651. Also closes #624, which reports the same symptom for the `replication_status ... is unknown` message.

### Problem

VShard error objects are plain Lua tables carrying `__tostring = json.encode` (`vshard/error.lua`). Tarantool's `log` module formats its arguments through `string.format()`, and only recent LuaJIT versions call `__tostring` for `%s`. On the older Tarantool versions vshard still supports, an error object is printed as a bare table address:

```
E> Error during searching in cluster for recovery of 9 bucket: table: 0x402b2f78
```

The recovery service was already fixed in c347440c ("Let's fix only recovery for now"). This does the rest.

### Change

`json_encode(err)` at every remaining log/`set_status_error` site that prints an error object:

- `vshard/storage/init.lua`: recovery, garbage collector, master_sync and rebalancer services, plus the rebalancer worker and route-applying messages.
- `vshard/router/init.lua`: both discovery failure messages.
- `vshard/replicaset.lua`: initial-handshake close, `replica_call` exception, replica backoff, master locate/search, failover ping and step, worker service failure, and the `replication_status` update from `replica_failover_ping` (that last one is #624).
- `vshard/util.lua`: the reloadable fiber loop failure message.

Two notes on scope:

- Where the value comes from `pcall()` and the surrounding code already normalised it with `lerror.make()`, the `lerror.make()` call is kept and wrapped. In `replica_call()` the existing `lerror.make(err)` is hoisted above the log so the logged object and the returned object are the same one.
- Messages built from `pcall()` over plain box API calls (dropping `on_replace` triggers, `_schema` upgrade/downgrade in `vshard/storage/schema.lua`) are deliberately left alone: those errors are `box.error` cdata, which never prints as a table address.

`json.encode` is safe for every shape that reaches these sites: an error table, a `box.error` cdata (serialised to its message) and a plain string all encode without raising.

### Testing

No new test, matching `NO_TEST=minor` on c347440c. The bug is invisible on any Tarantool where `string.format('%s', obj)` honours `__tostring`, which is the case for every version I can run locally — the pre-fix code already emits the same json there. `test/storage-luatest/storage_1_1_1_test.lua` already asserts json content (`'Error during downloading rebalancer states:.*"replicaset_id":"%s"'`) for one of these messages; this change makes that hold on the old versions too.

Full suite run on Tarantool 3.8.1 (macOS): **93 passed, 1 skipped, 5 disabled, 1 failed**. The single failure is `upgrade/upgrade.test.lua`, which fails identically on unmodified `master` here with a test-run harness error (`RuntimeError: dictionary changed size during iteration` in `preprocessor.py`), so it is unrelated to this change. I could not test against an old Tarantool where the bug actually reproduces, so the behaviour on those versions is reasoned about rather than observed.


vshard/router/init.lua
```diff
@@ -24,6 +24,7 @@ local lreplicaset = require('vshard.replicaset')
 local lservice_info = require('vshard.service_info')
 local lratelimit = require('vshard.log_ratelimit')
 local util = require('vshard.util')
+local json_encode = require('json').encode
```
Serpentian: Let's introduce the test, the problem is not that minor: error logging is crucial during incident investigation. I'd propose to test one error from the changed ones (check, that it's properly printed with `grep_log`).

> Messages built by pcall() over plain box API calls are left alone: those errors are box.error cdata, which is never printed as a table address.

Let's also check, that these cdata errors are also printed correctly, again, any one place in code.

vshard/storage/init.lua
```diff
@@ -1251,7 +1251,8 @@ local function recovery_service_f(service, limiter)
             if not ok then
                 is_all_recovered = false
                 limiter:log_error(total, service:set_status_error(
-                    'Error during %s buckets recovery: %s', status, total))
+                    'Error during %s buckets recovery: %s', status,
+                    json_encode(total)))
```
Serpentian: >  Tarantool's log module formats its arguments with string.format(), and only recent LuaJIT versions call __tostring for '%s'.

Maybe it's better to use `tostring` then? And the `box.error` should be logged as `json_encode(lerror.make(err))`

vshard/router/init.lua
```diff
@@ -24,6 +24,7 @@ local lreplicaset = require('vshard.replicaset')
 local lservice_info = require('vshard.service_info')
```
Serpentian: Let's name the commit msg smth like: `log: encode error objects into json before logging them`. You can see the commit msg guidelines here: https://www.tarantool.io/en/doc/2.11/dev_guide/developer_guidelines/#how-to-write-a-commit-message

vshard/replicaset.lua
```diff
@@ -708,9 +709,11 @@ local function replica_call(replica, func, args, opts)
            err.type == 'ClientError') then
             err = lerror.from_string(err.message) or err
         end
+        err = lerror.make(err)
         replica.limiter:log_error(err,
-            "Exception during calling '%s' on '%s': %s", func, replica, err)
-        return false, nil, lerror.make(err)
+            "Exception during calling '%s' on '%s': %s", func, replica,
+            json_encode(err))
```
Serpentian: The `json.encode` actually is not that safe and can cause LuaJit errors:

```lua
unix/:./data/router_1.control> require('json').encode(require('vshard.error').make(box.error.new{type = 'ReviewError', reason = 'original application error', payload = {[false] = 'value'}}))
---
- error: table key must be a number or string
...
unix/:./data/router_1.control> require('json').encode(require('vshard.error').make({message = 'original error', context = function() end}))
---
- error: unsupported Lua type 'function'
...
```

Let's use `tostring` everywhere and make the `__tostring` in `error` module safer? WDYT?

Serpentian (commented):
Congratulations on your first contribution! Great patch, let's work a little bit more with it)

